PR #3, opened 2015-10-26
Adds ssh client to V2.0.0

alcortesm:
SSH client for go-git.v2.


clients/ssh/git_upload_pack.go
```diff
@@ -0,0 +1,194 @@
...
+	"fmt"
+	"io"
+	"io/ioutil"
+	"net"
+	"net/url"
+	"os"
+
+	"github.com/sourcegraph/go-vcsurl"
+
+	"golang.org/x/crypto/ssh"
```
mcuadros: the third party packages goes at the end of the import list

clients/ssh/git_upload_pack.go
```diff
@@ -0,0 +1,194 @@
...
+	return client, nil
+}
+
+// TODO: this must be done securely
+func connectWithPasswd(host, user string) (*ssh.Client, error) {
+	var pass string = "your password"
+
+	sshConfig := &ssh.ClientConfig{
+		User: user,
+		Auth: []ssh.AuthMethod{ssh.Password(pass)},
```
mcuadros: Use certs instead of passwords

clients/ssh/git_upload_pack_test.go
```diff
@@ -0,0 +1,111 @@
...
+	"io/ioutil"
+	"testing"
+
+	"gopkg.in/src-d/go-git.v2/clients/common"
+	. "gopkg.in/src-d/go-git.v2/clients/ssh"
+)
+
+const fixtureRepo = "git@github.com:tyba/git-fixture.git"
+
+func TestConnect(t *testing.T) {
```
mcuadros: use gocheck for the tests

clients/ssh/git_upload_pack.go
alcortesm: TODO.

clients/ssh/git_upload_pack.go
alcortesm: Actually I'm using certs, connectPasswd is only called if no certs are found.

In any case, the current implementation is not the right way to do it, so instead of leave it there with a comment I will just remove it.

TODO.

clients/ssh/git_upload_pack_test.go
alcortesm: OK, I will use the http checks as a template for the ssh ones.

TODO.

clients/ssh/git_upload_pack.go
```diff
@@ -0,0 +1,194 @@
...
+	}
+
+	client, err := ssh.Dial("tcp", host, sshConfig)
+	if err != nil {
+		return nil, fmt.Errorf("failed to dia: %v", err)
```
toqueteos: Typo: `dia` -> `dial`

clients/ssh/git_upload_pack.go
```diff
@@ -0,0 +1,194 @@
...
+	}
+
+	return client, nil
+}
+
+func (s *GitUploadPackService) Info() (i *common.GitUploadPackInfo, err error) {
+	if !s.connected {
+		return nil, fmt.Errorf("not connected")
```
toqueteos: Just in this case, because it's the same as `errors.New("not connected")`, you can make the error a public variable, like `ErrNotConnected` so library users can check against it. All other errors need a common error type.

clients/ssh/git_upload_pack.go
```diff
@@ -0,0 +1,194 @@
...
+	}
+	session.Wait()
+
+	// read until the header of the second answer
+	soBuf := bufio.NewReader(so)
+	token := "0000"
+	for {
+		line, err := soBuf.ReadString('\n')
+		if err == io.EOF {
+			return nil, fmt.Errorf("Bad answer format to git-upload-pack")
```
toqueteos: Same as before: `var ErrGitUploadBadAnswer = errors.New("Bad answer format to git-upload-pack")`, error name is just a proposal.

clients/ssh/git_upload_pack.go
```diff
@@ -0,0 +1,194 @@
...
+func (s *GitUploadPackService) Disconnect() (err error) {
+	if !s.connected {
+		return fmt.Errorf("not connected")
+	}
+	if err = s.client.Close(); err != nil {
+		return err
+	}
+	s.client = nil
+	s.vcs = nil
+	return nil
```
toqueteos: Same as:

```go
s.client = nil
s.vcs = nil
return s.client.Close()
```

mcuadros:
@alcortesm ping

clients/ssh/git_upload_pack.go
```diff
@@ -0,0 +1,223 @@
...
+)
+
+var (
+	InvalidAuthMethodErr      = errors.New("invalid ssh auth method: a ssh.SSHAuthMethod should be provided.")
+	AuthRequiredErr           = errors.New("cannot connect: auth required.")
+	NotConnectedErr           = errors.New("not connected")
+	AlreadyConnectedErr       = errors.New("already connected")
+	UploadPackAnswerFormatErr = errors.New("git-upload-pack bad answer format")
+	UnsupportedVCSErr         = errors.New("only git is supported")
+	UnsupportedRepoErr        = errors.New("only github.com is supported")
```
toqueteos: Minor thing related to Go style, could we move the Err on all of these to the beginning? Examples:
https://golang.org/pkg/os/#pkg-variables, https://golang.org/pkg/net/http/#pkg-variables

alcortesm: I would love to do that, should I @mcuadros ?

toqueteos:
LGTM

alcortesm:
The current version does not implement the new AuthMethod functionality correctly and I have not understand what should the String() method does. 

@mcuadros suggested I should take a look at ssh.AuthMethod, so all of them can be passed to the SSH client.

I have not fully understand all Máximo comments: I will need to think about it. Things will be much easier if the appropriate sections of go-git where documented.

alcortesm:
Done.

But I still don't understand the purpose of having all this auth methods in go-git.client.ssh, there should be only one, that wraps a generic an externally generated []ssh.AuthMethod in a go-git.client.common.AuthMethod. Should we discuss about it?

mcuadros:
I dont understand how the your AuthMethods works, how I should set for example the credentials here:
https://github.com/src-d/go-git/pull/3/files#diff-93a872f722160e3c55c4194ccfb29664R57ç

BTW can you squash

alcortesm:
I have gotten myself into a squashing hell, let's continue this PR in a new one.

PR #6, opened 2015-11-23
Timezone fixes

klaidliadon:
Adding timezone infos to commits

### DIFF
--- a/objects.go
+++ b/objects.go
@@ -76,7 +76,15 @@ func (s *Signature) Decode(b []byte) {
 			if c == ' ' || end {
 				t, err := strconv.ParseInt(string(b[from:i]), 10, 64)
 				if err == nil {
-					s.When = time.Unix(t, 0)
+					loc := time.UTC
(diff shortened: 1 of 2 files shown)


objects.go
```diff
@@ -76,7 +76,22 @@ func (s *Signature) Decode(b []byte) {
...
+					loc := time.UTC
+					if len(b[i:]) >= 6 {
+						var sign int
+						switch b[i+1] {
+						case '+':
+							sign = 1
+						case '-':
+							sign = -1
+						}
+						h, _ := strconv.ParseInt(string(b[i+2:i+4]), 10, 64)
```
mcuadros: I prefer use the Parse function of time:

```
package main

import (
	"fmt"
	"time"
)

func main() {

	now := time.Now()

	tz, err := time.Parse("-0700", "-1200")
	fmt.Println(tz.Location(), err)

	fmt.Println(now.In(tz.Location()))

}
```

PR #7, opened 2015-11-25
Blame

alcortesm:
This also includes a diff package and revlist package (needed by
blame)

Some extra packfiles (<1MB) are also included, to be used as fixtures in
the tests.


blame/blame.go
```diff
@@ -0,0 +1,223 @@
...
+	result := make([]*git.Commit, 0, len(fVs))
+	for _, v := range fVs {
+		c := git.Commit(*v)
+		result = append(result, &c)
+	}
+	return result, nil
+}
+
+// this struct is internally used by the blame function to hold its
+// intputs, outputs and state.
```
toqueteos: Typo: `intputs -> inputs`

blame/blame_test.go
```diff
@@ -0,0 +1,810 @@
...
+}
+
+// create the repositories of the fixtures
+func (s *SuiteCommon) SetUpSuite(c *C) {
+	s.repos = make(map[string]*git.Repository, 0)
+	for _, fixRepo := range fixtureRepos {
+		s.repos[fixRepo.url] = git.NewPlainRepository()
+
+		d, err := os.Open(fixRepo.packfile)
+		defer func() {
```
toqueteos: CAUTION: defer inside for won't close at end of for

blame/blame_test.go
```diff
@@ -0,0 +1,810 @@
...
+			c.Assert(d.Close(), IsNil)
+		}()
+		c.Assert(err, IsNil)
+
+		r := packfile.NewReader(d)
+		r.Format = packfile.OFSDeltaFormat // TODO: how to know the format of a pack file ahead of time?
```
toqueteos: The pktline we get before the packfile specifies which capabilities the packfile does support.

blame/blame_test.go
```diff
@@ -0,0 +1,810 @@
...
+
+		_, err = r.Read(s.repos[fixRepo.url].Storage)
+		c.Assert(err, IsNil)
+	}
+}
+
+var blameTests = [...]struct {
+	// input data to revlist
+	repo   string
+	branch string // TODO: remove this, it is no longer needed for local packfiles
```
toqueteos: TODO?

blame/blame_test.go
```diff
@@ -0,0 +1,810 @@
...
+	r := make([]string, 0, n)
+	for i := 0; i < n; i++ {
+		r = append(r, s)
+	}
+	return r
+}
+
+// utility function to concat slices
+func concat(dst *[]string, vargs ...[]string) []string {
+	for _, ss := range vargs {
```
toqueteos: Apparently all tests use a `&[]string{}` as first input, wouldn't it be easier to just declare a `var dst []string` in here and return that?

common_test.go
```diff
@@ -42,3 +43,113 @@ func (s *MockGitUploadPackService) Fetch(*common.GitUploadPackRequest) (io.ReadC
...
+}
+
+// create the repositories of the fixtures
+func (s *SuiteCommon) SetUpSuite(c *C) {
+	s.repos = make(map[string]*Repository, 0)
+	for _, fixRepo := range fixtureRepos {
+		s.repos[fixRepo.url] = NewPlainRepository()
+
+		d, err := os.Open(fixRepo.packfile)
+		defer func() {
```
toqueteos: CAUTION: defer inside for

blame/blame.go
```diff
@@ -0,0 +1,223 @@
...
+//    struct, that is not exported.
+//
+//    TODO: ways to improve the efficiency of this function:
+//
+//    1. Improve revlist
+//
+//	  2. Improve how to traverse the history (example a backward
+//	  traversal will be much more efficient)
+//
+//    TODO: ways to improve the functrion in general
```
mvader: `functrion` Typo or intended?

blame/blame.go
```diff
@@ -0,0 +1,223 @@
...
+//
+//    1. Add memoization betweenn revlist and assign.
+//
+//    2. It is using much more memmory than needed, see the TODOs below.
+func Blame(repo *git.Repository, commit *git.Commit, path string) ([]*git.Commit, error) {
```
mvader: the function is kinda big, can't be splitted into a few smaller funcs?

blame/blame.go
```diff
@@ -0,0 +1,223 @@
...
+// revision
+func (b *blame) assignOrigin(c, p int) {
+	// assign origin based on diff info
+	hunks := diff.Do(b.data[p], b.data[c])
+	sl := -1 // source line
+	dl := -1 // destination line
+	for h := range hunks {
+		hLines := git.CountLines(hunks[h].Text)
+		for hl := 0; hl < hLines; hl++ {
+			// fmt.Printf("file=%q, rev=%d, r=%d, h=%d, hunk=%v, hunkLine=%d\n", file, rev, r, h, hunks[h], hl)
```
mvader: letfover?

blame/blame.go
```diff
@@ -0,0 +1,223 @@
...
+			case hunks[h].Type == 0:
+				sl++
+				dl++
+				b.graph[c][dl] = b.graph[p][sl]
+			case hunks[h].Type == 1:
+				dl++
+				b.graph[c][dl] = vertex(b.revs[c])
+			case hunks[h].Type == -1:
+				sl++
+			default:
```
mvader: if this is unreachable why panic or have the default clause at all?

blame/blame.go
```diff
@@ -0,0 +1,223 @@
...
+				panic("unreachable")
+			}
+		}
+	}
+}
+
+// This will print the results of a Blame as in git-blame.
```
mvader: By convention, the docs of the functions are:
```go
// FunctionName whatever whatever whatever
```

blame/blame_test.go
```diff
@@ -0,0 +1,810 @@
...
+	)},
+	{"https://github.com/tyba/git-fixture.git", "master", "6ecf0ef2c2dffb796033e5a02219af86ec6584e5", "php/crappy.php", concat(&[]string{},
+		repeat("918c48b83bd081e863dbe1b80f8998f058cd8294", 259),
+	)},
+	{"https://github.com/tyba/git-fixture.git", "master", "6ecf0ef2c2dffb796033e5a02219af86ec6584e5", "vendor/foo.go", concat(&[]string{},
+		repeat("6ecf0ef2c2dffb796033e5a02219af86ec6584e5", 7),
+	)},
+	/*
+		// Failed
+		{"https://github.com/spinnaker/spinnaker.git", "master", "f39d86f59a0781f130e8de6b2115329c1fbe9545", "InstallSpinnaker.sh", concat(&[]string{},
```
mvader: letfovers?

blame/blame_test.go
mvader: +1 to that

common.go
```diff
@@ -0,0 +1,49 @@
...
+	nEol := strings.Count(s, "\n")
+	if strings.HasSuffix(s, "\n") {
+		return nEol
+	}
+	return nEol + 1
+}
+
+// FindFile searches for a path in a commit. Returns the file and true if found.
+// Returns nil and false if not found.
+// TODO: should this be a method of git.Commit instead?
```
mvader: IMHO, yes

common_test.go
```diff
@@ -42,3 +43,113 @@ func (s *MockGitUploadPackService) Fetch(*common.GitUploadPackRequest) (io.ReadC
...
+	i string // the string we want to count lines from
+	e int    // the expected number of lines in i
+}{
+	{"", 0},
+	{"a", 1},
+	{"a\n", 1},
+	{"a\nb", 2},
+	{"a\nb\n", 2},
+	{"a\nb\nc", 3},
+	{"a\nb\nc\n", 3},
```
mvader: "a\n\n\nb\n" => 4 is intended, right? I mean, we want all the lines and not just the lines with text in it, right?

revlist/revlist.go
```diff
@@ -0,0 +1,250 @@
...
+func parentsContainingPath(path string, c *git.Commit) []*git.Commit {
+	var result []*git.Commit
+	iter := c.Parents()
+	for {
+		parent, err := iter.Next()
+		if err != nil {
+			if err == io.EOF {
+				return result
+			}
+			panic("unreachable")
```
mvader: do we want unreachable panics?

mvader:
Except for the aforementioned things, LGTM

blame/blame.go
toqueteos: Branch exhaustiveness, this is automatically handled by latest compiler versions (since 1.4 I think..) It shouldn't apply here because there's no return inside each case.

blame/blame_test.go
alcortesm: That's nice, but the packfiles I'm using has been copied directly from .git/objects/pack/*.pack after a "git gc", so I don't have access to the pktline for those packfiles.

alcortesm: Anybody knows how to ask a packfile about its format?

(maybe the solution is: try to read a packfile using one format, and if there is an error, then try the other format).

blame/blame.go
alcortesm: Typo :)

blame/blame.go
alcortesm: good catch!

blame/blame_test.go
alcortesm: No. I want to keep this tests commented on purpose.

These tests fail because the output of a git-blame is not unique. Right now our library and git-blame disagree on some cases. In the future we might be able to fix this discrepancies and we will need these tests cases to prove it.

Finding repos with cases like this takes a lot of time (hours), so I don't want to delete them now that I have found some.

common_test.go
alcortesm: Yes, we want all the lines, even empty lines, as empty lines are part of some languages syntax.

alcortesm: I have added a test to clarify this.

blame/blame_test.go
mvader: :+1:

revlist/revlist.go
alcortesm: I'd say yer, we need them.

Without the panic, If iter.Next() changes in the future, and start returning other errors, they will be invisible to our code.

This way, the problem can be easily identified and fixed.

alcortesm: I just realize that my previous answer was not really an answer.

Will it be OK to return an error instead of a panic. Something like this:

```
		parent, err := iter.Next()
		if err != nil {
			if err == io.EOF {
				return result. nil
			}
			return nil, err
		}
```

This way there won't be any panics in the code (which is good), but we are losing the panic semantics that warns us about future library changes.

blame/blame_test.go
alcortesm: Interesting: https://codewords.recurse.com/issues/three/unpacking-git-packfiles

alcortesm:
@mcuadros wrote on slack:

Create a Blame struct to use as the blame output (and only entry point to blame), so we can marshall it to DBs.

alcortesm:
PTAL, this should solve all unanswered previous comments, including @mcuadros one on slack.

blame/blame.go
```diff
@@ -0,0 +1,286 @@
...
+//    TODO: ways to improve the efficiency of this function:
+//
+//    1. Improve revlist
+//
+//	  2. Improve how to traverse the history (example a backward
+//	  traversal will be much more efficient)
+//
+//    TODO: ways to improve the function in general:
+//
+//    1. Add memoization betweenn revlist and assign.
```
toqueteos: typo: `betweenn`

blame/blame.go
```diff
@@ -0,0 +1,286 @@
...
+//
+//    2. It is using much more memmory than needed, see the TODOs below.
```
toqueteos: typo: `memmory`

blame/blame.go
```diff
@@ -0,0 +1,286 @@
...
+type blame struct {
+	repo  *git.Repository // the repo holding the history of the file to blame
+	path  string          // the path of the file to blame
+	fRev  *git.Commit     // the commit of the final revision of the file to blame
+	revs  revlist.Revs    // the chain of revisions affecting the the file to blame
+	data  []string        // the contents of the file across all its revisions
+	graph [][]*git.Commit // the graph of the lines in the file across all the revisions TODO: not all commits are needed, only the current rev and the prev
+}
+
+// calculte the history of a file "path", from commit "from, sorted by commit date.
```
toqueteos: > from commit "from

???

blame/blame.go
```diff
@@ -0,0 +1,286 @@
...
+//    commit. Assign that commit as the creator of all its lines.
+//
+//    Then jump to the nodes in the next commit, and calculate the diff
+//    between the two files. Newly created lines get
+//    assigned the new commit as its origin. Modified lines also get
+//    this new commit. Untouched lines retain the old commit.
+//
+//    All this work is done in the assignOrigin function.
+//
+//    This function holds all the internal relevant data in a blame
```
toqueteos: Rephrase proposal: `All this work is done in the assignOrigin function which holds all the internal relevant data in a blame struct, that is not exported.`

blame/blame.go
```diff
@@ -0,0 +1,286 @@
...
+				sl++
+			default:
+				panic("unreachable")
+			}
+		}
+	}
+}
+
+// PrettyPrint prints the results of a Blame using git-blame's style.
+func (b *blame) PrettyPrint() string {
```
toqueteos: Candidate for [GoStringer](https://golang.org/pkg/fmt/#GoStringer)?

alcortesm:
Thanks, will fix.

revlist/revlist.go
```diff
@@ -0,0 +1,272 @@
...
+		var empty core.Hash
+		return empty, found
+	}
+	return file.Hash, true
+}
+
+// Returns a new slice of commits, with duplicates removed.  Expects a
+// sorted commit list.  Duplication is defined according to "comp".  It
+// will always keep the first commit of a series of duplicated commits.
+func removeComp(path string, cs []*git.Commit, comp func(string, *git.Commit, *git.Commit) (bool, error)) ([]*git.Commit, error) {
```
toqueteos: Create a type for `func(string, *git.Commit, *git.Commit) (bool, error)`

revlist/revlist.go
```diff
@@ -0,0 +1,272 @@
...
+			return nil, err
+		}
+		if !equals {
+			result = append(result, cs[i])
+		}
+	}
+	return result, nil
+}
+
+// Equivalent commits are commits whos patch is the same.
```
toqueteos: typo? `whos` -> `whose`?

revlist/revlist.go
```diff
@@ -0,0 +1,272 @@
...
+	return true
+}
+
+func sameDiff(a, b diffmatchpatch.Diff) bool {
+	if a.Type != b.Type {
+		return false
+	}
+	switch a.Type {
+	case 0:
+		return git.CountLines(a.Text) == git.CountLines(b.Text)
```
toqueteos: Is just counting the lines enough here?

toqueteos:
I probably missed something but it's late enough, I'll continue with the review during the weekend. Take a break @alcortesm!

revlist/revlist.go
alcortesm: Yes, we don't care about the contents of the lines only if the number of "untouched" ones are the same, that is what make to diffs equivalents.

PR #10, opened 2015-12-22
Adds external protocol insertion for the clients and protocol selection based on the scheme

alcortesm:
This patch allows the user to inject its own protocols to access git repositories, not just the standard HTTP and SSH.

It also adds automatic selection of the protocol to use based on the scheme of the URL of the repository.


toqueteos:
LGTM :godmode:

PR #14, opened 2016-01-21
Fix commit.File() gorutine leak

alcortesm:
Commit.File() was leaking a goroutine because it was looping over
an iterator without closing its channel.

Now commit.File() calls the new Tree.File() method that searches
the file in the repository by trasversing the dir tree instead
of using the tree.Files() iterator.

This not only prevent the goroutine leak, but also speeds up
file searching.


tree.go
```diff
@@ -26,6 +28,84 @@ type TreeEntry struct {
 	Hash core.Hash
 }
 
+// New errors defined by this package.
+var ErrFileNotFound = errors.New("file not found")
+
+func (t *Tree) File(path string) (*File, error) {
```
toqueteos: This is a public method, it needs a comment :tada:

toqueteos:
LGTM

Issue #19, opened 2016-01-30
Add support for file-based object storage


JoshuaSjoding:
The go-git library currently relies on in-memory storage of objects. I am interested in adding support for file-based object storage. My long term goal is to use this as a lightweight library for cloning and pulling repositories to local disk.

Right now RAWObjectStorage is the only implementation of the ObjectStorage interface. I would like to contribute by adding a FileObjectStorage implementation that reads and writes objects to the file system.

Is this something that the developers are amenable to?

To meet my goal there would of course be a number of other required features, such as index file support, but those can be tracked as separate issues.

JoshuaSjoding:
The current implementation of the Repository [directly includes RAWObjectStorage](https://github.com/src-d/go-git/blob/e9164d853bed12ed9f6787a608db32504c894a64/repository.go#L22). I suggest that the Repository be changed to work through the [ObjectStorage interface](https://github.com/src-d/go-git/blob/e9164d853bed12ed9f6787a608db32504c894a64/core/object.go#L20) instead. This way a FileObjectStorage implementation could take advantage of it.

To do this and still support the needs of the Repository, the ObjectStorage interface will need to provide some means of iterating over objects. This means an API change is necessary.

As an experiment last week I familiarized myself with the code by making such a change. You can see the result of the experiment [here](https://github.com/scjalliance/go-git/commit/855517cb6c6796c8893c0d9191e88ccf6e1b5c7a). You'll notice that I messed with the Iter implementations pretty heavily.

alcortesm:
Dear Joshua,

We are already working on that, index and everything.

We will contact you again again with more details in a few days, when we
come back from FOSDEM (Brussels) to our offices in Madrid.

On Saturday, January 30, 2016, Joshua Sjoding <notifications@github.com>
wrote:

> The current implementation of the Repository directly includes
> RAWObjectStorage
> <https://github.com/src-d/go-git/blob/e9164d853bed12ed9f6787a608db32504c894a64/repository.go#L22>.
> I suggest that the Repository be changed to work through the ObjectStorage
> interface
> <https://github.com/src-d/go-git/blob/e9164d853bed12ed9f6787a608db32504c894a64/core/object.go#L20>
> instead. This way a FileObjectStorage implementation could take advantage
> of it.
>
> To do this and still support the needs of the Repository, the
> ObjectStorage interface will need to provide some means of iterating over
> objects. This means an API change is necessary.
>
> As an experiment last week I familiarized myself with the code by making
[...]

PR #20, opened 2016-02-02
Iterable ObjectStorage interface for use in Repository struct

JoshuaSjoding:
The current implementation of the Repository struct includes a [pointer to core.RAWObjectStorage](https://github.com/src-d/go-git/blob/e9164d853bed12ed9f6787a608db32504c894a64/repository.go#L22). This seems inappropriate because it prevents the development of alternative implementations of [ObjectStorage](https://github.com/src-d/go-git/blob/e9164d853bed12ed9f6787a608db32504c894a64/core/object.go#L20). I presume that the direct reference was included because it provided convenient and fast access to the commit objects.

This pull request proposes the addition of one function to the ObjectStorage interface: [IterType](https://github.com/scjalliance/go-git/blob/4172d931089d9851614217df7c8f57693587399e/core/object.go#L24). This function returns an iterator of objects matching the given object type. The Repository is modified to make use of this new function, and RAWObjectStorage is modified to provide it.

The return value of IterType is a new interface called [ObjectIter](https://github.com/scjalliance/go-git/blob/4172d931089d9851614217df7c8f57693587399e/core/object.go#L57). Two implementations of ObjectIter are provided:

* ObjectLookupIter, which iterates over a slice of object hashes and looks up the given objects as needed. This is used to iterate through parent commits, for example.
* ObjectSliceIter, which iterates over a slice of objects.

Additional implementations could easily be added.

Both CommitIter and TreeIter are modified to make use of the new ObjectIter interface, as is Commit.Parents(). Note that this does away with the previous channel-based iterators.

The proposed [implementation of RAWObjectStorage.IterType](https://github.com/scjalliance/go-git/blob/4172d931089d9851614217df7c8f57693587399e/core/object.go#L192) simply flattens the map of the requested type into a slice on the fly. This is probably fine for small to medium repositories, but *might* cause trouble for large repositories or if it's called frequently. Further optimization should be done as warranted.

Note that this pull request **modifies several public types**. As such it should probably be considered breaking for v2; it may be more appropriate to pull this instead into a v3 branch. I expect that other breaking changes will be needed to meet the needs of #19, and I suspect that all such work should be carried out in a branch until it's ready to be rolled out formally as v3. Feedback on this is most welcome.

### DIFF
--- a/commit_test.go
+++ b/commit_test.go
@@ -170,8 +170,8 @@ func (s *SuiteCommit) TestFile(c *C) {
 func makeObjectSlice(hashes []string, storage core.ObjectStorage) []core.Object {
 	series := make([]core.Object, 0, len(hashes))
 	for _, member := range hashes {
-		obj, ok := storage.Get(core.NewHash(member))
-		if ok {
+		obj, err := storage.Get(core.NewHash(member))
+		if err == nil {
 			series = append(series, obj)
 		}
 	}
--- a/commit_test.go
+++ b/commit_test.go
@@ -40,11 +40,14 @@ func (s *SuiteCommit) SetUpSuite(c *C) {
 	}
 }
 
+// FIXME: Test the new CommitIter
+/*
 func (s *SuiteCommit) TestIterClose(c *C) {
 	i := &iter{ch: make(chan core.Object, 1)}
 	i.Close()
 	i.Close()
 }
+*/
 
 var fileTests = []struct {
 	repo     string // the repo name as in localRepos
--- a/commit_test.go
+++ b/commit_test.go
@@ -1,6 +1,7 @@
 package git
 
 import (
+	"io"
 	"os"
 
 	"gopkg.in/src-d/go-git.v2/core"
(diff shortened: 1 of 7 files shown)


core/object.go
```diff
@@ -21,6 +21,7 @@ type ObjectStorage interface {
 	New() Object
 	Set(Object) Hash
 	Get(Hash) (Object, bool)
+	IterType(ObjectType) ObjectIter
```
mcuadros: I prefer call this method just `Iter`, is less verbose and clear enough

core/object.go
```diff
@@ -52,6 +53,82 @@ func (t ObjectType) Bytes() []byte {
...
 }
 
+// ObjectIter is a generic closable interface for iterating over objects.
+type ObjectIter interface {
+	Next() (Object, error)
+	Close()
+}
+
+// ObjectLookupIter yields a series of objects by retrieving each one from
+// object storage.
```
mcuadros: please explain that is an implementation of `ObjectIter`

core/object.go
```diff
@@ -52,6 +53,82 @@ func (t ObjectType) Bytes() []byte {
...
+// the end it will return io.EOF as an error. If the object is retreieved
+// successfully error will be nil.
+func (iter *ObjectLookupIter) Next() (Object, error) {
+	if iter.pos >= len(iter.series) {
+		return nil, io.EOF
+	}
+	hash := iter.series[iter.pos]
+	obj, ok := iter.storage.Get(hash)
+	if !ok {
+		// FIXME: Consider making ObjectStorage.Get return an actual error that we
```
mcuadros: Make sense, you can elaborate another PR with this change.

commit_test.go
```diff
@@ -40,11 +40,14 @@ func (s *SuiteCommit) SetUpSuite(c *C) {
 	}
 }
 
+// FIXME: Test the new CommitIter
```
mcuadros: what about this test?

mcuadros:
@JoshuaSjoding great job, we will merge this as soon as the minor things I commented get fixed. You can keep the PR to master, because we are using gopkg.in, is not a problem having an unreleased master.

Thanks!

core/object.go
JoshuaSjoding: On the one hand I like having `ObjectStorage.Get()` use the same interface as a standard map, but on the other hand we might want to report more detailed errors from an underlying file system.

Either way, I agree that it should be addressed in a separate PR. Having a functional `FileObjectStorage` implementation will also be informative.

JoshuaSjoding: In the meantime I made `ObjectLookupIter.Next()` return an `ObjectNotFoundErr` when the underlying storage provider can't return the requested object. This way it can at least be distinguished from a regular `io.EOF`.

JoshuaSjoding:
@mcuadros I think that I've addressed all of your comments with the latest commit.

`CommitIter` has some decent test coverage now. `Tree` probably needs some work and should use a `FileIter` to be consistent, but I'll handle that with a separate issue and PR if it's wanted.

I also merged in the latest work from `src-d:master` for convenience. Let me know if you'd prefer that I not do that in the future.

core/object.go
mcuadros: Make sense having a more verbose output of  `ObjectStorage.Get()` because now a simple call to it can fail in many different way if is base for example in filesystem.

JoshuaSjoding:
@mcuadros I went ahead and updated the `ObjectStorage` functions `New`, `Get` and `Set` so that they can return errors now.

PR #21, opened 2016-02-16
Add codebeat badge

marcinwyszynski:
Brilliant piece of Go code, congrats! Out of curiosity - have you tried benchmarking this against Go [libgit2 bindings](https://github.com/libgit2/git2go) running from ramdisk? For a small side project where I needed `git blame` functionality I got 10-20x speed boost from doing just that.


mcuadros:
Hi Marcin,

Thanks for the badge, we are working in a proper README right now. :dancer: 

About `git2go` in ramdisk vs `go-git`, we use go-git with some tweaks, for example we discard every binary file when we are reading the packfile so this is much faster than `git2go`, I don't have the numbers  here but we compared it some months ago. 

BTW We are working also in a [`ObjectStorage`](https://godoc.org/gopkg.in/src-d/go-git.v2/core#ObjectStorage) implementation that analyzes the files and discarded the content making the memory foot print very low. 

About the git-blame on ramdisk, our current implementation of blame is in early stages so we cannot compare it. But yes maybe will be faster run the libgit2 version.

Obviously a ramdisk is an easy solution for this problems, but dealing with big repositories becomes a memory problem if you don't discard the binaries.

marcinwyszynski:
Thanks!

borat-dancing

PR #28, opened 2016-02-16
storages: memory object

mcuadros:
New scaffolding for different storage methods also `core.RawObject` has be move to `memory.Object`. 

/cc @JoshuaSjoding


JoshuaSjoding:
@mcuadros This looks great to me. The only thing I propose changing is to rename the directory from `storages` to `storage`. Whether or not "storages" is a legitimate plural in English is probably a matter of some debate, but simply using "storage" here sounds more natural to me.

This changeset sets us up nicely for a file-based storage implementation.

Issue #31, opened 2016-02-17
Annotated Tags support


themihai:
This is a request to support annotated tags in a similar way Commits are supported. Currently there doesn't seem to be a straightforward way to pull the tag annotations(unless I'm wrong). Below are my attempts. The first one hangs (it seems there is a leak/channel blocking indefinitely) before I can do anything useful. The second one errors out with ``unexpected client error: EOF``

```
package main

import (
	"flag"
	"fmt"
	log "github.com/golang/glog"
	git "gopkg.in/src-d/go-git.v2"
	"gopkg.in/src-d/go-git.v2/clients/http"
	"io"
)

func main() {
	flag.Parse()
	defer log.Flush()
	auth := http.NewBasicAuth("myuser", ";xag]")
	r, err := git.NewRepository("https://bitbucket.org/myuser/tools", auth)
	if err != nil {
		log.Error(err)
		return
	}

	if err := r.Pull("origin", "refs/tags/0.0.1^{}"); err != nil {
		log.Error(err)
		return
	}
	triter := git.NewTreeIter(r)
	defer triter.Close()
	for {
		tree, err := triter.Next()
		if err != nil {
			if err == io.EOF {
				break
			}

			log.Error(err)
			return
		}
		log.Errorf("new tree %v", tree)
		return
	}

}
```

```
...
              
		req := &common.GitUploadPackRequest{}
		req.Have(previousTagSHA)
		req.Want(latestTagSHA)
 
		rc, err := repo.Fetch(req)
		if err != nil {
			log.Error(cx, err) 
			return nil, err
		}
```

themihai:
I ended-up using GitUploadPackRequest and set only ``req.Want(TagSHA)`` . Still there doesn't seem to be a way to fetch the annotation message alone. After reading the protocol specs it seems to be a shortfall of the git protocol rather than go-git so I'm closing this.

JoshuaSjoding:
While you can't pull just the annotated tags via the protocol, it should be possible to [clone an entire repository and then work with the tags](http://stackoverflow.com/questions/791959/download-a-specific-tag-with-git). `go-git` is currently lacking that feature.

I'll work on a PR to add annotated `Tag` support. I intend to add a `Repository.Tags()` function similar to `Repository.Commits()` that returns a `TagIter`. I'll add a `Repository.Tag()` function that retrieves a particular tag as well.

I recommend reopening this issue until the PR is accepted.

themihai:
Sounds awesome!

[...]

PR #32, opened 2016-02-19
Annotated tags

JoshuaSjoding:
This PR adds basic support for annotated tags. It provides a `Repository.Tag()` function to lookup tags by their hash, and a `Repository.Tags()` function to iterate through all of the tag objects in the repository.

It does not provide a convenient means of looking up a tag by its name. That feature can be added in the future, but its implementation doesn't seem as obvious to me right now so I'm leaving it out of this PR.

I got tired of duplicating the fixture loading code when writing tests, so I added an `unpackFixtures` function to help with that. Right now only the `Repository` and `Tag` tests make use of it.

For anyone reviewing the tag test data: the standard git tools do everything they can to hide the hash of the annotated tag objects themselves. I ended up extracting them programmatically from one of the test fixtures in order to write the test. There might be an easier way that I don't know about. Once I had the hashes I was able to grab the relevant data via the [github API](https://developer.github.com/v3/git/tags/). Sadly, github doesn't have a tag enumeration API.

Addresses #31.

### DIFF
--- a/common_test.go
+++ b/common_test.go
@@ -7,6 +7,7 @@ import (
 
 	"gopkg.in/src-d/go-git.v3/clients/common"
 	"gopkg.in/src-d/go-git.v3/core"
+	"gopkg.in/src-d/go-git.v3/formats/packfile"
 
 	. "gopkg.in/check.v1"
 )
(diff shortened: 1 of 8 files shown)


common_test.go
```diff
@@ -46,15 +47,41 @@ func (s *MockGitUploadPackService) Fetch(*common.GitUploadPackRequest) (io.ReadC
...
+			if _, existing := repos[fixture.url]; existing {
+				continue
+			}
+			repos[fixture.url] = NewPlainRepository()
+
+			d, err := os.Open(fixture.packfile)
+			c.Assert(err, IsNil)
+
+			r := packfile.NewReader(d)
+			r.Format = packfile.OFSDeltaFormat // TODO: how to know the format of a pack file ahead of time?
```
mcuadros: You can't is defined in the call git-upload-pack.

tag.go
```diff
@@ -0,0 +1,135 @@
...
+			t.object = core.NewHash(string(split[1]))
+		case "type":
+			t.Type, err = core.ParseObjectType(string(split[1]))
+			if err != nil {
+				return err
+			}
+		case "tag":
+			t.Name = string(split[1])
+		case "tagger":
+			t.Tagger.Decode(split[1])
```
mcuadros: Can we generate `Signature` objects instead of store it inline?

tag.go
```diff
@@ -0,0 +1,135 @@
...
+
+// Tree returns the tree pointed to by the tag. If the tag points to a
+// different type of object an error will be returned.
+func (t *Tag) Tree() (*Tree, error) {
+	// TODO: If the tag is of type commit, follow the commit to its tree?
+	return t.r.Tree(t.object)
+}
+
+// TODO: Add support for retrieving blobs? We don't have a type for that
+//       currently.
```
mcuadros: Blob is basically a `Object`, so just for readability we can create a type alias?

mcuadros:
Great addition this was a very relevant missing feature in the core.

tag.go
JoshuaSjoding: `Tag` does have an `Object()` function which could be used to read blob data via `Object.Reader()`, but that won't be obvious to new users without an example.

I think we should at least add a `Tag.Blob()` function that only returns a value if the object's type is actually `BlobObject`. I think that's worthwhile even if it just returns a `core.Object` or a new `core.Blob` type alias. It could be used like so:

```
func processTag(tag Tag) {
	switch tag.Type {
	case core.CommitObject:
		commit, err := tag.Commit()
		// Do something with the commit
	case core.BlobObject:
		blob, err := tag.Blob()
		// Do something with the blob
	default:
		// Something we aren't interested in
	}
}
```

I'll go ahead and add the `Tag.Blob()` function to this PR. Giving a blob a friendly name and message is a legitimate use case for annotated tags, so I think it's important to support that scenario.

We could go so far as to create a new `Blob` struct along with a `BlobIter`, but if we do I think we should put that in a separate PR.

tag.go
JoshuaSjoding: I'm not sure I follow your question. `Tag.Tagger` is a `Signature` type already, and this PR uses `Signature.Decode()` to decode that part of the tag.

tag.go
JoshuaSjoding: I just realized that there is already a `Blob` struct defined in `objects.go`. Somehow I totally missed that. I'll just use that then.

tag.go
mcuadros: Ahh yes, sorry I thought that was a `type  Tagger` sorry :D

PR #34, opened 2016-02-25
Refactor to use core.ObjectReader and core.ObjectWriter

JoshuaSjoding:
* New function signatures provide the necessary interface to stream data from disk when using filesystem-based storage in the future
* New function signatures provide proper error handling
* ObjectReader and ObjectWriter interfaces added to avoid future refactoring, currently are type aliases for io.ReadCloser and io.WriteCloser respectively
* Object.Reader now returns (ObjectReader, error)
* Object.Writer now returns (ObjectWriter, error)
* File.Contents now returns (string, error)
* File.Lines now returns ([]string, error)
* Blob.Reader now returns (core.ObjectReader, error)
* Added internal close helper function for deferred calls to Close that need to check the return value

### DIFF
--- a/blame.go
+++ b/blame.go
@@ -82,7 +82,10 @@ func (c *Commit) Blame(path string) (*Blame, error) {
 	if err != nil {
 		return nil, err
 	}
-	finalLines := file.Lines()
+	finalLines, err := file.Lines()
+	if err != nil {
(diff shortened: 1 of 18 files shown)


mcuadros:
Make sense.

PR #37, opened 2016-02-27
Added support for objfile format

JoshuaSjoding:
This changeset adds support for the loose object file format used on disk when objects aren't packed. This is a necessary component for the implementation of #19.

This changeset also fixes an accidental shadowing of the bultin close function introduced in e2c9ad1.

### DIFF
--- a/core/object.go
+++ b/core/object.go
@@ -79,6 +79,11 @@ func (t ObjectType) Bytes() []byte {
 	return []byte(t.String())
 }
 
+// Valid returns true if t is a valid ObjectType.
+func (t ObjectType) Valid() bool {
+	return t >= CommitObject && t <= REFDeltaObject
+}
+
 // ParseObjectType parses a string representation of ObjectType. It returns an
 // error on parse failure.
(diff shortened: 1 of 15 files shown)


core/object.go
```diff
@@ -79,6 +79,11 @@ func (t ObjectType) Bytes() []byte {
 	return []byte(t.String())
 }
 
+// Valid returns true if t is a valid ObjectType.
+func (t ObjectType) Valid() bool {
```
mcuadros: Why? This can happen?

mcuadros:
Awesome work. Looks greats.

core/object.go
JoshuaSjoding: I meant to use this to check the validity of the provided object type in objfile.NewWriter. I'll update the PR.

Object types default to 0, which is an invalid value, so I added this function for a consistent way to check that someone didn't forget to provide a valid value.

JoshuaSjoding: objfile.NewWriter has been updated to use this function now.

PR #39, opened 2016-03-03
Added Object interface for Commit, Tree, Blob and Tag

JoshuaSjoding:
* New Object interface is distinct from core.Object
* New Object interface is used in places where returned object could be of any type
* Object is implemented by Commit, Tree, Blob, File and Tag
* Added Repository.Object function for retrieving objects of any type
* Tag.Object now returns Object instead of core.Object
* Tag target hash is now publicly accessible
* Renamed Tag.Type field to Tag.TargetType, making it distinct from Tag.Type function
* Fixed infinite recursive loop in TagIter.Close
* TreeWalker.Next now returns Object instead of core.Object
* Removed some duplicate test setup code

### DIFF
--- a/commit.go
+++ b/commit.go
@@ -43,13 +43,28 @@ func (c *Commit) NumParents() int {
 }
 
 // File returns the file with the specified "path" in the commit and a
-// nil error if the file exists. If the file does not exists, it returns
+// nil error if the file exists. If the file does not exist, it returns
 // a nil file and the ErrFileNotFound error.
 func (c *Commit) File(path string) (file *File, err error) {
 	return c.Tree().File(path)
 }
 
-// Decode transform an core.Object into a Blob struct
+// ID returns the object ID of the commit. The returned value will always match
+// the current value of Commit.Hash.
+//
+// ID is present to fufill the Object interface.
+func (c *Commit) ID() core.Hash {
+	return c.Hash
+}
+
+// Type returns the type of object. It always returns core.CommitObject.
+//
+// Type is present to fufill the Object interface.
+func (c *Commit) Type() core.ObjectType {
+	return core.CommitObject
+}
+
+// Decode transforms a core.Object into a Commit struct.
 func (c *Commit) Decode(o core.Object) (err error) {
 	if o.Type() != core.CommitObject {
 		return ErrUnsupportedObject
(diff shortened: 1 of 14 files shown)


objects.go
```diff
@@ -18,9 +48,22 @@ type Blob struct {
 	obj core.Object
 }
 
-var ErrUnsupportedObject = errors.New("unsupported object type")
+// ID returns the object ID of the blob. The returned value will always match
+// the current value of Blob.Hash.
+//
+// ID is present to fufill the Object interface.
+func (b *Blob) ID() core.Hash {
```
JoshuaSjoding: I really wanted to call this function `Hash()` so that the name matched `core.Object`'s equivalent, but that would have conflicted with the existing `Hash` field on `Commit`, `Tree`, `Blob` and `Tag`.

objects.go
```diff
@@ -10,6 +10,36 @@ import (
...
+//   		// o is a Blob
+//   	case *Tag:
+//   		// o is a Tag
+//   	}
+//   }
+//
+// This interface is intentionally different from core.Object, which is a lower
+// level interface used by storage implementations to read and write objects.
+type Object interface {
+	ID() core.Hash
```
JoshuaSjoding: I really wanted to call this function `Hash()` so that the name matched `core.Object`'s equivalent, but that would have conflicted with the existing `Hash` field on `Commit`, `Tree`, `Blob` and `Tag`.

commit.go
```diff
@@ -107,23 +122,33 @@ func (c *Commit) String() string {
...
 	return &CommitIter{iter, r}
 }
 
+// Next moves the iterator to the next commit and returns a pointer to it. If it
+// has reached the end of the set it will return io.EOF.
 func (iter *CommitIter) Next() (*Commit, error) {
-	obj, err := iter.ObjectIter.Next()
-	if err != nil {
-		return nil, err
+	for {
```
mcuadros: why a loop? ObjectStorage has a index of commits to avoid this

JoshuaSjoding: I added it mostly as a fail-safe, in case someone creates their own ObjectIter that doesn't already filter for the desired type.

Perhaps it would be better instead to create an implementation of ObjectIter that filters for a given type? Then someone could wrap iterators like so:

```
var source ObjectIter // some source of mixed objects
filtered := NewObjectTypeFilter(source, CommitObject)
commits := NewCommitIter(r, filtered)
```

Such a filter could be added in a later commit. In any case, I'm okay with omitting the loop here and in the other iterators if you think it best to keep them simple.

mcuadros: Yep, I prefer remove it, adds complexity and you will get a error trying to recover a non-commit as commit. so the case of a iteration over a non-commits is cover.

JoshuaSjoding: PR has been updated.

Issue #45, opened 2016-05-07
Panic in the example to print commits


TripleDogDare:
Trying to run the example code seems to give an error.
It appears that the `remote.upInfo` isn't getting set.
This can be resolved by inserting a `Connect` call on the default remote. 
This can also be resolved by modifying the `Pull` `PullDefault` code.
Options are in referenced pull requests.

FYI: I'm not sure how to run tests locally (they are failing) so I hope your CI works...

```
$ cd $GOPATH
$ go get -u github.com/src-d/go-git
$ go install github.com/src-d/go-git/examples/basic
$ ./bin/basic 'https://github.com/src-d/go-git'
Retrieving "https://github.com/src-d/go-git" ...
panic: runtime error: invalid memory address or nil pointer dereference
[signal 0xb code=0x1 addr=0x0 pc=0x464005]

goroutine 1 [running]:
gopkg.in/src-d/go-git%2ev3.(*Remote).DefaultBranch(0xc20803a140, 0x0, 0x0)
	/home/tripledogdare/go/src/gopkg.in/src-d/go-git.v3/remote.go:81 +0x35
gopkg.in/src-d/go-git%2ev3.(*Repository).PullDefault(0xc20803d530, 0x0, 0x0)
	/home/tripledogdare/go/src/gopkg.in/src-d/go-git.v3/repository.go:101 +0x1a6
main.main()
	/home/tripledogdare/go/src/github.com/src-d/go-git/examples/basic/main.go:18 +0x1aa

goroutine 2 [runnable]:
runtime.forcegchelper()
	/home/tripledogdare/repositories/go/src/runtime/proc.go:90
runtime.goexit()
	/home/tripledogdare/repositories/go/src/runtime/asm_amd64.s:2232 +0x1

goroutine 3 [runnable]:
runtime.bgsweep()
	/home/tripledogdare/repositories/go/src/runtime/mgc0.go:82
runtime.goexit()
	/home/tripledogdare/repositories/go/src/runtime/asm_amd64.s:2232 +0x1

goroutine 4 [runnable]:
runtime.runfinq()
	/home/tripledogdare/repositories/go/src/runtime/malloc.go:712
runtime.goexit()
	/home/tripledogdare/repositories/go/src/runtime/asm_amd64.s:2232 +0x1

```

mcuadros:
Thanks

PR #46, opened 2016-05-07
Fix panic in the basic example code

TripleDogDare:
The `remote.upInfo` for origin isn't set before it used in PullDefault.
Calling `Connect` on the remote before calling `PullDefault` fixes the problem.
Relates to Issue #45 

This is not required if you accept #47

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,6 +45,10 @@ if err != nil {
 	panic(err)
 }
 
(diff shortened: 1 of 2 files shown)


mcuadros:
You other PR fix the README.md. Thanks

PR #52, opened 2016-06-09
add "local" remotes

alcortesm:
By using the "dir://" scheme, remotes from a local ".git" repository are now allowed.


README.md
```diff
@@ -87,6 +87,29 @@ Date:   2015-12-11 17:57:10 +0100 +0100
 ...
 ```
 
+Bare and non-bare local repositories are also supported using the `dir://`
```
mcuadros: what about using `local://` as scheme for the local remotes?

blame_test.go
```diff
@@ -25,13 +25,13 @@ func (s *BlameCommon) SetUpSuite(c *C) {
 		d, err := os.Open(fixRepo.packfile)
 		c.Assert(err, IsNil)
 
-		r := packfile.NewReader(d)
+		decoder := packfile.NewDecoder(d)
```
mcuadros: I prefer keep short names to short scope variables

clients/common.go
```diff
@@ -16,6 +16,7 @@ import (
 	"net/url"
 
 	"gopkg.in/src-d/go-git.v3/clients/common"
+	"gopkg.in/src-d/go-git.v3/clients/file"
```
mcuadros: really we  must to choose, "local", "dir" or "file" as client name.

clients/common_test.go
```diff
@@ -3,18 +3,36 @@ package clients
...
 func Test(t *testing.T) { TestingT(t) }
 
-type SuiteCommon struct{}
+type SuiteCommon struct {
+	dirFixturePath string
+}
 
 var _ = Suite(&SuiteCommon{})
 
+const fixtureTGZ = "../formats/gitdir/fixtures/spinnaker-gc.tgz"
```
mcuadros: gitdir, call the dir as the name of the client

README.md
alcortesm: `local://` will be just perfect.

blame_test.go
alcortesm: OK, will do.

clients/common.go
alcortesm: ok, local would be perfect if we are going to use `local://` as the scheme.

alcortesm: Altough, it is not really a client as there is no server, but that is a different problem we must solve :).

clients/common_test.go
alcortesm: OK, will do.

alcortesm:
I'll remove go 1.5 from the CI, so the travis don't complain anymore about my builds.

sona-tar:
file:// scheme is more common than the local:// scheme.

https://en.m.wikipedia.org/wiki/File_URI_scheme
https://docs.oracle.com/javase/8/docs/api/java/io/File.html

And we can use file:// scheme as link on Windows.

alcortesm:
You are right @sona-tar , thanks for the advice!

sona-tar:
@alcortesm Thanks!! I'm looking forward to merging file scheme to master.

alcortesm:
As per instructed by @mcuadros I'm closing this PR.

All the features in this PR will be included in a future PR along with some new features a quite an important refactor.

PR #53, opened 2016-06-12
[WIP] Python

vmarkovtsev:
This is an ongoing effort to wrap go-git with Python.

### Internals
1) `go build -o libgogit.so -buildmode=c-shared github.com/src-d/go-git/cshared` makes a shared library for general usage
2) Python uses cffi <- libgogit.h
3) Go and Python interfaces must be written manually

### Goals
1) Python 2.7 and 3.3+ support
2) Cover 95% of go-git API
3) 100% test coverage

### What currently works
See README.python.md
Basically, what's already wrapped should work. I am not using TDD here for several reasons so tests will be written in the end.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,2 @@
+.idea
+__pycache__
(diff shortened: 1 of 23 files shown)


mcuadros:
Please make the PR to https://github.com/src-d/gypogit

PR #54, opened 2016-06-14
Add cshared files to allow building wrappers in other languages

vmarkovtsev:
cgo requires the package to be named exactly "main". This sucks but we have to live with this so I add `// +build ignore` in order not to mess up during ordinary go-git build.

Build the library with `go build -o libgogit.so -buildmode=c-shared github.com/src-d/go-git/cshared`

### DIFF
--- a/cshared/auth_method_cshared.go
+++ b/cshared/auth_method_cshared.go
@@ -1,0 +1,192 @@
+// +build ignore
+package main
+
+import (
+	"C"
+	"strings"
+
+	"golang.org/x/crypto/ssh"
+	"gopkg.in/src-d/go-git.v3/clients/http"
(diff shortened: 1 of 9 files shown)


alcortesm:
LGTM.

cshared/auth_method_cshared.go
```diff
@@ -0,0 +1,192 @@
...
+package main
+
+import (
+	"C"
+	"reflect"
+	"strings"
+
+	"golang.org/x/crypto/ssh"
+	"gopkg.in/src-d/go-git.v3/clients/http"
+	. "gopkg.in/src-d/go-git.v3/clients/ssh"
```
mcuadros: Can you remove the dot import, is a bad practice in Go -> https://github.com/golang/go/wiki/CodeReviewComments#import-dot

Issue #51, opened 2016-06-04
Support to open local repository or file transport protocol.


sona-tar:
Hello
I have some questions.

How can I open local repository or file transport protocol? 
If not supported now, do you have a plan to support it?

I'd like to use go-git instead of https://github.com/gogits/git.
See https://github.com/gogits/git/issues/17

Thanks.

oplibs:
Hello. Can I init a local repositories by this library? I need implement some functions for multi-users. Init their repositories would be needed for me. If not, any plans?

sona-tar:
Thanks to support local repository!! 

But I want to open repository from HEAD.

```
hash, _ := repo.Head("")

for commit := range repo.Commits(hash) {
 fmt.Println(commit)
}
```

Now you supported to packfile but not supported non packed object `.git/objects/HASH[0:2]/HASH[2:64]`.
Do you have a plan to support non packed object ?

mcuadros:
we don't support spare objects right now, you can save this issues running `git gc` before open it with go-git

mcuadros:
@oplibs we are working on this right now, this is on the branch v4

PR #55, opened 2016-06-17
Adds support to open local repositories and to use file-based object storage

alcortesm:
This patch addresses https://github.com/src-d/go-git/issues/51 and https://github.com/src-d/go-git/issues/19.

Local repositories can now be opened by using the `NewRepositoryFromFS(path string)` repository constructor.

The default storage for such repositories is just a proxy to the original repository, meaning no git objects are stored on memory and all queries to the repository are resolved by accessing the original repository packfile and idx file in its ".git" directory.


alcortesm:
Please, don't merge yet, i just realize support for processing REF deltas when creating the index from raw packfiles is not supported yet. Give a few days to finish this.

themihai:
@alcortesm  Could NewRepositoryFromFS could accept an interface instead of a file path?  That way we can use a remote service (e.g. amazon S3) as local storage 

_Not affiliated with the project._

alcortesm:
Ready to merge, IMHO.

alcortesm:
Got some comments from @mcuadros IRL, implementing them right now.

PR #68, opened 2016-08-17
Support non packed objects

sona-tar:
Support non packed objects for v4 branch.

Enable to get objects from `.git/objects/HASHDIR/HASHFILE` in `func (s *ObjectStorage) Get(h core.Hash) (core.Object, error)`.
Now not support Iterator in `func (s *ObjectStorage) Iter(t core.ObjectType) (core.ObjectIter, error)` , but we should support iterator for `func (r *Repository) Commits() (*CommitIter, error)`.
I'm considering how to implement iterator.

#51


sona-tar:
https://github.com/src-d/go-git/pull/68/commits/70741202d6c50f285c529b54d347a66db2cb1e6b support iterator. but it has a problem.

`Commits()` iterator isn't sorted.

mcuadros:
@alcortesm PTAL

formats/objfile/reader.go
```diff
@@ -121,3 +121,12 @@ func (r *Reader) Close() (err error) {
 
 	return
 }
+
+func (r *Reader) FillObject(obj core.Object) error {
+	obj.SetType(r.header.t)
+	obj.SetSize(r.header.size)
+	w, _ := obj.Writer()
```
alcortesm: Please, don't ignore errors.

formats/objfile/reader.go
```diff
@@ -121,3 +121,12 @@ func (r *Reader) Close() (err error) {
...
+	_, err := io.Copy(w, r.r)
+	return err
+
```
alcortesm: Style: swap line 130 and line 131 to adapt your code to our style.

storage/filesystem/internal/dotgit/dotgit.go
```diff
@@ -51,9 +53,7 @@ func New(fs fs.FS, path string) (*DotGit, error) {
 // Symbolic references are resolved and included in the output.
 func (d *DotGit) Refs() ([]*core.Reference, error) {
 	var refs []*core.Reference
-	if err := d.addRefsFromPackedRefs(&refs); err != nil {
-		return nil, err
-	}
+	_ = d.addRefsFromPackedRefs(&refs)
```
alcortesm: We cannot ignore errors, think of using a database to implement the underlying file system.

storage/filesystem/internal/dotgit/dotgit.go
```diff
@@ -103,3 +103,52 @@ func (d *DotGit) Idxfile() (fs.FS, string, error) {
 
 	return nil, "", ErrIdxNotFound
 }
+
+func (d *DotGit) Objectfiles() (fs.FS, []core.Hash, error) {
```
alcortesm: This function is public, add a proper comment for GoDoc.

formats/objfile/reader.go
```diff
@@ -121,3 +121,12 @@ func (r *Reader) Close() (err error) {
 
 	return
 }
+
+func (r *Reader) FillObject(obj core.Object) error {
```
alcortesm: This method is public, add comments for Go Doc.

Also add some tests for this method.

storage/filesystem/internal/dotgit/dotgit.go
alcortesm: Also add tests for this method.

storage/filesystem/internal/dotgit/dotgit.go
```diff
@@ -103,3 +103,52 @@ func (d *DotGit) Idxfile() (fs.FS, string, error) {
...
+				fmt.Println(name)
+				objects = append(objects, core.NewHash(name))
+			}
+		}
+	}
+
+	return d.fs, objects, nil
+}
+
+func (d *DotGit) Objectfile(h core.Hash) (fs.FS, string, error) {
```
alcortesm: Add comments for Go Doc and tests.

storage/filesystem/internal/dotgit/dotgit.go
```diff
@@ -103,3 +103,52 @@ func (d *DotGit) Idxfile() (fs.FS, string, error) {
...
+	hash := h.String()
+	objFile := d.fs.Join(d.path, "objects", hash[0:2], hash[2:40])
+
+	if _, err := d.fs.Stat(objFile); err != nil {
+		if os.IsNotExist(err) {
+			return nil, "", ErrNotFound
+		}
+		return nil, "", err
+	}
+	return d.fs, objFile, nil
```
alcortesm: Style: add an empty line between the close bracket of the if and the return clause.

storage/filesystem/internal/dotgit/dotgit.go
```diff
@@ -103,3 +103,52 @@ func (d *DotGit) Idxfile() (fs.FS, string, error) {
 
 	return nil, "", ErrIdxNotFound
 }
+
+func (d *DotGit) Objectfiles() (fs.FS, []core.Hash, error) {
+	dotGitobjcts := d.fs.Join(d.path, "objects")
```
alcortesm: There is an 'e' missing from this variable name and the second 'o' must be in uppercase:

dotGitobjcts -> dotGitObjects

Maybe just rename it to objsDir instead, it will be more coherent with the rest of the code in this file.

storage/filesystem/internal/dotgit/dotgit.go
```diff
@@ -103,3 +103,52 @@ func (d *DotGit) Idxfile() (fs.FS, string, error) {
...
+	files, err := d.fs.ReadDir(dotGitobjcts)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	var objDirs []string
+	reDir, _ := regexp.Compile("[a-z0-9]{2}")
```
alcortesm: Use regexp.MustCompile instead of regexp.Compile.

Also move the resulting regexp to a package level variable an initialize it in the package init.

alcortesm: The regexp is wrong in several ways:

- wrong characters: characters like `g` or `z` are allowed, while the only characters allowed for the directory names are 0-9 and a-f.

- The refgexp matches non valid directory names like '000' or 'pack' (as 'ac' is valid). What is the point of having a regexp then?

alcortesm: Also: using regexp is quite expensive for the CPU while the problem is quite simple, is there a faster solution?

storage/filesystem/internal/dotgit/dotgit.go
```diff
@@ -103,3 +103,52 @@ func (d *DotGit) Idxfile() (fs.FS, string, error) {
...
+	for _, f := range files {
+		if f.IsDir() && reDir.MatchString(f.Name()) {
+			objDirs = append(objDirs, f.Name())
+		}
+	}
+
+	var objects []core.Hash
+	reObj, _ := regexp.Compile("[a-z0-9]{38}")
```
alcortesm: the same as above.

storage/filesystem/internal/dotgit/dotgit.go
```diff
@@ -103,3 +103,52 @@ func (d *DotGit) Idxfile() (fs.FS, string, error) {
...
+	for _, dir := range objDirs {
```
alcortesm: why are we looping twice? one to fill an slice of dirs, and then again to fill the slice of objects.

We don't really need the slice of objDirs, just visit the dirs if they match the regexp and fill the objects slice inplace.

storage/filesystem/internal/dotgit/dotgit.go
```diff
@@ -103,3 +103,52 @@ func (d *DotGit) Idxfile() (fs.FS, string, error) {
...
+		objs, err := d.fs.ReadDir(d.fs.Join(dotGitobjcts, dir))
+
```
alcortesm: remove this empty line.

storage/filesystem/internal/dotgit/dotgit.go
```diff
@@ -103,3 +103,52 @@ func (d *DotGit) Idxfile() (fs.FS, string, error) {
...
+		if err != nil {
+			return nil, nil, err
+		}
+
+		for _, obj := range objs {
+			if reObj.MatchString(obj.Name()) {
+				name := dir + obj.Name()
+				fmt.Println(name)
```
alcortesm: The library must not print anything to stdout.

alcortesm:
Thanks for your contribution @sona-tar, There are some issues in you PR, I have leave you some comments about them:

 - minor style changes
 - bugs in your code (wrong regexps)
 - it is missing tests
 - and I think you have break the `FsSuite.TestIterCompareWithMemoryStorage` test.

storage/filesystem/object.go
```diff
@@ -36,9 +37,66 @@ func (s *ObjectStorage) Set(core.Object) (core.Hash, error) {
 	return core.ZeroHash, fmt.Errorf("not implemented yet")
 }
 
+func (s *ObjectStorage) Get(h core.Hash) (core.Object, error) {
```
alcortesm: Add a comment to this public method, so go doc can know about it.

storage/filesystem/object.go
```diff
@@ -36,9 +37,66 @@ func (s *ObjectStorage) Set(core.Object) (core.Hash, error) {
...
+	var obj core.Object
+	var err error
+
+	obj, err = s.getFromObject(h)
```
alcortesm: the function `getFromObject` needs a better name, it returns an object from an object?

storage/filesystem/object.go
```diff
@@ -36,9 +37,66 @@ func (s *ObjectStorage) Set(core.Object) (core.Hash, error) {
...
+	if err == nil {
```
alcortesm: It is ok to look for the object in the packfile if the object has not been found in the objects dir, but what if getFromObject fails for other reasons? (for example if we don't have the permissions or because the connection with the (remote) file system has fail).

We must do proper error checking of the return values of getFromObject, defining specific error for when the object has not been found.

storage/filesystem/object.go
alcortesm: Style: Remove the var definitions at 41 and 42 and use a `:=` here.

storage/filesystem/object.go
```diff
@@ -36,9 +37,66 @@ func (s *ObjectStorage) Set(core.Object) (core.Hash, error) {
...
+	objReader, err := objfile.NewReader(f)
+	if err != nil {
+		return nil, err
+	}
+	err = objReader.FillObject(obj)
+	if err != nil {
+		return nil, err
+	}
+
+	err = objReader.Close()
```
alcortesm: defer this close, so we don't forget to close objReader in the future when someone changes this code.

storage/filesystem/object.go
```diff
@@ -36,9 +37,66 @@ func (s *ObjectStorage) Set(core.Object) (core.Hash, error) {
...
+	if err == nil {
+		return obj, nil
+	}
+
+	return nil, err
+}
+
+// Get returns the object with the given hash, by searching for it in
+// the git object directories.
+func (s *ObjectStorage) getFromObject(h core.Hash) (core.Object, error) {
```
alcortesm: You must name your return values here, otherwise the defer function cannot overwrite the error value.

storage/filesystem/object.go
```diff
@@ -36,9 +37,66 @@ func (s *ObjectStorage) Set(core.Object) (core.Hash, error) {
...
+	}
+
+	err = objReader.Close()
+	if err != nil {
+		return nil, err
+	}
+	return obj, nil
+}
+
 // Get returns the object with the given hash, by searching for it in
```
alcortesm: This comment is wrong now.

storage/filesystem/object.go
```diff
@@ -36,9 +37,66 @@ func (s *ObjectStorage) Set(core.Object) (core.Hash, error) {
...
 // the packfile.
-func (s *ObjectStorage) Get(h core.Hash) (core.Object, error) {
+func (s *ObjectStorage) getFromPackfile(h core.Hash) (core.Object, error) {
+	if s.index == nil {
+		return nil, dotgit.ErrIdxNotFound
```
alcortesm: This error is not the correct error to return, as we have not search for an Idx, we cannot say that the Idx has not been found.

storage/filesystem/object.go
```diff
@@ -79,8 +137,23 @@ func (s *ObjectStorage) Get(h core.Hash) (core.Object, error) {
...
+			return nil, err
+		}
+		if object.Type() == t {
+			objects = append(objects, object)
+		}
+	}
+
 	for hash := range s.index {
-		object, err := s.Get(hash)
+		object, err := s.getFromPackfile(hash)
```
alcortesm: Creating an slice of decoded objects and then returning an iterator over it is very inefficient and wastes a lot of memory, it will be much better if the iterator would be able to decode each object on demand with each Next() invocation.

This is not your fault, it is something we will change in the future.

alcortesm:
It is ok if the commit iterator does not return commits sorted by Commiter.When.

storage/filesystem/object.go
sona-tar: > as we have not search for an Idx, we cannot say that the Idx has not been found.

I think that idx file had been searched by `buildIndex()` in `ObjectStorage()`, also if idx file found, `s.index` created before `getFromPackfile()` called.

If It's wrong, please recommend better error.

sona-tar:
Thanks for your review.

> minor style changes
> bugs in your code (wrong regexps)
> it is missing tests
> and I think you have broken the FsSuite.TestIterCompareWithMemoryStorage test.

I fixed it for https://github.com/src-d/go-git/pull/68/commits/25e4a41cdf8980cc8ed97c15880b3d5ae25ea75f and https://github.com/src-d/go-git/pull/68/commits/d346525b08eca8491b43a36601d19198ddd5dec4 .

> Also, using regexp can be quite CPU expensive for such a simple task, we are not sure we can afford to waste so much CPU.
I didn't fix regexp problem.

It is difficult to ignore illegal object. like `objects/gg/46525b08eca8491b43a36601d19198ddd5dec4`.

If we need not ignore illegal object directory. It is only necessary to check object directory name size which is 2 character, and we require all object files in object directory.

```diff
- if f.IsDir() && objDirRegExp.MatchString(f.Name()) {
+ if f.IsDir() && len(f.Name()) == 2 {

- if objFileRegExp.MatchString(obj.Name()) {
```

Could you give me some advice?

alcortesm:
> Could you give me some advice?

Of course.

```
if f.IsDir() && len(f.Name) == 2 && isHex(f.Name) {
...
}

func isHex(s string) bool {
	for _, b := range []byte(s) {
		if isNum(b) {
			continue
		}
		if isHexAlpha(b) {
			continue
		}

		return false
	}

	return true
}

func isNum(b byte) bool {
	return b >= '0' && b <= '9'
}

func isHexAlpha(b byte) bool {
	return b >= 'a' && b <= 'f' || b >= 'A' && b <= 'F' // not sure if uppercase is allowed by git
}
```

This is at least 10 times faster than a regexp, according to my benchmarks.

sona-tar:
Thanks for advice!! 
I fixed it. 😄 

Please let me know if there is a problem.

sona-tar:
Should I rebase v4 and resolve conflicts?

storage/filesystem/internal/dotgit/dotgit_test.go
```diff
@@ -77,6 +119,8 @@ func (s *SuiteDotGit) SetUpSuite(c *C) {
 
 		f.packfile = init.packfile
 		f.idxfile = init.idxfile
+		f.objectfiles = make([]fixtureObject, len(init.objectfiles))
+		copy(f.objectfiles, init.objectfiles)
```
alcortesm: Don't copy the slice at init.objfiles, just reference it.

storage/filesystem/internal/dotgit/dotgit.go
```diff
@@ -103,3 +106,70 @@ func (d *DotGit) Idxfile() (fs.FS, string, error) {
 
 	return nil, "", ErrIdxNotFound
 }
+
+// Objectfiles returns the hash slice of the objects files in the "objects/"
```
alcortesm: Grammar improvement: Objectfiles returns a slice with the hashes of objects found under the .git/objects/ directory.

storage/filesystem/internal/dotgit/dotgit.go
```diff
@@ -103,3 +106,70 @@ func (d *DotGit) Idxfile() (fs.FS, string, error) {
...
+
+	files, err := d.fs.ReadDir(objsDir)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	var objects []core.Hash
+	for _, f := range files {
+		if f.IsDir() && len(f.Name()) == 2 && isHex(f.Name()) {
+			objDir := f.Name()
```
alcortesm: The names of the vars `objDir` and `objsDir` are too similar and it is confusing.

Can you rename `objDir` as `d`? (that way it would be coherent with the `f` var naming and quite diferent form `objsDir`. You can rename the receiver as `dg`, for example.

storage/filesystem/internal/dotgit/dotgit.go
```diff
@@ -103,3 +106,70 @@ func (d *DotGit) Idxfile() (fs.FS, string, error) {
...
+			objs, err := d.fs.ReadDir(d.fs.Join(objsDir, objDir))
+			if err != nil {
+				return nil, nil, err
+			}
+
+			for _, obj := range objs {
```
alcortesm: The same as above, probably `o` is just a better name for `obj` here.

storage/filesystem/internal/dotgit/dotgit.go
```diff
@@ -103,3 +106,70 @@ func (d *DotGit) Idxfile() (fs.FS, string, error) {
...
+func isNum(b byte) bool {
+	return b >= '0' && b <= '9'
+}
+
+func isHexAlpha(b byte) bool {
+	return b >= 'a' && b <= 'f' || b >= 'A' && b <= 'F'
+}
+
+// Objectfile returns the path of the object file (really, it returns the
+// path of the first file in the "objects/hash[0:2]/"
```
alcortesm: Can you be a little more specific about what this function does? Example: `Objectfile returns the path of the object file for a given hash *if the file exists*, otherwise returns an ErrObjfileNotFound error.`

Please remove the `(really, it returns the path of the first file in the "objects/hash[0:2]/"`, *it is wrong*, badly formatted (a `)` is missing) and it is a leftover from a too evident copy and paste operation.

storage/filesystem/internal/dotgit/dotgit_test.go
```diff
@@ -216,7 +260,92 @@ func (s *SuiteDotGit) TestPackfile(c *C) {
 	}
 }
 
+func (s *SuiteDotGit) TestObjectfiles(c *C) {
+	objectfiles := func(d *DotGit) (fs.FS, []core.Hash, error) {
```
alcortesm: I don't understand the point of this variable `objectfiles`.

Isn't the purpose of this method to test `DotGit.Objectfiles`?, then why are you wrapping it with the `objectfiles` variable?

Just call `DotGit.Objectfiles` directly on the tests, otherwise it is too convoluted.

storage/filesystem/internal/dotgit/dotgit_test.go
```diff
@@ -216,7 +260,92 @@ func (s *SuiteDotGit) TestPackfile(c *C) {
...
+		return d.Objectfiles()
+	}
+	for _, test := range [...]struct {
+		fixture string
+		fn      getObjectsHashesFn
```
alcortesm: Same as above, why do you need this field?, can you just call the Objecfiles method directly?

storage/filesystem/internal/dotgit/dotgit_test.go
```diff
@@ -216,7 +260,92 @@ func (s *SuiteDotGit) TestPackfile(c *C) {
...
+			c.Assert(len(hashes), Equals, len(fix.objectfiles), com)
+			for i, hash := range hashes {
+				c.Assert(hash.String(), Matches, fix.objectfiles[i].hash, com)
+			}
+		}
+	}
+}
+
+func (s *SuiteDotGit) TestObjectfile(c *C) {
+	objectfile := func(d *DotGit, h core.Hash) (fs.FS, string, error) {
```
alcortesm: same as above.

storage/filesystem/internal/dotgit/dotgit_test.go
```diff
@@ -216,7 +260,92 @@ func (s *SuiteDotGit) TestPackfile(c *C) {
...
+		return d.Objectfile(h)
+	}
+	for _, test := range [...]struct {
+		fixture string
+		fn      getObjectPathFn
```
alcortesm: same as above

storage/filesystem/internal/dotgit/dotgit_test.go
```diff
@@ -216,7 +260,92 @@ func (s *SuiteDotGit) TestPackfile(c *C) {
...
+			fixture: "no-packfile-no-idx",
+			fn:      objectfile,
+			err:     "object file not found",
+		},
+	} {
+		com := Commentf("fixture = %s", test.fixture)
+
+		fix, dir := s.newFixtureDir(c, test.fixture)
+
+		for _, fixobj := range fix.objectfiles {
```
alcortesm: we use camel case: `fixobj` -> `fixObj`

storage/filesystem/internal/dotgit/dotgit_test.go
```diff
@@ -216,7 +260,92 @@ func (s *SuiteDotGit) TestPackfile(c *C) {
...
+}
+
+func (s *SuiteDotGit) TestObjectfile(c *C) {
+	objectfile := func(d *DotGit, h core.Hash) (fs.FS, string, error) {
+		return d.Objectfile(h)
+	}
+	for _, test := range [...]struct {
+		fixture string
+		fn      getObjectPathFn
+		err     string // error regexp
```
alcortesm: You have defined a new error for objectfile not found (`ErrObjfileNotFound`), will it be better to compare the var instead of using a regexp?

storage/filesystem/internal/dotgit/dotgit_test.go
```diff
@@ -216,7 +260,92 @@ func (s *SuiteDotGit) TestPackfile(c *C) {
...
+				c.Assert(strings.HasSuffix(path, fixobj.path),
+					Equals, true, com)
+			}
+		}
+	}
+}
+
 type getPathFn func(*DotGit) (fs.FS, string, error)
+type getObjectPathFn func(*DotGit, core.Hash) (fs.FS, string, error)
+type getObjectsHashesFn func(d *DotGit) (fs.FS, []core.Hash, error)
```
alcortesm: Same as above, these two lines doesn't make sense to me.

storage/filesystem/internal/dotgit/refs.go
```diff
@@ -32,10 +32,9 @@ func (d *DotGit) addRefsFromPackedRefs(refs *[]*core.Reference) (err error) {
 	path := d.fs.Join(d.path, packedRefsPath)
 	f, err := d.fs.Open(path)
 	if err != nil {
-		if err == os.ErrNotExist {
+		if os.IsNotExist(err) {
```
alcortesm: good catch, thanks!

storage/filesystem/object.go
```diff
@@ -37,8 +38,67 @@ func (s *ObjectStorage) Set(core.Object) (core.Hash, error) {
...
+	if err != dotgit.ErrObjfileNotFound {
+		return nil, err
+	}
+
+	obj, err = s.getFromPackfile(h)
+	if err == nil {
+		return obj, nil
+	}
+
+	return nil, err
```
alcortesm: substitute lines 51 to 56 by:

```
    return s.getFromPackfile(h)
```

storage/filesystem/object.go
```diff
@@ -37,8 +38,67 @@ func (s *ObjectStorage) Set(core.Object) (core.Hash, error) {
...
+			err = errClose
+		}
+	}()
+
+	err = objReader.FillObject(obj)
+	if err != nil {
+		return nil, err
+	}
+
+	return obj, nil
```
alcortesm: Change lines 89-94 to:

```
    return obj, objReader.FillObject(obj)
```

storage/filesystem/object.go
```diff
@@ -37,8 +38,67 @@ func (s *ObjectStorage) Set(core.Object) (core.Hash, error) {
...
+}
+
+func (s *ObjectStorage) getFromPackfile(h core.Hash) (obj core.Object, err error) {
+	if s.index == nil {
+		return nil, dotgit.ErrIdxNotFound
```
alcortesm: I'm not sure about this error, it is misleading, the problem is that there is no index, but the errors says there is no idx file, which is different. Maybe it is better to return a new error to be true to the real problem instead of reusing dotgit.ErrIdxNotFound here.

alcortesm:
https://github.com/src-d/go-git/pull/68#issuecomment-241699379:

> Should I rebase v4 and resolve conflicts?


Yes please.

alcortesm:
This PR is starting to look good, I've left your some more comments.

storage/filesystem/internal/dotgit/dotgit.go
sona-tar: Sound good!

I fixed it.

storage/filesystem/object.go
sona-tar: 1.  Change return value to core.ErrObjectNotFound  in `Get()` like following line
https://github.com/src-d/go-git/blob/v4/storage/memory/storage.go#L136

2. Remove L98-L100 in `getFromPackfile()`

Please give me your comment.

sona-tar:
I rebased to v4.

Can I fixup this PR commits to one commit?
rebase is so hard...

PR #69, opened 2016-08-18
storage: Add object type hint parameter to ObjectStorage.Get.

smola:
Some storage backends can optimize object lookup if they get
the object type that is expected. So we the signature of the Get
method is now Get(Hash, ObjectType).

### DIFF
--- a/core/object.go
+++ b/core/object.go
@@ -38,17 +38,22 @@ type Object interface {
 	Writer() (ObjectWriter, error)
 }
 
-// ObjectType internal object type's
+// ObjectType internal object type
+// Integer values from 0 to 7 map to those exposed by git.
+// AnyObject is used to represent any from 0 to 7.
 type ObjectType int8
 
 const (
-	InvalidObject  ObjectType = 0
-	CommitObject   ObjectType = 1
-	TreeObject     ObjectType = 2
-	BlobObject     ObjectType = 3
-	TagObject      ObjectType = 4
+	InvalidObject ObjectType = 0
+	CommitObject  ObjectType = 1
+	TreeObject    ObjectType = 2
+	BlobObject    ObjectType = 3
+	TagObject     ObjectType = 4
+	// 5 reserved for future expansion
 	OFSDeltaObject ObjectType = 6
 	REFDeltaObject ObjectType = 7
+
+	AnyObject ObjectType = -127
 )
 
 func (t ObjectType) String() string {
@@ -132,7 +137,7 @@ func (iter *ObjectLookupIter) Next() (Object, error) {
 		return nil, io.EOF
 	}
 	hash := iter.series[iter.pos]
-	obj, err := iter.storage.Get(hash)
+	obj, err := iter.storage.Get(hash, AnyObject)
 	if err == nil {
 		iter.pos++
 	}
@@ -146,7 +151,7 @@ func (iter *ObjectLookupIter) ForEach(cb func(Object) error) error {
 	defer iter.Close()
 
 	for _, hash := range iter.series {
-		obj, err := iter.storage.Get(hash)
+		obj, err := iter.storage.Get(hash, AnyObject)
 		if err != nil {
 			return err
 		}
--- a/core/storage.go
+++ b/core/storage.go
@@ -9,7 +9,16 @@ var ErrStop = errors.New("stop iter")
 type ObjectStorage interface {
 	NewObject() Object
 	Set(Object) (Hash, error)
-	Get(Hash) (Object, error)
+	// Get an object by hash with the given ObjectType.
+	//
+	// Implementors should return (nil, core.ErrObjectNotFound) if an object
+	// doesn't exist with both the given hash and object type.
+	//
+	// Valid ObjectType values are CommitObject, BlobObject, TagObject, TreeObject
+	// and AnyObject.
+	//
+	// If AnyObject is given, the object must be looked up regardless of its type.
+	Get(Hash, ObjectType) (Object, error)
 	Iter(ObjectType) (ObjectIter, error)
 }
 
--- a/storage/filesystem/object.go
+++ b/storage/filesystem/object.go
@@ -38,7 +38,7 @@ func (s *ObjectStorage) Set(core.Object) (core.Hash, error) {
 
 // Get returns the object with the given hash, by searching for it in
 // the packfile.
-func (s *ObjectStorage) Get(h core.Hash) (core.Object, error) {
+func (s *ObjectStorage) Get(h core.Hash, t core.ObjectType) (core.Object, error) {
 	offset, err := s.index.Get(h)
 	if err != nil {
 		return nil, err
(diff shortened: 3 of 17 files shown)


core/object.go
```diff
@@ -49,6 +49,9 @@ const (
 	TagObject      ObjectType = 4
 	OFSDeltaObject ObjectType = 6
 	REFDeltaObject ObjectType = 7
+
+	// Used as a placeholder to request an object of any time.
```
alcortesm: s/time/type/

alcortesm: I'm thinking it will be nice to explain that all these types and their associated numbers are from the official git specification, except for InvalidObject and AnyObject, which are object types of our own invention, *that cannot be found inside a public repository*.

storage/filesystem/object_test.go
```diff
@@ -64,9 +64,11 @@ func (s *FsSuite) TearDownSuite(c *C) {
 
 func (s *FsSuite) TestHashNotFound(c *C) {
 	sto := s.newObjectStorage(c, "binary-relations")
-
-	_, err := sto.Get(core.ZeroHash)
-	c.Assert(err, Equals, core.ErrObjectNotFound)
+	types := []core.ObjectType{core.AnyObject, core.TagObject, core.CommitObject, core.BlobObject, core.TreeObject}
```
alcortesm: We should test cases where the hash is found for one type, but it is not found for other types.

storage/memory/storage.go
```diff
@@ -95,7 +95,7 @@ func (o *ObjectStorage) Set(obj core.Object) (core.Hash, error) {
 }
 
 // Get returns a object with the given hash
-func (o *ObjectStorage) Get(h core.Hash) (core.Object, error) {
+func (o *ObjectStorage) Get(h core.Hash, t core.ObjectType) (core.Object, error) {
```
alcortesm: Improve the comment to explain the objectType

storage/memory/storage_test.go
```diff
@@ -41,7 +41,7 @@ func (s *StorageSuite) TestObjectStorageSetAndGet(c *C) {
 	c.Assert(err, IsNil)
 	c.Assert(h.String(), Equals, "dcf5b16e76cce7425d0beaef62d79a7d10fce1f5")
 
-	e, err := os.Get(h)
+	e, err := os.Get(h, core.AnyObject)
```
alcortesm: we should test that this particular hash is found if you ask for Any or Commit and it is not found when you ask for the rest object types.

core/object.go
smola: Done.

tree.go
```diff
@@ -139,6 +139,7 @@ func (t *Tree) Type() core.ObjectType {
 // Decode transform an core.Object into a Tree struct
 func (t *Tree) Decode(o core.Object) (err error) {
 	if o.Type() != core.TreeObject {
+
```
alcortesm: remove this line

tree_diff.go
```diff
@@ -106,6 +106,8 @@ func newWithEmpty(a, b *Tree) (Changes, error) {
 		if err == io.EOF {
 			break
 		} else if err != nil {
+			fmt.Println("tree a:", a)
```
alcortesm: remove debug leftovers

tree_walker.go
```diff
@@ -66,12 +66,16 @@ func (w *TreeWalker) Next() (name string, entry TreeEntry, obj Object, err error
 			return
 		}
 
-		obj, err = w.r.Object(entry.Hash)
-		if err == ErrObjectNotFound {
-			// FIXME: Avoid doing this here in case the caller actually cares about
-			//        missing objects.
+		if entry.Mode == 0160000 {
```
alcortesm: Make this a constant (const submodule = "160000"), the left zero can be removed.

storage/filesystem/object_test.go
smola: We agreed that behaviour when receiving these types is undefined. Get(Hash, ObjectType) is documented with commit, tag, tree, blob and any as valid object types.

storage/memory/storage.go
smola: Already documented in core/storage.go, which is the interface. I don't think I have to duplicate it in every implementation, right?

storage/memory/storage_test.go
smola: Now tested with the generic suite at storage/test package.

tree.go
smola: Done.

tree_diff.go
smola: Ooops.

tree_walker.go
smola: Done both for submodule and directory modes.

storage/memory/storage.go
alcortesm: Probably not.

storage/filesystem/object.go
```diff
@@ -71,7 +71,14 @@ func (s *ObjectStorage) Get(h core.Hash) (core.Object, error) {
 	p := packfile.NewParser(r)
 
 	obj := s.NewObject()
-	return obj, p.FillObject(obj)
+	err = p.FillObject(obj)
+	if err != nil {
+		return nil, err
+	}
+	if core.AnyObject != t && obj.Type() != t {
```
alcortesm: I think this check should be made before the FillObject operation, it is the less surprise path as it would be the logical order to avoid doing unnecessary extra work in the case of a wrong obj.Type.

storage/filesystem/object_test.go
alcortesm: :+1:

storage/filesystem/object.go
alcortesm: LGTM execpt for that style detail.

smola: We don't know the type until we apply FillObject.
We could, however, call p.ReadObjectTypeAndLength() if we want to retrieve the type before full parsing. That would be more efficient for AnyObject and only slightly less efficient for the others.
What do you think?

smola: Calling p.ReadObjectTypeAndLength before would complicate FillObject logic a little bit though, because it would consume from the input reader, so either FillObject would need to take object type and length as parameters, or ReadObjectTypeAndLength should cache the result for multiple invocations.

core/storage.go
```diff
@@ -9,7 +9,16 @@ var ErrStop = errors.New("stop iter")
 type ObjectStorage interface {
 	NewObject() Object
 	Set(Object) (Hash, error)
-	Get(Hash) (Object, error)
+	// Get an object by hash with the given ObjectType.
+	//
+	// Implementors should return (nil, core.ErrObjectNotFound) if an object
+	// those not exist with both the given hash and object type.
```
alcortesm: Typo: s/those not/doesn't/

storage/filesystem/object.go
alcortesm: Sorry, you were right, obj is needed before checking and reading ahead for the type is convoluted.

Just leave it like it is right now, I think it is the best solution.

alcortesm:
LGTM, looking even better if https://github.com/src-d/go-git/pull/69#discussion_r76196549 is fixed.

In any case, you can merge whenever you want.

PR #70, opened 2016-08-22
object: Add Encode method to all objects.

smola:
Encode method encodes a typed object (commit, tree,
tag, blob) into raw core.Object representation.

Additionally, Decode does not trim commit message
lines. This is needed for Decode/Encode to be
idempotent.

### DIFF
--- a/commit.go
+++ b/commit.go
@@ -97,8 +97,8 @@ func (c *Commit) Decode(o core.Object) (err error) {
 			return err
 		}
 
-		line = bytes.TrimSpace(line)
 		if !message {
+			line = bytes.TrimSpace(line)
 			if len(line) == 0 {
 				message = true
 				continue
@@ -116,7 +116,7 @@ func (c *Commit) Decode(o core.Object) (err error) {
 				c.Committer.Decode(split[1])
 			}
 		} else {
-			c.Message += string(line) + "\n"
+			c.Message += string(line)
 		}
 
 		if err == io.EOF {
@@ -135,6 +135,40 @@ func (c *Commit) History() ([]*Commit, error) {
 
 	ReverseSortCommits(commits)
 	return commits, err
+}
+
+// Encode transforms a Commit into a core.Object.
+func (b *Commit) Encode(o core.Object) error {
+	o.SetType(core.CommitObject)
+	w, err := o.Writer()
+	if err != nil {
+		return err
+	}
+	defer checkClose(w, &err)
+	if _, err = fmt.Fprintf(w, "tree %s\n", b.tree.String()); err != nil {
+		return err
+	}
+	for _, parent := range b.parents {
+		if _, err = fmt.Fprintf(w, "parent %s\n", parent.String()); err != nil {
+			return err
+		}
+	}
+	if _, err = fmt.Fprint(w, "author "); err != nil {
+		return err
+	}
+	if err = b.Author.Encode(w); err != nil {
+		return err
+	}
+	if _, err = fmt.Fprint(w, "\ncommitter "); err != nil {
+		return err
+	}
+	if err = b.Committer.Encode(w); err != nil {
+		return err
+	}
+	if _, err = fmt.Fprintf(w, "\n\n%s", b.Message); err != nil {
+		return err
+	}
+	return err
 }
 
 func (c *Commit) String() string {
--- a/commit_test.go
+++ b/commit_test.go
@@ -2,6 +2,7 @@ package git
 
 import (
 	"io"
+	"time"
 
 	"gopkg.in/src-d/go-git.v4/core"
 
@@ -60,6 +61,42 @@ func (s *SuiteCommit) TestParents(c *C) {
 
 	c.Assert(err, IsNil)
 	c.Assert(output, DeepEquals, expected)
+}
+
+func (s *SuiteCommit) TestCommitEncodeDecodeIdempotent(c *C) {
+	ts, err := time.Parse(time.RFC3339, "2006-01-02T15:04:05-07:00")
+	c.Assert(err, IsNil)
+	commits := []*Commit{
+		&Commit{
+			Author:    Signature{Name: "Foo", Email: "foo@example.local", When: ts},
+			Committer: Signature{Name: "Bar", Email: "bar@example.local", When: ts},
+			Message:   "Message\n\nFoo\nBar\nWith trailing blank lines\n\n",
+			tree:      core.NewHash("f000000000000000000000000000000000000001"),
+			parents:   []core.Hash{core.NewHash("f000000000000000000000000000000000000002")},
+		},
+		&Commit{
+			Author:    Signature{Name: "Foo", Email: "foo@example.local", When: ts},
+			Committer: Signature{Name: "Bar", Email: "bar@example.local", When: ts},
+			Message:   "Message\n\nFoo\nBar\nWith no trailing blank lines",
+			tree:      core.NewHash("0000000000000000000000000000000000000003"),
+			parents: []core.Hash{
+				core.NewHash("f000000000000000000000000000000000000004"),
+				core.NewHash("f000000000000000000000000000000000000005"),
+				core.NewHash("f000000000000000000000000000000000000006"),
+				core.NewHash("f000000000000000000000000000000000000007"),
+			},
+		},
+	}
+	for _, commit := range commits {
+		obj := &core.MemoryObject{}
+		err = commit.Encode(obj)
+		c.Assert(err, IsNil)
+		newCommit := &Commit{}
+		err = newCommit.Decode(obj)
+		c.Assert(err, IsNil)
+		commit.Hash = obj.Hash()
+		c.Assert(newCommit, DeepEquals, commit)
+	}
 }
 
 func (s *SuiteCommit) TestFile(c *C) {
--- a/core/memory.go
+++ b/core/memory.go
@@ -50,6 +50,7 @@ func (o *MemoryObject) Writer() (ObjectWriter, error) {
 
 func (o *MemoryObject) Write(p []byte) (n int, err error) {
 	o.cont = append(o.cont, p...)
+	o.sz = int64(len(o.cont))
 	return len(p), nil
 }
 
--- a/objects.go
+++ b/objects.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"errors"
 	"fmt"
+	"io"
 	"strconv"
 	"time"
 
@@ -39,6 +40,7 @@ type Object interface {
 	ID() core.Hash
 	Type() core.ObjectType
 	Decode(core.Object) error
+	Encode(core.Object) error
 }
 
 // Blob is used to store file data - it is generally a file.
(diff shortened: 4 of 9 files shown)


alcortesm:
I have some doubts about how should we do this.

should we follow the idiomatic way to do this things (see for example encoding/json or encoding/xml)?

I think we should discuss this and decide how are we going to proceed from now on with this kind of things.

commit.go
```diff
@@ -137,6 +137,30 @@ func (c *Commit) History() ([]*Commit, error) {
...
+	for _, parent := range b.parents {
+		w.Write([]byte("\nparent "))
+		w.Write([]byte(parent.String()))
+	}
+	w.Write([]byte("\nauthor "))
+	b.Author.Encode(w)
+	w.Write([]byte("\ncommitter "))
+	b.Committer.Encode(w)
+	w.Write([]byte("\n\n"))
+	w.Write([]byte(b.Message))
```
alcortesm: we should check the error value returned by all those writes.

commit_test.go
```diff
@@ -62,6 +63,47 @@ func (s *SuiteCommit) TestParents(c *C) {
...
+			tree:      core.NewHash("b029517f6300c2da0f4b651b8642506cd6aaf45d"),
+			parents:   []core.Hash{core.NewHash("b029517f6300c2da0f4b651b8642506cd6aaf45d")},
+		},
+		&Commit{
+			Author:    Signature{Name: "Foo", Email: "foo@example.local", When: ts},
+			Committer: Signature{Name: "Bar", Email: "bar@example.local", When: ts},
+			Message:   "Message\n\nFoo\nBar\nBaz\n\n",
+			tree:      core.NewHash("b029517f6300c2da0f4b651b8642506cd6aaf45d"),
+			parents: []core.Hash{
+				core.NewHash("b029517f6300c2da0f4b651b8642506cd6aaf45d"),
```
alcortesm: If we ever check the integrity of the commit being decoded, this test will fail, as a commit can not be its own parent, maybe you can change this value to any other to avoid this.

commit_test.go
```diff
@@ -62,6 +63,47 @@ func (s *SuiteCommit) TestParents(c *C) {
...
+				core.NewHash("e029517f6300c2da0f4b651b8642506cd6aaf45d"),
+			},
+		},
+	}
+	for _, commit := range commits {
+		// Compute hash
+		obj := &core.MemoryObject{}
+		err = commit.Encode(obj)
+		c.Assert(err, IsNil)
+		commit.Hash = obj.Hash()
```
alcortesm: I am not sure I understand the purpose of 91-95, but it looks as if a bug in encode will make this test pass when it shouldn't.

will it be safer to precompute the hash as a fixture instead of relaying in commit.Encode to generate it?

common_test.go
```diff
@@ -156,3 +156,13 @@ func (s *SuiteCommon) TestCountLines(c *C) {
 		c.Assert(o, Equals, t.e, Commentf("subtest %d, input=%q", i, t.i))
 	}
 }
+
+func objectContentString(o core.Object) string {
```
alcortesm: It is difficult to tell on github, but this looks as dead code.

core/memory.go
```diff
@@ -50,6 +50,7 @@ func (o *MemoryObject) Writer() (ObjectWriter, error) {
 
 func (o *MemoryObject) Write(p []byte) (n int, err error) {
 	o.cont = append(o.cont, p...)
+	o.sz = int64(len(o.cont))
```
alcortesm: This maybe breaking the intended use of the setters of this object.

It is difficult to tell as the intended usage is not written anywhere, but if I remember correctly, @mcuadros wants to be able to write the size of the content and the content as to independent things, meaning you can have a content of 10 bytes but a length of 13.

This means this change is a bug and can break how are we using this library from the outside, but I am not sure.

Please @mcuadros, comment on this.

objects.go
```diff
@@ -106,6 +124,15 @@ func (s *Signature) Decode(b []byte) {
 	}
 }
 
+// Encode encodes a Signature into a byte slice
```
alcortesm: Should this comment be: `Encode encodes a Signature into a writer`?

commit.go
```diff
@@ -137,6 +137,30 @@ func (c *Commit) History() ([]*Commit, error) {
...
+// Encode transforms a Commit into a core.Object.
+func (b *Commit) Encode(o core.Object) error {
+	o.SetType(core.CommitObject)
+
+	w, err := o.Writer()
+	if err != nil {
+		return err
+	}
+	w.Write([]byte("tree "))
+	w.Write([]byte(b.tree.String()))
```
alcortesm: should we change this to:
```
fmt.Fprintf(w, "tree %s", b.tree.String())
```

I have not benchmark this against the proposed solution, but if it is not too slow, it is definitely more idiomatic and clear, especially if you do the same on lines 154-...

objects.go
```diff
@@ -133,6 +160,12 @@ func (s *Signature) decodeTimeAndTimeZone(b []byte) {
 	s.When = s.When.In(tl.Location())
 }
 
+func (s *Signature) encodeTimeAndTimeZone(w io.Writer) {
+	w.Write([]byte(strconv.FormatInt(s.When.Unix(), 10)))
+	w.Write([]byte{' '})
+	w.Write([]byte(s.When.Format("-0700")))
```
alcortesm: Maybe we should use fmt.Fprint here.

objects_test.go
```diff
@@ -49,7 +56,7 @@ func (s *ObjectsSuite) TestNewCommit(c *C) {
 	c.Assert(commit.Author.Name, Equals, "Máximo Cuadros")
 	c.Assert(commit.Author.When.Format(time.RFC3339), Equals, "2015-03-31T13:47:14+02:00")
 	c.Assert(commit.Committer.Email, Equals, "mcuadros@gmail.com")
-	c.Assert(commit.Message, Equals, "Merge pull request #1 from dripolles/feature\n\nCreating changelog\n")
+	c.Assert(commit.Message, Equals, "Merge pull request #1 from dripolles/feature\n\nCreating changelog")
```
alcortesm: Is the ending OEL not part of the commit message?

commit_test.go
```diff
@@ -62,6 +63,47 @@ func (s *SuiteCommit) TestParents(c *C) {
...
+		obj = &core.MemoryObject{}
+		err = commit.Encode(obj)
+		c.Assert(err, IsNil)
+		newCommit := &Commit{}
+		err = newCommit.Decode(obj)
+		c.Assert(err, IsNil)
+		c.Assert(newCommit, DeepEquals, commit)
+	}
+}
+
```
alcortesm: I believe a test for encode is missing. Is it because we have a test for decode and a test for encode-decode idempotence?

objects_test.go
```diff
@@ -108,6 +115,24 @@ func (s *ObjectsSuite) TestBlobHash(c *C) {
 	c.Assert(string(data), Equals, "FOO")
 }
 
+func (s *ObjectsSuite) TestBlobDecodeEncodeIdempotent(c *C) {
+	var objects []*core.MemoryObject
+	obj := &core.MemoryObject{}
+	obj.Write([]byte("foo"))
+	obj.SetType(core.BlobObject)
+	obj.Hash()
```
alcortesm: What is the point of calling this method?, something is not right in the API.

objects_test.go
```diff
@@ -108,6 +115,24 @@ func (s *ObjectsSuite) TestBlobHash(c *C) {
...
+	objects = append(objects, obj)
+	for _, object := range objects {
```
alcortesm: there is only on object in `objects`, what is the point of having a slice of only one element?

tag.go
```diff
@@ -105,6 +105,27 @@ func (t *Tag) Decode(o core.Object) (err error) {
...
+	w.Write([]byte(t.Target.String()))
+	w.Write([]byte("\ntype "))
+	w.Write(t.TargetType.Bytes())
+	w.Write([]byte("\ntag "))
+	w.Write([]byte(t.Name))
+	w.Write([]byte("\ntagger "))
+	t.Tagger.Encode(w)
+	w.Write([]byte("\n\n"))
+	w.Write([]byte(t.Message))
+	o.Hash()
```
alcortesm: why are we calling this method here?

tree.go
```diff
@@ -193,6 +194,28 @@ func (t *Tree) Decode(o core.Object) (err error) {
...
 
+// Encode transforms a Tree into a core.Object.
+func (t *Tree) Encode(o core.Object) error {
+	o.SetType(core.TreeObject)
+	w, err := o.Writer()
+	if err != nil {
+		return err
+	}
+	for _, entry := range t.Entries {
+		//if idx != 0 {
```
alcortesm: remove dead code

commit.go
```diff
@@ -137,6 +137,30 @@ func (c *Commit) History() ([]*Commit, error) {
...
+		w.Write([]byte("\nparent "))
+		w.Write([]byte(parent.String()))
+	}
+	w.Write([]byte("\nauthor "))
+	b.Author.Encode(w)
+	w.Write([]byte("\ncommitter "))
+	b.Committer.Encode(w)
+	w.Write([]byte("\n\n"))
+	w.Write([]byte(b.Message))
+	o.Hash() // Ensure that Hash is computed
```
alcortesm: why?

commit.go
```diff
@@ -137,6 +137,30 @@ func (c *Commit) History() ([]*Commit, error) {
...
+	}
+	w.Write([]byte("tree "))
+	w.Write([]byte(b.tree.String()))
+	for _, parent := range b.parents {
+		w.Write([]byte("\nparent "))
+		w.Write([]byte(parent.String()))
+	}
+	w.Write([]byte("\nauthor "))
+	b.Author.Encode(w)
+	w.Write([]byte("\ncommitter "))
```
alcortesm: will it be better to use templates for this task?, not only here, but in the tag, blob...

commit_test.go
smola: In order to know the exact object that we expect from Decode(Encode(Commit)), we need to know the hash of the encoded commit. If we use a git repo fixture as with other tests, this would be a non-issue, since we'd test Encode(Decode(Object)) and the Object hash would be one actually generated by git.

In order to do it in reverse, we need to encode the object to obtain the hash.

I'll change this to be a bit more concise though.

objects_test.go
smola: We have to encode a Blob into a core.Object. Blob implementation contains an inner core.Object, but we're not guaranteed that this inner core.Object is of the same type as the one we want to encode to.

If we want to change it, we'd probably need to simplify the whole thing and make core.MemoryObject the only representation of object and/or make .Encode() create its own core.Object, instead of receiving one.

tag.go
smola: It's the only way to ensure that the object has a computed hash.
If we want to change it, we could introduce a Finalize method, or something like that, that can be called once the object is written to complete any further fields (e.g. hash, size).

It seems to me that changing that is way beyond the scope of this PR though.

commit.go
smola: See above.

commit.go
smola: I would refrain from introducing anything here that leads to a loss of performance, since some tasks will require massive usage of Encode.

objects_test.go
smola: That's right. Before this PR, it was normalized during Decode. This PR removes the normalization step so we preserve the input as it is.
Otherwise, Encode/Decode would not be idempotent.

commit_test.go
smola: Yes. I used to this this with a git repository fixture, doing Decode first and then Encode. Changing it to mocks + Encode + Decode requires us to calculate the hash of the encoded object to know what should we expect.

I changed it to be a bit more concise though.

tag.go
smola: You're right. I'm removing it and adapting the tests.

core/memory.go
mcuadros: Exactly this collides with the intent of SetSize, I am changing it after merge

PR #79, opened 2016-09-23
Avoid panic if repository is not specified

jrevillas:


mcuadros:
Thanks for the PRs, we are lunching soon the [v4](https://github.com/src-d/go-git/tree/v4), and those example are being updated too.

Anyways, thanks!

PR #85, opened 2016-10-17
Fix StorageSuite.TestConfigStorageRemotes

alcortesm:
This tests checks the contents of a map, assuming the maps stores the elements in a certain order. It only failed like once every 6 or 7 runs.

I fixed it by sorting the returned elements before comparing them.


smola:
Could you rebase this from origin/master?

alcortesm:
I have tried to rebase it several times, with weird rebase conflicts I cannot solve. I will try again later.

PR #86, opened 2016-10-18
Fix StorageSuite.TestConfigStorageRemotes

alcortesm:
Don't assume the remotes are retrieved in chronological order.


alcortesm:
fuuuuuck!

PR #88, opened 2016-10-18
utils: fs.TempFile

mcuadros:
- New fs.TempFile method in the fs.Filesystem interface, this abstraction is required to filesystem where the creation of a temporal file should be explicit 
- utils.fs: new TestSuite for test any fs.Filesystem implementation

### DIFF
--- a/utils/fs/fs_test.go
+++ b/utils/fs/fs_test.go
@@ -1,0 +1,121 @@
+package fs
+
+import (
+	"io"
+	"io/ioutil"
+	"testing"
+
+	. "gopkg.in/check.v1"
+)
+
+func Test(t *testing.T) { TestingT(t) }
+
+type FilesystemSuite struct {
+	fs Filesystem
+}
+
+func (s *FilesystemSuite) TestCreate(c *C) {
+	f, err := s.fs.Create("foo")
+	c.Assert(err, IsNil)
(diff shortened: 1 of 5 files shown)


utils/fs/os.go
```diff
@@ -92,24 +99,56 @@ func (fs *OS) Open(filename string) (File, error) {
...
+// Stat returns the FileInfo structure describing file.
 func (fs *OS) Stat(filename string) (FileInfo, error) {
-	fullpath := fs.Join(fs.RootDir, filename)
+	fullpath := fs.Join(fs.base, filename)
 	return os.Stat(fullpath)
 }
 
+func (fs *OS) TempFile(dir, prefix string) (File, error) {
+	fullpath := fs.Join(fs.base, dir)
+	if err := fs.createDir(fullpath + "/"); err != nil {
```
alcortesm: Is "/" portable?

should we use os.PathSeparator instead?

utils/fs/fs_test.go
```diff
@@ -0,0 +1,149 @@
...
+	. "gopkg.in/check.v1"
+)
+
+func Test(t *testing.T) { TestingT(t) }
+
+type FilesystemSuite struct {
+	fs Filesystem
+}
+
+func (s *FilesystemSuite) TestCreate(c *C) {
```
mcuadros: Done

PR #89, opened 2016-10-19
Fix pktline api

alcortesm:
The old pktline API only had one method: New, that receives the
payloads and return the new PktLine, that was an io.Reader.  This means
you have to prepare the contents beforehand, in a [][]byte or a []string and then
call the ctor to build the pktlines:

```go
payloads := []string{
    "hello\n",
    "world!\n",
    "",
}

p, _ := pktline.NewFromStrings(payloads...)

io.Copy(os.Stdout, p)
```

Now, the construction of the pktlines and the methods to add payloads and flush-pkts are
separated and the reader is an attribute:

```go
p := pktlines.New()

_ = p.AddString(
    "hello\n",
    "world!\n",
)
p.AddFlush()

io.Copy(os.Stdout, p.R)
```

I have also change the name of the package to its plural version: pktlines, as the package purpose is to handle several pktlines together, not just one.


mcuadros:
IMHO the structure can be Packfiles, but the name of the package should be the name of the format.

alcortesm:
Agree @mcuadros, will do that.

alcortesm:
I will also change the public attribute to a Reader method.

formats/packp/pktlines/pktlines.go
```diff
@@ -0,0 +1,136 @@
...
+	ErrPayloadTooLong = errors.New("payload is too long")
+	// ErrEmptyPayload is returned by the Add methods when an empty
+	// payload is provided.
+	ErrEmptyPayload = errors.New("cannot add empty payloads")
+)
+
+// New returns an empty PktLines (with no payloads) ready to be used.
+func New() *PktLines {
+	return &PktLines{
+		R: io.MultiReader([]io.Reader{}...),
```
dpordomingo: maybe better if it would be something like p.Reader()

alcortesm: You are right, will do!

alcortesm:
Package name changes to `pktline` (singular).

Public reader attribute removed, now PktLines are Readers themselves.

PR #90, opened 2016-10-19
OSFile: implement io.ReaderAt.

smola:


utils/fs/os.go
```diff
@@ -170,3 +170,7 @@ func (f *OSFile) Close() error {
 
 	return f.file.Close()
 }
+
+func (f *OSFile) ReadAt(p []byte, off int64) (n int, err error) {
```
dpordomingo: return with named values? why not just "(int, error)" as done in the other os.methods?

smola: Right, forgot after copypasta.

PR #91, opened 2016-10-20
formats: Index read support

mcuadros:
This is a basic implementation of the [Index format](https://raw.githubusercontent.com/git/git/35f6318d44379452d8d33e880d8df0267b4a0cd0/Documentation/technical/index-format.txt)

The index format contains the information about which objects are currently checked out in the worktree, having information about the working files. Changes in worktree are detected using this Index. The Index is also used during merge.

Three types of version are available 2, 3 and 4, this implementation only supports 2, and I am planning to add support for version 3 and 4, in this same PR

This format also have several extensions, this implementation only have the cache Tree extension. 

So the pending stuff is:
- [x] Support version 3 and 4
- [x] Proper decoding of nested types like `Flags` and `Mode`
- [x] Implementation of `Resolve undo` extension
- [ ] Implementation of `Untracked cache` extension
- [ ] Support split index

I must say that this is one of the most complex format in git 

_UPDATE_: I am not implementing `Split Index` and `Untracked cache` extensions because are not supported by libgit or jgit, also because are optional and arcane extension. The reader is ready for review


formats/index/decoder.go
```diff
@@ -0,0 +1,389 @@
...
+
+		idx.Cache = t
+	}
+
+	return nil
+}
+
+func validateHeader(r io.Reader) (version uint32, err error) {
+	var s = make([]byte, 4)
+	if _, err := r.Read(s); err != nil {
```
alcortesm: This must be a ReadFull.

formats/index/decoder.go
```diff
@@ -0,0 +1,389 @@
...
+		&e.Size,
+		&e.Hash,
+		&e.Flags,
+	}
+
+	if err := readBinary(d.r, flow...); err != nil {
+		return nil, err
+	}
+
+	read := 62
```
alcortesm: this is a magic number.

formats/index/decoder.go
```diff
@@ -0,0 +1,389 @@
...
+	e.CreatedAt = time.Unix(int64(msec), int64(mnsec))
+	e.ModifiedAt = time.Unix(int64(sec), int64(nsec))
+	e.Stage = Stage(e.Flags>>12) & 0x3
+
+	if e.Flags&EntryExtended != 0 {
+		read += 2
```
alcortesm: Remove this all other magic numbers by making readBinary return the number of bytes read.

formats/index/decoder.go
```diff
@@ -0,0 +1,389 @@
...
+			return nil, err
+		}
+
+		e.IntentToAdd = extended&intentToAddMask != 0
+		e.SkipWorktree = extended&skipWorkTreeMask != 0
+	}
+
+	if err := d.readEntryName(idx, e); err != nil {
+		return nil, err
+
```
alcortesm: remove this empty line

formats/index/decoder.go
```diff
@@ -0,0 +1,389 @@
...
+func (d *Decoder) readEntryName(idx *Index, e *Entry) error {
+	var name string
+	var err error
+
+	switch idx.Version {
+	case 2, 3:
+		name, err = d.doReadEntryName(e)
+	case 4:
+		name, err = d.doReadEntryNameV4()
+	}
```
alcortesm: Version is a public field, there should be a default in this switch.

formats/index/decoder.go
```diff
@@ -0,0 +1,389 @@
...
+			break
+		}
+	}
+
+	if err == io.EOF {
+		return nil
+	}
+
+	return nil
+}
```
alcortesm: The are some checks missing here, otherwise the last 5 lines can be replaced by:

```
return nil
```

And whole method does not need to return anything.

formats/index/decoder.go
```diff
@@ -0,0 +1,389 @@
...
+
+func (d *Decoder) readExtension(idx *Index) error {
+	var s = make([]byte, 4)
+	if _, err := d.r.Read(s); err != nil {
```
alcortesm: ReadFull

formats/index/decoder.go
```diff
@@ -0,0 +1,389 @@
...
+		return err
+	}
+
+	var len uint32
+	if err := binary.Read(d.r, binary.BigEndian, &len); err != nil {
+		return err
+	}
+
+	switch {
+	case bytes.Equal(s, treeExtSignature):
```
alcortesm: why this switch has only one case? There should be a default with some unsupported extension comment or warning.

formats/index/decoder.go
```diff
@@ -0,0 +1,389 @@
...
+	}
+
+	if version < IndexVersionSupported.Min || version > IndexVersionSupported.Max {
+		return 0, ErrUnsupportedVersion
+	}
+
+	return
+}
+
+type TreeExtensionDecoder struct {
```
alcortesm: move this to its own file.

formats/index/decoder.go
```diff
@@ -0,0 +1,389 @@
...
+	}
+
+	return nil
+}
+
+func readUntil(r io.Reader, delim byte) ([]byte, error) {
+	var buf [1]byte
+	value := make([]byte, 0, 16)
+	for {
+		if _, err := r.Read(buf[:]); err != nil {
```
alcortesm: ReadFull or check n!=0 then continue with a sleep or something.

formats/index/decoder.go
```diff
@@ -0,0 +1,389 @@
...
+			return err
+		}
+	}
+
+	return nil
+}
+
+func readUntil(r io.Reader, delim byte) ([]byte, error) {
+	var buf [1]byte
+	value := make([]byte, 0, 16)
```
alcortesm: Why did you choose 16 as a good initial size?

formats/index/decoder.go
```diff
@@ -0,0 +1,389 @@
...
+	}
+
+	i, err := strconv.Atoi(string(count))
+	if err != nil {
+		return nil, err
+	}
+
+	e.Entries = i
+
+	trees, err := readUntil(d.r, 10)
```
alcortesm: NO WAY, use `\n`!

formats/index/decoder.go
```diff
@@ -0,0 +1,389 @@
...
+		return nil, err
+	}
+
+	e.Trees = i
+
+	if err := binary.Read(d.r, binary.BigEndian, &e.Hash); err != nil {
+		return nil, err
+	}
+
+	return e, nil
```
alcortesm: Can you replace all this function with `fmt.Fscanf(d.r, "%s\x00%d %s\n%s" &e.Path, &e.Entries, &e.Tree, &e.Hash)` ? or some other one-liner?

formats/index/decoder.go
```diff
@@ -0,0 +1,389 @@
...
+
+		if buf[0] == delim {
+			return value, nil
+		}
+
+		value = append(value, buf[0])
+	}
+}
+
+//     dheader[pos] = ofs & 127;
```
alcortesm: The last 30 lines look familiar, didn't we have code this before?

This comment must be explained.

formats/index/decoder.go
```diff
@@ -0,0 +1,389 @@
...
+		}
+
+		t.Entries = append(t.Entries, *e)
+	}
+}
+
+func (d *TreeExtensionDecoder) readEntry() (*TreeEntry, error) {
+	e := &TreeEntry{}
+
+	path, err := readUntil(d.r, 0)
```
alcortesm: For some reason using an integer here sounds weird, I think it will be much better if you use `\x00` here as the delimiter.

formats/index/decoder.go
mcuadros: right, fixing it

formats/index/decoder.go
mcuadros: right, fixing it

formats/index/decoder.go
mcuadros: right, fixing it

formats/index/decoder.go
mcuadros: nice catch

formats/index/decoder.go
mcuadros: right, fixing it

formats/index/decoder.go
mcuadros: now we have more than one

formats/index/decoder.go
mcuadros: right, fixing it

formats/index/decoder.go
mcuadros: I am making it private

formats/index/decoder.go
mcuadros: solved

formats/index/decoder.go
mcuadros: hahaha

formats/index/decoder.go
mcuadros: The logic now is a bit more complex, so... I prefer not

formats/index/decoder.go
mcuadros: yep don't worry I am making another PR creating a package for binary reads

PR #93, opened 2016-10-24
utils/fs: move 'os' and 'test' to separate packages.

smola:
* create utils/fs/test package to expose generic test suite to 3rd party fs implementations.
* move 'os' to its own package to avoid cyclic dependency (test -> fs -> test, becomes test -> fs, os -> test, os -> fs).
* remove TestCreateAndWrite: some of our implementations cannot read a file that was just created, written and not closed yet.


smola:
@alcortesm We moved the generic fs suite to a separate package `test` so that is can be used by 3rd party fs implementations. Now the `os` implementation depends on `test` and `test` depends on `fs`, so it is a cyclic dependency and not allowed. So we need to have `os` in its own package if we want to expose the `test` package. I'm updating the PR description with this.

PR #97, opened 2016-10-25
formats/config: Added encoder/decoder for git config files.

smola:


formats/config/decoder.go
```diff
@@ -0,0 +1,233 @@
...
+		return err
+	}
+
+	fset := token.NewFileSet()
+	file := fset.AddFile("", fset.Base(), len(src))
+	c := warnings.NewCollector(func(error) bool {
+		return true
+	})
+
+	err = decode(c, config, fset, file, src)
```
ferhatelmas: simply `return decode(c, config, fset, file, src)`

formats/config/common.go
```diff
@@ -0,0 +1,147 @@
...
+type Comment string
+
+// IsKey returns true if the given key matches
+// this options' key in a case-insensitive comparison.
+func (o *Option) IsKey(key string) bool {
+	return strings.ToLower(o.Key) == strings.ToLower(key)
+}
+
+func (c *Config) Section(name string) *Section {
+	for i := len(c.Sections) - 1; i >= 0; i-- {
```
ferhatelmas: Why not you prefer `range` ? Is there any special meaning iterating in reverse order ?

alcortesm: Does it has to do with disambiguating sections with the same name to the one that was appended last?, is name supposed to be an unique id?

formats/config/common.go
```diff
@@ -0,0 +1,147 @@
...
+	for i := len(opts) - 1; i >= 0; i-- {
+		o := opts[i]
+		if o.IsKey(key) {
+			return o.Value
+		}
+	}
+	return ""
+}
+
+func (s *Config) AddOption(section string, subsection string, key string, value string) *Config {
```
alcortesm: Does this method returns its receiver to be chainable?

formats/config/common.go
```diff
@@ -0,0 +1,147 @@
...
+		if o.IsKey(key) {
+			o.Value = value
+			return s
+		}
+	}
+
+	return s.AddOption(key, value)
+}
+
+func (s *Subsection) SetOption(key string, value string) *Subsection {
```
alcortesm: there is a lot of duplicated code here.

Would it be possible to avoid it by embedding subsection into section or something along those lines...

formats/config/decoder.go
```diff
@@ -0,0 +1,233 @@
...
+				if err := c.Collect(errfn("expected section header")); err != nil {
+					return err
+				}
+			}
+			if err := c.Collect(errfn("expected section header or variable declaration")); err != nil {
+				return err
+			}
+		}
+	}
+	panic("never reached")
```
alcortesm: This function is 150 lines long.

Can we slipt it without impairing its legibility?

formats/config/decoder.go
smola: Right. Thanks.

formats/config/common.go
smola: When we have multiple options with the same key, the standard git config behaviour for Get is that the last value wins. So we iterate in reverse order and return the first value.

For Sections/Subsections we do the same. Although this will probably have to be changed in the future if we want to handle merging of repeated sections. At the moment, we do not require such functionality, but we'll have to handle this when we want to work with git config files in the wild (at the moment, we just want to get/set/delete remotes for our storage backend).

formats/config/common.go
smola: Right.

formats/config/decoder.go
smola: Removed. It was a copy of an internal gcfg function that we needed to modify. This is now in our own gcfg fork: https://github.com/src-d/gcfg

formats/config/common.go
alcortesm: :+1:

formats/config/common.go
alcortesm: :+1:

formats/config/common.go
alcortesm: :+1:

formats/config/fixtures_test.go
```diff
@@ -0,0 +1,92 @@
...
+	{
+		Raw: `
+			[sect1]
+			opt1 = value1
+			[sect1 "subsect1"]
+			opt2 = value2
+		`,
+		Text: `[sect1]
+	opt1 = value1
+[sect1 "subsect1"]
```
ajnavarro: That's the expected format?

formats/config/fixtures_test.go
```diff
@@ -0,0 +1,92 @@
...
+			opt1 = value1b
+			[sect1 "subsect1"]
+			opt2 = value2b
+			[sect1 "subsect2"]
+			opt2 = value2
+		`,
+		Text: `[sect1]
+	opt1 = value1
+	opt1 = value1b
+[sect1 "subsect1"]
```
ajnavarro: That's the expected format?

formats/config/common.go
```diff
@@ -0,0 +1,92 @@
...
+	}
+
+	return s
+}
+
+// SetOption is a convenience method to set an option to a given
+// section and subsection.
+// If subsection is empty, then it's taken as no subsection.
+func (s *Config) SetOption(section string, subsection string, key string, value string) *Config {
+	if subsection == "" {
```
alcortesm: This is similar to what happen with flush-pkt in the past, that we should pass a constant instead of making the user aware of the specific value.  How about a `NoSection` constant with the value of ""?

formats/config/section.go
```diff
@@ -0,0 +1,76 @@
...
+	Options     Options
+	Subsections Subsections
+}
+
+type Subsection struct {
+	Name    string
+	Options Options
+}
+
+type Sections []*Section
```
alcortesm: `Options`, `Sections` and `Subsections` are currently slices. Willl it make sense to use maps instead, given that we are somewhat assuming names will be unique ids?

formats/config/common.go
```diff
@@ -0,0 +1,92 @@
...
+
+type Includes []*Include
+
+// A reference to a included configuration.
+type Include struct {
+	Path   string
+	Config *Config
+}
+
+type Comment string
```
alcortesm: This is weird, did you forgot to add some methods to this type? otherwise, why can we use the string type instead of Comment?

formats/config/fixtures_test.go
smola: Yeah, tabs at the left are preserved when using `, and I need no tabs at the left.

formats/config/fixtures_test.go
smola: Same here.

formats/config/common.go
smola: Yep. It's still missing comment handling code such as returning comment without comment symbols (;. #).

formats/config/section.go
smola: There might be multiple sections with the same name. It's up to us to offer convenience methods in the future that hide this for common use case, but the structure needs to preserve this.

formats/config/common.go
smola: Done.

formats/config/section.go
alcortesm: :+1:

formats/config/common_test.go
```diff
@@ -0,0 +1,103 @@
...
+
+	"gopkg.in/src-d/go-git.v4/formats/config"
+
+	. "gopkg.in/check.v1"
+)
+
+func Test(t *testing.T) { TestingT(t) }
+
+type CommonSuite struct {
+}
```
smola: Done.

formats/config/common_test.go
```diff
@@ -0,0 +1,103 @@
+package config_test
```
smola: Done.

formats/config/common.go
```diff
@@ -0,0 +1,147 @@
+// Package config implements decoding, encoding and
```
smola: Done.

formats/config/common.go
```diff
@@ -0,0 +1,147 @@
...
+	return s
+}
+
+func (s *Subsection) AddOption(key string, value string) *Subsection {
+	s.Options = append(s.Options, &Option{key, value})
+	return s
+}
+
+func (s *Section) SetOption(key string, value string) *Section {
+	for i := len(s.Options) - 1; i >= 0; i-- {
```
ferhatelmas: To me, subsection is a section that has no children.

That's why I think we can remove subsection and put children (subsections) as a reference to itself (`[]*Section`). As a result it will be defined fully recursive and there will be no repetition.

PR #104, opened 2016-10-31
utils/fs: add OpenFile method to filesystem interface.

smola:
* added OpenFile to fs.Filesystem interface.
* added OpenFile implementation to 'os' filesystem.
* bring back BaseFile.

### DIFF
--- a/utils/fs/fs.go
+++ b/utils/fs/fs.go
@@ -14,10 +14,9 @@ var (
 )
 
 type Filesystem interface {
-	//Create opens a file in write-only mode.
 	Create(filename string) (File, error)
-	//Open opens a file in read-only mode.
 	Open(filename string) (File, error)
+	OpenFile(filename string, flag int, perm os.FileMode) (File, error)
 	Stat(filename string) (FileInfo, error)
 	ReadDir(path string) ([]FileInfo, error)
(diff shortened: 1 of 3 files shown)


utils/fs/os/os.go
```diff
@@ -21,7 +21,8 @@ func NewOS(baseDir string) *OS {
 	}
 }
 
-// Create creates a new GlusterFSFile
+// Create creates a file and opens it with standard permissions
+// and modes O_RDWR, O_CREATE and O_TRUNC.
 func (fs *OS) Create(filename string) (File, error) {
```
mcuadros: to avoid duplicate code you can call `fs.OpenFile(filename, os.O_RDWR|os.O_CREATE|os.O_TRUNC, 0666)`, and the same for Open

utils/fs/fs.go
```diff
@@ -38,3 +37,16 @@ type File interface {
 }
 
 type FileInfo os.FileInfo
+
+type BaseFile struct {
```
mcuadros: I though that we are adding this to the bare package not to the OS one

smola: Removed wildcard import.

PR #106, opened 2016-10-31
ulreq: adds encoder and decoder for upload-request messages

alcortesm:
This PRs adds a new package: `upreq` to handle upload-requests messages from a git-upload-pack service.  You will find a description of the message [here](https://github.com/git/git/blob/master/Documentation/technical/pack-protocol.txt#L196).

The package introduces a new type for representing upload-requests messages in memory, an its encoder and a decoder.

The SSH client has been updated to use this new type and its encoder and the code for stepping through the pack protocol has been cleaned.

### DIFF
--- a/formats/packp/ulreq/decoder_test.go
+++ b/formats/packp/ulreq/decoder_test.go
@@ -1,0 +1,542 @@
+package ulreq_test
+
+import (
+	"bytes"
+	"io"
+	"sort"
+	"time"
+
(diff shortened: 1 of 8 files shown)


formats/packp/ulreq/decoder_test.go
```diff
@@ -0,0 +1,542 @@
+package ulreq_test
```
alcortesm: Of course.

PR #108, opened 2016-11-02
utils: fs, new memory filesystem	

mcuadros:
This is a new `fs.Filesytem` based on memory.  take notice of is not thread-safe. 
Also I renamed  os.NewOS` to `os.New`

### DIFF
--- a/utils/fs/memory/memory.go
+++ b/utils/fs/memory/memory.go
@@ -1,0 +1,333 @@
+package memory
+
+import (
+	"errors"
+	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+	"strings"
+	"time"
+
+	"gopkg.in/src-d/go-git.v4/utils/fs"
+)
+
+const separator = '/'
+
+// Memory a very convenient filesystem based on memory files
+type Memory struct {
+	base string
+	s    *storage
+}
+
+//New returns a new Memory filesystem
+func New() *Memory {
+	return &Memory{
+		base: "/",
+		s:    &storage{make(map[string]*file, 0)},
+	}
+}
+
+func (fs *Memory) Create(filename string) (fs.File, error) {
+	return fs.OpenFile(filename, os.O_RDWR|os.O_CREATE|os.O_TRUNC, 0666)
+}
+
+func (fs *Memory) Open(filename string) (fs.File, error) {
+	return fs.OpenFile(filename, os.O_RDONLY, 0)
+}
+
+func (fs *Memory) OpenFile(filename string, flag int, perm os.FileMode) (fs.File, error) {
+	fullpath := fs.Join(fs.base, filename)
+	f, ok := fs.s.files[fullpath]
+	if !ok && flag&os.O_CREATE == 0 {
+		return nil, os.ErrNotExist
+	}
+
+	if f == nil {
+		fs.s.files[fullpath] = newFile(fs.base, fullpath, flag)
+		return fs.s.files[fullpath], nil
+	}
+
+	n := newFile(fs.base, fullpath, flag)
+	n.c = f.c
+
+	if flag&os.O_APPEND != 0 {
+		n.p = n.c.size
+	}
+
+	if flag&os.O_TRUNC != 0 {
+		n.c.Truncate()
+	}
+
+	return n, nil
+}
+
+func (fs *Memory) Stat(filename string) (fs.FileInfo, error) {
+	fullpath := fs.Join(fs.base, filename)
+
+	if _, ok := fs.s.files[filename]; ok {
+		return newFileInfo(fs.base, fullpath, fs.s.files[filename].c.size), nil
+	}
+
+	info, err := fs.ReadDir(filename)
+	if err == nil && len(info) != 0 {
+		return newFileInfo(fs.base, fullpath, int64(len(info))), nil
+	}
+
+	return nil, os.ErrNotExist
+}
+
+func (fs *Memory) ReadDir(base string) (entries []fs.FileInfo, err error) {
+	base = fs.Join(fs.base, base)
+
+	dirs := make(map[string]bool, 0)
+	for fullpath, f := range fs.s.files {
+		if !strings.HasPrefix(fullpath, base) {
+			continue
+		}
+
+		fullpath, _ = filepath.Rel(base, fullpath)
+		parts := strings.Split(fullpath, string(separator))
+
+		if len(parts) != 1 {
+			dirs[parts[0]] = true
+			continue
+		}
+
+		entries = append(entries, newFileInfo(fs.base, fullpath, f.c.size))
+	}
+
+	for path := range dirs {
+		entries = append(entries, &fileInfo{
+			name:  path,
+			isDir: true,
+		})
+	}
+
+	return
+}
+
+func (fs *Memory) TempFile(dir, prefix string) (fs.File, error) {
+	filename := fmt.Sprintf("%s_%d", prefix, time.Now().UnixNano())
+	fullpath := fs.Join(fs.base, dir, filename)
+	return fs.Create(fullpath)
+}
+
+func (fs *Memory) Rename(from, to string) error {
+	from = fs.Join(fs.base, from)
+	to = fs.Join(fs.base, to)
+
+	if _, ok := fs.s.files[from]; !ok {
+		return os.ErrNotExist
+	}
+
+	fs.s.files[to] = fs.s.files[from]
+	fs.s.files[to].BaseFilename = to
+	delete(fs.s.files, from)
+
+	return nil
+}
+
+func (fs *Memory) Remove(filename string) error {
+	if _, err := fs.Stat(filename); err != nil {
+		return err
+	}
+
+	fullpath := fs.Join(fs.base, filename)
+	delete(fs.s.files, fullpath)
+	return nil
+}
+
+func (fs *Memory) clean(path string) string {
+	if len(path) <= 1 {
+		if path != string(separator) {
+			return path
+		}
+
+		return ""
+	}
+
+	if path[0] == separator {
+		path = path[1:]
+	}
+
+	l := len(path)
+	if path[l-1] == separator {
+		path = path[:l-1]
+	}
+
+	return path
+}
+
+func (fs *Memory) Join(elem ...string) string {
+	return filepath.Join(elem...)
+}
+
+func (fs *Memory) Dir(path string) fs.Filesystem {
+	return &Memory{
+		base: fs.Join(fs.base, path),
+		s:    fs.s,
+	}
+}
+
+func (fs *Memory) Base() string {
+	return fs.base
+}
+
+type file struct {
+	fs.BaseFile
+
+	c     *content
+	p     int64
+	flags int
+}
+
+func newFile(base, fullpath string, flags int) *file {
+	filename, _ := filepath.Rel(base, fullpath)
+
+	return &file{
+		BaseFile: fs.BaseFile{BaseFilename: filename},
+		c:        &content{},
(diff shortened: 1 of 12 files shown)


utils/fs/memory/memory.go
```diff
@@ -0,0 +1,333 @@
...
+		base: fs.Join(fs.base, path),
+		s:    fs.s,
+	}
+}
+
+func (fs *Memory) Base() string {
+	return fs.base
+}
+
+type file struct {
```
smola: Maybe it's worth to move file, fileInfo and content to a different file.

utils/fs/memory/memory.go
```diff
@@ -0,0 +1,333 @@
...
+	if off >= c.size {
+		return 0, io.EOF
+	}
+
+	l := int64(len(b))
+	if off+l > c.size {
+		l = c.size - off
+	}
+
+	n := copy(b, c.bytes[off:l])
```
alcortesm: This cannot be right, l can be less than off here, but that is clearly impossible. I think you have to change `c.bytes[off:l]` to `c.bytes[off:off+l]`. I think there has been mistake testing this code, what was its test coverage?

utils/fs/memory/memory.go
```diff
@@ -0,0 +1,333 @@
...
+	return n, nil
+}
+
+func (c *content) Truncate() {
+	c.bytes = []byte{}
```
alcortesm: use make.

utils/fs/memory/memory.go
```diff
@@ -0,0 +1,333 @@
...
+	return f.p, nil
+}
+
+func (f *file) Write(p []byte) (int, error) {
+	if f.IsClosed() {
+		return 0, fs.ErrClosed
+	}
+
+	if f.flags&os.O_RDWR != 0 && f.flags&os.O_WRONLY != 0 {
+		return 0, errors.New("read not supported")
```
alcortesm: Should we return errors like the ones returned by golang os package?

utils/fs/memory/memory.go
```diff
@@ -0,0 +1,333 @@
...
+	return nil
+}
+
+type storage struct {
+	files map[string]*file
+}
+
+type content struct {
+	bytes []byte
+	size  int64
```
mcuadros: because the len of bytes is bigger than the current data.

utils/fs/memory/memory.go
```diff
@@ -0,0 +1,333 @@
...
+}
+
+func (c *content) WriteAt(p []byte, off int64) (int, error) {
+	l := len(p)
+	if int(off)+l > len(c.bytes) {
```
mcuadros: done

utils/fs/memory/memory.go
```diff
@@ -0,0 +1,333 @@
...
+		buf := make([]byte, 2*len(c.bytes)+l)
```
mcuadros: done

utils/fs/memory/memory.go
mcuadros: fixes and coverage improved

utils/fs/memory/memory.go
mcuadros: done

utils/fs/memory/memory.go
```diff
@@ -0,0 +1,333 @@
...
+func (fi *fileInfo) Name() string {
+	return fi.name
+}
+
+func (fi *fileInfo) Size() int64 {
+	return fi.size
+}
+
+func (fi *fileInfo) Mode() os.FileMode {
+	return os.FileMode(0)
```
mcuadros: I though that storing and retuning modes, may run on confusions because doesn't have any effect

utils/fs/memory/memory.go
```diff
@@ -0,0 +1,333 @@
...
+
+	return n, err
+}
+
+func (f *file) Close() error {
+	f.Closed = true
+	return nil
+}
+
+func (f *file) Open() error {
```
mcuadros: why? on os you can open a file several times

utils/fs/memory/memory.go
```diff
@@ -0,0 +1,333 @@
...
+	}
+
+	n, err := f.c.WriteAt(p, f.p)
+	f.p += int64(n)
+
+	return n, err
+}
+
+func (f *file) Close() error {
+	f.Closed = true
```
mcuadros: fixed

utils/fs/memory/memory.go
```diff
@@ -0,0 +1,333 @@
...
+	n, err := f.c.ReadAt(b, f.p)
+	f.p += int64(n)
+
+	return n, err
+}
+
+func (f *file) Seek(offset int64, whence int) (int64, error) {
+	switch whence {
+	case io.SeekCurrent:
+		if offset == 0 {
```
mcuadros: true

utils/fs/memory/memory.go
```diff
@@ -0,0 +1,333 @@
...
+type file struct {
+	fs.BaseFile
+
+	c     *content
+	p     int64
+	flags int
+}
+
+func newFile(base, fullpath string, flags int) *file {
+	filename, _ := filepath.Rel(base, fullpath)
```
mcuadros: yup, because is the filesystem

utils/fs/memory/memory.go
```diff
@@ -0,0 +1,333 @@
...
+
+func (fs *Memory) Base() string {
+	return fs.base
+}
+
+type file struct {
+	fs.BaseFile
+
+	c     *content
+	p     int64
```
mcuadros: done

utils/fs/memory/memory.go
```diff
@@ -0,0 +1,333 @@
...
+func (fs *Memory) Stat(filename string) (fs.FileInfo, error) {
+	fullpath := fs.Join(fs.base, filename)
+
+	if _, ok := fs.s.files[filename]; ok {
+		return newFileInfo(fs.base, fullpath, fs.s.files[filename].c.size), nil
+	}
+
+	info, err := fs.ReadDir(filename)
+	if err == nil && len(info) != 0 {
+		return newFileInfo(fs.base, fullpath, int64(len(info))), nil
```
mcuadros: A dir doesn't exists if is empty

utils/fs/memory/memory.go
```diff
@@ -0,0 +1,333 @@
...
+
+		if len(parts) != 1 {
+			dirs[parts[0]] = true
+			continue
+		}
+
+		entries = append(entries, newFileInfo(fs.base, fullpath, f.c.size))
+	}
+
+	for path := range dirs {
```
mcuadros: done

utils/fs/memory/memory.go
```diff
@@ -0,0 +1,333 @@
...
+		})
+	}
+
+	return
+}
+
+func (fs *Memory) TempFile(dir, prefix string) (fs.File, error) {
+	filename := fmt.Sprintf("%s_%d", prefix, time.Now().UnixNano())
+	fullpath := fs.Join(fs.base, dir, filename)
+	return fs.Create(fullpath)
```
mcuadros: I did another solution

utils/fs/memory/memory.go
```diff
@@ -0,0 +1,333 @@
...
+		return 0, errors.New("read not supported")
+	}
+
+	n, err := f.c.ReadAt(b, f.p)
+	f.p += int64(n)
+
+	return n, err
+}
+
+func (f *file) Seek(offset int64, whence int) (int64, error) {
```
mcuadros: done

utils/fs/memory/memory.go
```diff
@@ -0,0 +1,333 @@
...
+}
+
+func (fs *Memory) Open(filename string) (fs.File, error) {
+	return fs.OpenFile(filename, os.O_RDONLY, 0)
+}
+
+func (fs *Memory) OpenFile(filename string, flag int, perm os.FileMode) (fs.File, error) {
+	fullpath := fs.Join(fs.base, filename)
+	f, ok := fs.s.files[fullpath]
+	if !ok && flag&os.O_CREATE == 0 {
```
mcuadros: I really love this kind of abstractions, but this is more a general problem, not only related with this package.

utils/fs/memory/memory.go
mcuadros: errr, I don't like, is everything is part of the same thing, and since you don't have test only for the files, is weird have a `file.go` without the `file_test.go`

utils/fs/memory/memory.go
```diff
@@ -39,7 +40,7 @@ func (fs *Memory) Open(filename string) (fs.File, error) {
 func (fs *Memory) OpenFile(filename string, flag int, perm os.FileMode) (fs.File, error) {
 	fullpath := fs.Join(fs.base, filename)
 	f, ok := fs.s.files[fullpath]
-	if !ok && flag&os.O_CREATE == 0 {
+	if !ok && !isCreate(flag) {
```
alcortesm: you are not checking if flag *is* O_CREATE but if flags *has* O_CREATE set. I would rename all the check functions to hasFoo: hasCreate(flag), hasAppend(flag)... And I would create the negate ones also: hasNotCreate(flag), hasNotAppend(flag)...

utils/fs/memory/memory.go
```diff
@@ -89,26 +90,25 @@ func (fs *Memory) ReadDir(base string) (entries []fs.FileInfo, err error) {
...
+		appendedDirs[parts[0]] = true
 	}
 
 	return
 }
 
 func (fs *Memory) TempFile(dir, prefix string) (fs.File, error) {
-	filename := fmt.Sprintf("%s_%d", prefix, time.Now().UnixNano())
+	fs.tempCount++
+	filename := fmt.Sprintf("%s_%d_%d", prefix, fs.tempCount, time.Now().UnixNano())
```
alcortesm: This does not solve the problem.

What is the purpose of TempFile()? If its purpose is to create a new file, then a new file has to be created, instead of truncating a file with a highly uncommon name. Also explain what this method is supposed to do and why it works differently than the standard TempFile method.

utils/fs/memory/memory.go
```diff
@@ -313,6 +291,16 @@ func (c *content) WriteAt(p []byte, off int64) (int, error) {
 	return n, nil
 }
 
+func (c *content) grow(off, l int) {
+	if int(off+l) <= cap(c.bytes) {
+		return
+	}
+
+	buf := make([]byte, 2*len(c.bytes)+l)
```
alcortesm: Don't use length and capacity interchangeably.

Even if in this code they always have the same value, they represent quite different things. Choose one and stick to it. The one you are looking for is capacity, not length.

utils/fs/memory/memory.go
```diff
@@ -298,12 +281,7 @@ type content struct {
 }
 
 func (c *content) WriteAt(p []byte, off int64) (int, error) {
```
alcortesm: There is no need for grow or copy here, just use the append function provided by the language. And remove size, this is what the length of a slice is for. You are rewriting here all the internal code of a slice.

utils/fs/memory/memory.go
```diff
@@ -323,11 +311,35 @@ func (c *content) ReadAt(b []byte, off int64) (int, error) {
...
 	}
 
-	n := copy(b, c.bytes[off:l])
+	n := copy(b, c.bytes[off:off+l])
 	return n, nil
 }
 
 func (c *content) Truncate() {
-	c.bytes = []byte{}
+	c.bytes = make([]byte, 0)
```
alcortesm: Do we really want to throw away the preallocated buffer when truncating, we are wasting resources here.

PR #116, opened 2016-11-06
utils/fs: Fix O_CREATE flag check in OpenFile

jinroh:
Flag checking is using a `|` instead of `&` against `os.O_CREATE`.


mcuadros:
cool, how you found the issue? this requires a test @smola

jinroh:
I just reviewed the code changes from rc1 to rc2, and it caught my eyes.

smola:
Tests added. Thanks @jinroh!

PR #118, opened 2016-11-07
new plumbing package

mcuadros:
This introduce a new plumbing package, this package contains all the core objects, and formats and clients packages. The goal is to introduce the logic division in git of plumbing and porcelain. All the package outside of the package plumbing are consider porcelain. 

The main reason is because after think about where put the new `IndexStorer`, I figure out, that the Index, Objects, clients are consider plumbing, and instead of having a cross reference between two package of same level as `core` and `formats/index`


mcuadros:
PTAL @alcortesm

common_test.go
```diff
@@ -6,12 +6,12 @@ import (
 	"os"
 	"testing"
 
-	"gopkg.in/src-d/go-git.v4/clients"
-	"gopkg.in/src-d/go-git.v4/clients/common"
-	"gopkg.in/src-d/go-git.v4/core"
+	"gopkg.in/src-d/go-git.v4/plumbing/clients"
+	"gopkg.in/src-d/go-git.v4/plumbing/clients/common"
+	"gopkg.in/src-d/go-git.v4/plumbing"
```
smola: gofmt missing?

PR #119, opened 2016-11-08
commit_walker: simplify implementation.

smola:
* No recursive calls.
* Check if seen before adding to the queue.


commit_walker.go
```diff
@@ -6,62 +6,40 @@ import (
...
 
-type commitWalker struct {
-	seen  map[core.Hash]bool
-	stack []*CommitIter
-	start *Commit
-	cb    func(*Commit) error
-}
-
 // WalkCommitHistory walks the commit history
 func WalkCommitHistory(c *Commit, cb func(*Commit) error) error {
```
alcortesm: An awesome and very welcome simplification, thanks!

Using a queue here is quite dangerous though, as the garbage collector would not be able to free any of the already processed commits.  Consider using a stack instead, setting already processed commits' pointers to nil.

mcuadros:
@smola what we do with this PR?

mcuadros:
@smola can we retake this PR? I found interesting avoid having large stacks

smola:
@mcuadros I'm playing around in a new branch (https://github.com/smola/go-git/commits/new-iters) to unify iterators implementations (or, at least, tree walkers), including TreeWalker, CommitWalker and the new difftree iterators.

alcortesm:
+250 commits, lol.

PR #121, opened 2016-11-11
Add configurable http client factory (fixes #120)

### LINKED ISSUE #120: client http, support SSL insecure connections
As did @StupidHod we need to have optional support for insecure SSL connections:
https://github.com/StupidHod/go-git/commit/ed219250e2a3be58f8b75a47fc97ef7996924be6#diff-8c989aecb091aec747030aaddeaee67aR29

ferhatelmas:
* new http client factory ready to install/override default http(s)
 * mv GitUploadPackServiceFactory to clients.common pkg
 * rename http.HTTPError to http.Err
 * rename http.HTTPAuthMethod to http.AuthMethod
 * add doc and examples/ usage
 * general improvements:
   - update install link in readme to v4 (example are already pointing v4)
   - fix indentation in package doc (styling for godoc.org)
   - use http.Status constants instead of integers
   - close leaked response body
   - rm named returns which stutter in doc
   - fix one format string
   - rm unnecessary if checks
   - documentation fixes

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@ Installation
 The recommended way to install *go-git* is:
 
 ```
-go get -u gopkg.in/src-d/go-git.v3/...
+go get -u gopkg.in/src-d/go-git.v4/...
 ```
 
 
--- a/doc.go
+++ b/doc.go
@@ -6,31 +6,32 @@
 // extensions.
 //
 // Small example extracting the commits from a repository:
-// func ExampleBasic_printCommits() {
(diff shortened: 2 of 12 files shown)


mcuadros:
Can you provide an example in examples folder or maybe at package level? or even better both? 
Looks nice I like the way of having your or client, the initial goal was that, but we forgot in the road :D

ferhatelmas:
@mcuadros added examples (they don't do anything fancy but shows how client is customizable).

PR #127, opened 2016-11-16
transport: create Client interface

smola:
* move plumbing/client package to plumbing/transport.
* a Client can instantiate any client transport service.
* InstallProtocol installs a Client for a given protocol,
  instead of just a UploadPackService.
* Renamed GitUploadPackService to UploadPackService.


plumbing/transport/common.go
```diff
@@ -11,32 +11,31 @@
 // Each protocol has its own implementation of
 // `NewGitUploadPackService`, but you should generally not use them
 // directly, use this package's `NewGitUploadPackService` instead.
```
alcortesm: This comments are outdated.

plumbing/transport/common.go
```diff
@@ -11,32 +11,31 @@
...
 	Protocols[scheme] = f
 }
 
-// NewGitUploadPackService returns the appropriate upload pack service
-// among of the set of known protocols: HTTP, SSH. See `InstallProtocol`
-// to add or modify protocols.
-func NewGitUploadPackService(endpoint common.Endpoint) (common.GitUploadPackService, error) {
+// NewClient returns the appropriate client among of the set of known protocols:
+// HTTP, SSH. See `InstallProtocol` to add or modify protocols.
+func NewClient(endpoint common.Endpoint) (common.Client, error) {
```
alcortesm: I don't understand the point behind these changes.

Before: a client package that creates new "services": clients.NewGitUploadPackService()
After: a transport package that creates new "clients": transport.NewClient()

None of these things make much sense, maybe we should explain better the purpose of the package. Also the names of the files don't help: a common.go file, a common directory, with another common.go file inside. The word "common" is meaningless and should be avoided in general. Even if there is a reason to call something common, I doubt there is a reason to name 3 different things like that in the same package.

smola:
Closed in favor of: https://github.com/src-d/go-git/pull/132

PR #131, opened 2016-11-22
plumbing/packfile: PACK encoder

ajnavarro:
- Added simple PACK encoder, deltas not supported by now

### DIFF
--- a/plumbing/format/packfile/encoder.go
+++ b/plumbing/format/packfile/encoder.go
@@ -1,16 +1,14 @@
 package packfile
 
 import (
+	"compress/zlib"
 	"crypto/sha1"
 	"fmt"
-	"hash"
 	"io"
 
 	"gopkg.in/src-d/go-git.v4/plumbing"
 	"gopkg.in/src-d/go-git.v4/plumbing/storer"
 	"gopkg.in/src-d/go-git.v4/utils/binary"
-
-	"github.com/klauspost/compress/zlib"
 )
 
 // Encoder gets the data from the storage and write it into the writer in PACK
@@ -19,20 +17,22 @@ type Encoder struct {
 	storage storer.ObjectStorer
 	w       io.Writer
 	zw      *zlib.Writer
-	hash    hash.Hash
+	hasher  plumbing.Hasher
 }
 
 // NewEncoder creates a new packfile encoder using a specific Writer and
 // ObjectStorer
 func NewEncoder(w io.Writer, s storer.ObjectStorer) *Encoder {
-	h := sha1.New()
+	h := plumbing.Hasher{
+		Hash: sha1.New(),
+	}
 	mw := io.MultiWriter(w, h)
 	zw := zlib.NewWriter(mw)
 	return &Encoder{
 		storage: s,
 		w:       mw,
 		zw:      zw,
-		hash:    h,
+		hasher:  h,
 	}
 }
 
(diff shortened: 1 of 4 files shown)


plumbing/format/packfile/encoder.go
```diff
@@ -0,0 +1,103 @@
...
+
+// Encoder gets the data from the storage and write it into the writer in PACK format
+type Encoder struct {
+	storage storer.ObjectStorer
+	w       io.Writer
+	hash    hash.Hash
+}
+
+// Creates a new packfile Encoder using a specific Writer
+func NewEncoder(w io.Writer, storage storer.ObjectStorer) *Encoder {
```
mcuadros: use short variable names to very local variables, in this case `s`

plumbing/format/packfile/encoder.go
```diff
@@ -0,0 +1,103 @@
...
+	h := sha1.New()
+	mw := io.MultiWriter(w, h)
+	return &Encoder{storage, mw, h}
```
mcuadros: don't use unlabeled fields initializations

plumbing/format/packfile/encoder.go
```diff
@@ -0,0 +1,115 @@
...
+	return &Encoder{
+		storage: s,
+		w:       mw,
+		hash:    h,
+	}
+}
+
+// Encode encodes objects specified using a list of hashes. This objects must
+// exists into the storer
+func (e *Encoder) Encode(hashes []plumbing.Hash) (plumbing.Hash, error) {
```
alcortesm: From the signature of the method and the comments it is not clear if you can create a **single** packfile by calling several times in a row to **Encode**.

I would change the comment here to clarify that every time you call Encode it **creates a new packfile** and write it to the writer. Something like this:

```
// Encode creates a packfile containing all the objects referenced in hashes and writes it to the writer in the Encoder.
```

plumbing/format/packfile/encoder.go
```diff
@@ -0,0 +1,115 @@
...
+	if t == plumbing.OFSDeltaObject || t == plumbing.REFDeltaObject {
+		// TODO implements delta objects
+		return fmt.Errorf("delta object not supported: %v", t)
+	}
+
+	if err := e.entryHead(t, o.Size()); err != nil {
+		return err
+	}
+
+	zw := zlib.NewWriter(e.w)
```
alcortesm: A nice optimization would be for the encoder to have a **permanent** zlib.Writer, instead of creating a new one every time an entry is compressed, so we can reuse the dictionary.  Would it be possible to do this using `zlib.Reset()`?

plumbing/format/packfile/encoder.go
```diff
@@ -0,0 +1,118 @@
...
+	"github.com/klauspost/compress/zlib"
+)
+
+// Encoder gets the data from the storage and write it into the writer in PACK
+// format
+type Encoder struct {
+	storage storer.ObjectStorer
+	w       io.Writer
+	zw      *zlib.Writer
+	hash    hash.Hash
```
mcuadros: use our Hasher, https://godoc.org/gopkg.in/src-d/go-git.v4/plumbing#Hasher

plumbing/format/packfile/encoder.go
```diff
@@ -0,0 +1,118 @@
...
+	"crypto/sha1"
+	"fmt"
+	"hash"
+	"io"
+
+	"gopkg.in/src-d/go-git.v4/plumbing"
+	"gopkg.in/src-d/go-git.v4/plumbing/storer"
+	"gopkg.in/src-d/go-git.v4/utils/binary"
+
+	"github.com/klauspost/compress/zlib"
```
mcuadros: why? we are using "compress/zlib" in the Scanner

PR #132, opened 2016-11-22
transport: create Client interface

smola:
* plumbing: move plumbing/client package to plumbing/transport.
* a Client can instantiate any client transport service.
* InstallProtocol installs a Client for a given protocol, instead of just a UploadPackService.
* A Client can open a session for fetch-pack or send-pack for a specific Endpoint.
* Adapt ssh and http clients to the new client interface.


plumbing/transport/common.go
```diff
@@ -1,46 +1,114 @@
...
+type AuthMethod interface {
+	fmt.Stringer
+	Name() string
+}
+
+// FetchPackSession represents a git-fetch-pack session.
+type FetchPackSession interface {
+	Session
+	// AdvertisedReferences retrieves the advertised references for a
+	// repository. It must be called before Pack.
```
alcortesm: s/Pack/FetchPack/

alcortesm: > It must be called before Pack.

we should briefly explain why: is this because the protocol says so or because some internal limitation in our code?

plumbing/transport/common.go
```diff
@@ -1,46 +1,114 @@
 // Package clients includes the implementation for different transport protocols
```
smola: Updated docs, since they were outdated (not just the package name).

plumbing/transport/common.go
smola: Done. Added further clarifications.
It SHOULD be called before FetchPack and it MUST NOT be called after FetchPack. This is to respect the protocol, so that one of our "Sessions" match a single git-fetch-pack execution.

Issue #123, opened 2016-11-14
cloning an empty repository returns unexpected error


smola:
Cloning an empty repository such as https://github.com/git-fixtures/empty returns error: `unexpected client error: pkt-line 3: cannot read hash, pkt-line too short`

mcuadros:
Fixed https://github.com/src-d/go-git/pull/136

PR #136, opened 2016-11-23
plumbing/transport: add common tests and fixes.

smola:
* add common test suite for different transport implementations.
* fix different behaviour on error handling for ssh and http.
  fixes issue #123.
* support detecting unexisting repositories with SSH + GitHub/Bitbucket
  (apparently, there is no standard for all SSH servers).
* remove ssh.NewClient (only DefaultClient makes sense at the moment).
* make ssh.Client and http.Client private.
* transport: test actual objects fetched, not just packfile size.
* utils/ioutil: utilities to work with io interfaces.

### DIFF
--- a/plumbing/transport/fetch_pack.go
+++ b/plumbing/transport/fetch_pack.go
@@ -129,13 +129,13 @@ func (r *UploadPackRequest) Have(h ...plumbing.Hash) {
 }
 
 func (r *UploadPackRequest) IsEmpty() bool {
-	if len(r.Haves) < len(r.Wants) {
-		return false
-	}
+	return isSubset(r.Wants, r.Haves)
+}
 
-	for _, h := range r.Wants {
+func isSubset(needle []plumbing.Hash, haystack []plumbing.Hash) bool {
+	for _, h := range needle {
 		found := false
-		for _, oh := range r.Haves {
+		for _, oh := range haystack {
 			if h == oh {
 				found = true
 				break
--- a/plumbing/transport/fetch_pack.go
+++ b/plumbing/transport/fetch_pack.go
@@ -33,7 +33,7 @@ func (i *UploadPackInfo) Decode(r io.Reader) error {
 	ar := advrefs.New()
 	if err := d.Decode(ar); err != nil {
 		if err == advrefs.ErrEmpty {
-			return plumbing.NewPermanentError(err)
+			return err
 		}
 		return plumbing.NewUnexpectedError(err)
 	}
@@ -126,6 +126,28 @@ func (r *UploadPackRequest) Want(h ...plumbing.Hash) {
 
 func (r *UploadPackRequest) Have(h ...plumbing.Hash) {
 	r.Haves = append(r.Haves, h...)
+}
+
+func (r *UploadPackRequest) IsEmpty() bool {
+	if len(r.Haves) < len(r.Wants) {
+		return false
+	}
+
+	for _, h := range r.Wants {
+		found := false
+		for _, oh := range r.Haves {
+			if h == oh {
+				found = true
+				break
+			}
+		}
+
+		if !found {
+			return false
+		}
+	}
+
+	return true
 }
 
 func (r *UploadPackRequest) String() string {
--- a/plumbing/transport/ssh/common.go
+++ b/plumbing/transport/ssh/common.go
@@ -11,32 +11,22 @@ import (
 	"golang.org/x/crypto/ssh"
 )
 
-// New errors introduced by this package.
 var (
-	ErrAdvertistedReferencesAlreadyCalled = errors.New("cannot call AdvertisedReference twice")
-	ErrAlreadyConnected                   = errors.New("ssh session already created")
-	ErrAuthRequired                       = errors.New("cannot connect: auth required")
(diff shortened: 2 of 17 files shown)


plumbing/transport/ssh/common.go
```diff
@@ -11,32 +11,22 @@ import (
...
-// New errors introduced by this package.
 var (
-	ErrAdvertistedReferencesAlreadyCalled = errors.New("cannot call AdvertisedReference twice")
-	ErrAlreadyConnected                   = errors.New("ssh session already created")
-	ErrAuthRequired                       = errors.New("cannot connect: auth required")
-	ErrNotConnected                       = errors.New("not connected")
-	ErrUploadPackAnswerFormat             = errors.New("git-upload-pack bad answer format")
-	ErrUnsupportedVCS                     = errors.New("only git is supported")
-	ErrUnsupportedRepo                    = errors.New("only github.com is supported")
+	errAlreadyConnected = errors.New("ssh session already created")
```
mcuadros: why private?

utils/ioutil/common.go
```diff
@@ -0,0 +1,52 @@
...
+	io.Reader
+	closer io.Closer
+}
+
+func (r *readCloser) Close() error {
+	return r.closer.Close()
+}
+
+// NewReadCloser creates an `io.ReadCloser` with the given `io.Reader` and
+// `io.ReadCloser`.
```
mcuadros: it the last word is wrong, I guess should say `io.Closer`

mcuadros:
https://github.com/src-d/go-git/issues/123 is this issue fixed here?

plumbing/transport/fetch_pack.go
```diff
@@ -128,6 +128,28 @@ func (r *UploadPackRequest) Have(h ...plumbing.Hash) {
 	r.Haves = append(r.Haves, h...)
 }
 
+func (r *UploadPackRequest) IsEmpty() bool {
+	if len(r.Haves) < len(r.Wants) {
+		return false
+	}
```
alcortesm: From https://github.com/git/git/blob/master/Documentation/technical/pack-protocol.txt :
> Clients MUST NOT mention an obj-id in a 'want' command which did not appear in the response obtained through ref discovery.

As the `UploadPackRequest.Want` does not enforce this requisite we cannot count on it. This means we cannot relay on lines 132-134 for an early return.

plumbing/transport/ssh/common.go
smola: It's an internal error that can should not happen calling only public methods. connect() is called in newFetchPackSession and nowhere else. The user cannot make call it twice.

smola:
@mcuadros Yes. This fixes #123.

plumbing/transport/fetch_pack.go
```diff
@@ -128,6 +128,28 @@ func (r *UploadPackRequest) Have(h ...plumbing.Hash) {
...
+
+	for _, h := range r.Wants {
+		found := false
+		for _, oh := range r.Haves {
+			if h == oh {
+				found = true
+				break
+			}
+		}
```
alcortesm: Given the use we are doing of `UploadPackRequest.Wants` and  `UploadPackRequest.Haves` we should turn them into maps (actually, sets).  Also change the name of `h` or `w` and `oh` to `have`.

This two nested for loops can be clarified a lot: what we are doing is checking if `r.Wants` is a subset of `r.Haves`. So I would substitute the loop for:

```
if areSetSubset(r.Haves, r.Wants) {
    return true
}
```

and create the function `areSetSubset` somewhere.

We can also define a new type for a list of references (wants and haves) and define some common set operations like `complementary set` and `is subset`.

Also, I think `r.Wants` and `r.Haves` should be private, unless I am missing something.

plumbing/transport/fetch_pack.go
```diff
@@ -128,6 +128,28 @@ func (r *UploadPackRequest) Have(h ...plumbing.Hash) {
...
+
+		if !found {
+			return false
+		}
+	}
+
+	return true
+}
+
```
alcortesm: Probably unrelated with this PR, but something we must address at some point.

The `UploadPackRequest` should probably use a `ulreq.Ulreq`.

plumbing/transport/fetch_pack.go
alcortesm: A better solution would be to enforce it in the Want method, of course.

plumbing/transport/fetch_pack.go
mcuadros: lets do it now, and kill this duplicate logic

plumbing/transport/fetch_pack.go
smola: We do not perform an early return here. The early return is when IsEmpty = false. So, lines 132-134 are just a short-circuit in IsEmpty that avoids comparing every reference (unless the same reference is repeated multiple times in Wants.

plumbing/transport/fetch_pack.go
smola: Ok. Adding the areSetSubset.

r.Wants and r.Haves need to be public (with the current code). Let's revisit this in the next PR (refactor of UploadPack messsages).

Issue #137, opened 2016-11-24
reference solver


antham:
Hello,

I wrote a feature here : https://github.com/antham/gommit/blob/master/reference/reference_scanner.go to be able to solve this kind of reference ```HEAD~2^^```, I was wondering if it could be interesting for your project to have that, I could send a PR.

mcuadros:
very cool gommit, great to see that you are using go-git. 

The functionality cover by your ` reference` package is really relevant for go-git. So we will be more than happy to have a PR about this. Great job!

It requires a bit few job, cannot be merged directly because we need to agree the API, the current API should be improved. So if you want we can work on it, if you thing is to much work, we can do a cherry-pick on work on top of your work.  (I prefer the first options)

As a start, can you tell me the documentation you took as reference to implement it? I was ready reading https://www.kernel.org/pub/software/scm/git/docs/gitrevisions.html, looks like is called `Revision`

antham:
Yep I'was wondering about api to use in your project as well. 

I did it based on thing I found around and based only on the fact that I needed to solve ~ and ^, so some elements are missing, naming as you pointed out isn't correct, it's a good opportunity to improve that. About branching name, I followed this post on stackoveflow http://stackoverflow.com/questions/3651860/which-characters-are-illegal-within-a-branch-name.

mcuadros:
Ok, great, your help is highly appreciated.

We can go step by step with the PRs, first you can submit one with the Parser/Tokenizer/etc and the Resolver and after we can see the rev list functionality (aka FetchCommitInterval) because we are devoloping something similar.

My proposal is the following:
- Implement `Revision` entity as `type Revision string`  and not as plain string 
- Change the signature of `FetchCommitByID` to:`ResolveReference(r *git.Repository, rev Revision) (plumbing.Hash, error)`

- Place all the parser, tokenizer, etc in an internal package called `revision` at the root of the package. This package deals with strings, not with the type `Revision`
- Review the parser/tokenizer, to be formal with the """spec""" described here https://www.kernel.org/pub/software/scm/git/docs/gitrevisions.html, (or if you find a better source, this one), if you find something hard and arcane with can put a TODO, is not required implement all the functionality 

What do you thing? Thanks.

antham:
So if I get it right, you want to be able to create a revision and it's when you call ResolveReference against a revision that you obtain the hash tied to the revision. So everything is done in this method by calling stuffs defined in a revision package.

It's fine for me, I will do that.

PR #139, opened 2016-11-24
Add revision implementation

antham:
* Implement Revision entity as type Revision string and not as plain string

* Add a method ResolveReference(r *git.Repository, rev Revision) (plumbing.Hash, error)

* Place all the parser, tokenizer, etc in an internal package called revision at the root of the package. This package deals with strings, not with the type Revision

* Define a revision parser/tokenizer to comply with the """spec""" described here https://www.kernel.org/pub/software/scm/git/docs/gitrevisions.html (if something is hard and arcane this can be put in a TODO, is not required implement all the functionality)

* Resource for valid reference names : https://git-scm.com/docs/git-check-ref-format

### DIFF
--- a/internal/revision/parser.go
+++ b/internal/revision/parser.go
@@ -213,6 +213,18 @@ func (p *parser) validateFullRevision(chunks *[]revisioner) error {
 			}
 
 			return &ErrInvalidRevision{`":" statement is not valid, could be : :/<regexp>`}
+		case colonPath:
+			if i == len(*chunks)-1 && hasReference || len(*chunks) == 1 {
+				return nil
+			}
+
+			return &ErrInvalidRevision{`":" statement is not valid, could be : <revision>:<path>`}
+		case colonStagePath:
+			if len(*chunks) == 1 {
+				return nil
+			}
+
+			return &ErrInvalidRevision{`":" statement is not valid, could be : :<n>:<path>`}
 		}
 	}
 
--- a/internal/revision/parser.go
+++ b/internal/revision/parser.go
@@ -14,6 +14,9 @@ func (e *ErrInvalidRevision) Error() string {
 	return "Revision invalid : " + e.s
 }
 
+// ref represents a reference name
+type ref string
+
 // parser represents a parser.
 type parser struct {
 	s   *scanner
@@ -73,7 +76,7 @@ func (p *parser) parseRevSuffix() ([]string, error) {
 }
 
 // parseRef extract reference name
-func (p *parser) parseRef() (string, error) {
+func (p *parser) parseRef() (ref, error) {
 	var tok token
 	var prevTok token
 	var lit string
@@ -95,7 +98,7 @@ func (p *parser) parseRef() (string, error) {
 		}
 
 		if endOfRef {
-			return buf, nil
+			return ref(buf), nil
 		}
 
 		buf += lit
--- a/internal/revision/parser.go
+++ b/internal/revision/parser.go
@@ -15,12 +15,12 @@ func (e *ErrInvalidRevision) Error() string {
 	return "Revision invalid : " + e.s
 }
 
+// revisioner represents a revision component
+type revisioner interface {
+}
+
 // ref represents a reference name
 type ref string
-
-// suffixer represents a generic suffix
-type suffixer interface {
-}
 
 // tildeSuffix represents ~, ~{n}
 type tildeSuffixPath struct {
@@ -96,20 +96,20 @@ func (p *parser) scan() (tok token, lit string) {
 func (p *parser) unscan() { p.buf.n = 1 }
 
 // parseAtSuffix extract part following @
-func (p *parser) parseAtSuffix() (suffixer, error) {
+func (p *parser) parseAtSuffix() (revisioner, error) {
 	var tok, nextTok token
 	var lit, nextLit string
 
 	tok, lit = p.scan()
 
 	if tok != at {
-		return (suffixer)(struct{}{}), &ErrInvalidRevision{fmt.Sprintf(`"%s" found must be "@"`, lit)}
+		return (revisioner)(struct{}{}), &ErrInvalidRevision{fmt.Sprintf(`"%s" found must be "@"`, lit)}
 	}
 
 	tok, lit = p.scan()
 
 	if tok != obrace {
-		return (suffixer)(struct{}{}), &ErrInvalidRevision{fmt.Sprintf(`"%s" found must be "{" after @`, lit)}
+		return (revisioner)(struct{}{}), &ErrInvalidRevision{fmt.Sprintf(`"%s" found must be "{" after @`, lit)}
 	}
 
 	tok, lit = p.scan()
@@ -124,7 +124,7 @@ func (p *parser) parseAtSuffix() (suffixer, error) {
 		n, err := strconv.Atoi(lit)
 
 		if err != nil {
-			return []suffixer{}, &ErrInvalidRevision{fmt.Sprintf(`"%s" is not a number`, lit)}
+			return []revisioner{}, &ErrInvalidRevision{fmt.Sprintf(`"%s" is not a number`, lit)}
 		}
 
 		return atSuffixReflog{n}, nil
@@ -132,7 +132,7 @@ func (p *parser) parseAtSuffix() (suffixer, error) {
 		n, err := strconv.Atoi(nextLit)
 
 		if err != nil {
-			return []suffixer{}, &ErrInvalidRevision{fmt.Sprintf(`"%s" is not a number`, nextLit)}
+			return []revisioner{}, &ErrInvalidRevision{fmt.Sprintf(`"%s" is not a number`, nextLit)}
 		}
 
 		t, _ := p.scan()
@@ -144,18 +144,18 @@ func (p *parser) parseAtSuffix() (suffixer, error) {
 		return atSuffixCheckout{n}, nil
 	}
 
-	return (suffixer)(struct{}{}), &ErrInvalidRevision{fmt.Sprintf(`invalid expression "%s" in @{} structure`, lit)}
+	return (revisioner)(struct{}{}), &ErrInvalidRevision{fmt.Sprintf(`invalid expression "%s" in @{} structure`, lit)}
 }
 
 // parseTildeSuffix extract part following tilde
-func (p *parser) parseTildeSuffix() (suffixer, error) {
+func (p *parser) parseTildeSuffix() (revisioner, error) {
 	var tok token
 	var lit string
 
 	tok, lit = p.scan()
 
 	if tok != tilde {
-		return (suffixer)(struct{}{}), &ErrInvalidRevision{fmt.Sprintf(`"%s" found must be "~"`, lit)}
+		return (revisioner)(struct{}{}), &ErrInvalidRevision{fmt.Sprintf(`"%s" found must be "~"`, lit)}
 	}
 
 	tok, lit = p.scan()
@@ -165,7 +165,7 @@ func (p *parser) parseTildeSuffix() (suffixer, error) {
 		n, err := strconv.Atoi(lit)
 
 		if err != nil {
-			return (suffixer)(struct{}{}), &ErrInvalidRevision{fmt.Sprintf(`"%s" is not a number`, lit)}
+			return (revisioner)(struct{}{}), &ErrInvalidRevision{fmt.Sprintf(`"%s" is not a number`, lit)}
 		}
 
 		return tildeSuffixPath{n}, nil
@@ -173,19 +173,19 @@ func (p *parser) parseTildeSuffix() (suffixer, error) {
 		p.unscan()
 		return tildeSuffixPath{1}, nil
 	default:
-		return (suffixer)(struct{}{}), &ErrInvalidRevision{fmt.Sprintf(`"%s" is not a valid revision suffix component`, lit)}
+		return (revisioner)(struct{}{}), &ErrInvalidRevision{fmt.Sprintf(`"%s" is not a valid revision suffix component`, lit)}
 	}
 }
 
 // parseCaretSuffix extract part following caret
-func (p *parser) parseCaretSuffix() (suffixer, error) {
+func (p *parser) parseCaretSuffix() (revisioner, error) {
 	var tok token
 	var lit string
 
 	tok, lit = p.scan()
 
 	if tok != caret {
-		return (suffixer)(struct{}{}), &ErrInvalidRevision{fmt.Sprintf(`"%s" found must be "^"`, lit)}
+		return (revisioner)(struct{}{}), &ErrInvalidRevision{fmt.Sprintf(`"%s" found must be "^"`, lit)}
 	}
 
 	tok, lit = p.scan()
@@ -197,7 +197,7 @@ func (p *parser) parseCaretSuffix() (suffixer, error) {
 		r, err := p.parseCaretSuffixWithBraces()
 
 		if err != nil {
-			return (suffixer)(struct{}{}), err
+			return (revisioner)(struct{}{}), err
 		}
 
 		return r, nil
@@ -205,7 +205,7 @@ func (p *parser) parseCaretSuffix() (suffixer, error) {
 		n, err := strconv.Atoi(lit)
 
 		if err != nil {
-			return []suffixer{}, &ErrInvalidRevision{fmt.Sprintf(`"%s" is not a number`, lit)}
+			return []revisioner{}, &ErrInvalidRevision{fmt.Sprintf(`"%s" is not a number`, lit)}
 		}
 
 		return caretSuffixPath{n}, nil
@@ -213,13 +213,13 @@ func (p *parser) parseCaretSuffix() (suffixer, error) {
 		p.unscan()
 		return caretSuffixPath{1}, nil
 	default:
-		return (suffixer)(struct{}{}), &ErrInvalidRevision{fmt.Sprintf(`"%s" is not a valid revision suffix component`, lit)}
+		return (revisioner)(struct{}{}), &ErrInvalidRevision{fmt.Sprintf(`"%s" is not a valid revision suffix component`, lit)}
 	}
 }
 
 // parseCaretSuffixWithBraces extract suffix between braces following caret
 // todo : add regexp checker
-func (p *parser) parseCaretSuffixWithBraces() (suffixer, error) {
+func (p *parser) parseCaretSuffixWithBraces() (revisioner, error) {
 	var tok, nextTok token
 	var lit, _ string
 	start := true
@@ -228,7 +228,7 @@ func (p *parser) parseCaretSuffixWithBraces() (suffixer, error) {
 	tok, lit = p.scan()
 
 	if tok != obrace {
-		return []suffixer{}, &ErrInvalidRevision{fmt.Sprintf(`"%s" found must be "{" after ^`, lit)}
+		return []revisioner{}, &ErrInvalidRevision{fmt.Sprintf(`"%s" found must be "{" after ^`, lit)}
 	}
 
 	for {
@@ -245,11 +245,11 @@ func (p *parser) parseCaretSuffixWithBraces() (suffixer, error) {
 		case reg.re == "" && tok == emark && nextTok == minus:
 			reg.negate = true
 		case reg.re == "" && tok == emark:
-			return (suffixer)(struct{}{}), &ErrInvalidRevision{fmt.Sprintf(`revision suffix brace component sequences starting with "/!" others than those defined are reserved`)}
+			return (revisioner)(struct{}{}), &ErrInvalidRevision{fmt.Sprintf(`revision suffix brace component sequences starting with "/!" others than those defined are reserved`)}
 		case reg.re == "" && tok == slash:
 			p.unscan()
 		case tok != slash && start:
-			return (suffixer)(struct{}{}), &ErrInvalidRevision{fmt.Sprintf(`"%s" is not a valid revision suffix brace component`, lit)}
+			return (revisioner)(struct{}{}), &ErrInvalidRevision{fmt.Sprintf(`"%s" is not a valid revision suffix brace component`, lit)}
 		case tok != cbrace:
 			p.unscan()
 			reg.re += lit
@@ -263,7 +263,7 @@ func (p *parser) parseCaretSuffixWithBraces() (suffixer, error) {
 }
 
 // parseRef extract reference name
-func (p *parser) parseRef() (ref, error) {
+func (p *parser) parseRef() (revisioner, error) {
 	var tok, prevTok token
 	var lit, buf string
 	var endOfRef bool
--- a/internal/revision/parser.go
+++ b/internal/revision/parser.go
@@ -28,6 +28,17 @@ type revSuffixPath struct {
 	deep   int
 }
 
+// revSuffixReg represents ^{/foo bar} revision suffix
+type revSuffixReg struct {
+	re     string
+	negate bool
+}
+
+// revSuffixType represents ^{commit} revision suffix
+type revSuffixType struct {
+	object string
+}
+
 // parser represents a parser.
 type parser struct {
 	s   *scanner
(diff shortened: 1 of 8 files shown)


revision/internal/revision/scanner.go
```diff
@@ -0,0 +1,112 @@
+package revision
```
mcuadros: the package should be in the root of the package inside of a internal folder so:
`revision/internal/revision/` ->  `internal/revision/`

mcuadros:
Copy pasted from issue #137 

My proposal is the following:
- Implement `Revision` entity as `type Revision string`  and not as plain string 
- Change the signature of `FetchCommitByID` to:`ResolveReference(r *git.Repository, rev Revision) (plumbing.Hash, error)`

- Place all the parser, tokenizer, etc in an internal package called `revision` at the root of the package. This package deals with strings, not with the type `Revision`
- Review the parser/tokenizer, to be formal with the """spec""" described here https://www.kernel.org/pub/software/scm/git/docs/gitrevisions.html, (or if you find a better source, this one), if you find something hard and arcane with can put a TODO, is not required implement all the functionality

revision/internal/revision/scanner.go
antham: ah ok I thought you meant at the root of a revision package

antham:
@mcuadros If I read properly "revision spec" a revision is something like this "master", "HEAD" but not "master~2" for instance. So Revision type will only describe "master" part of an expression like "master~2" so how do you want to store the whole thing, a Revision type is not enough ?

mcuadros:
Base on the documentation looks like a revision is `HEAD@{5 minutes ago}`, `@{-1}` or just `HEAD`. So you can host all the functionally inside the type `Revision`

Just `HEAD` or `master` is a short reference.

antham:
Ok, it was this kind of sentence that was letting me wondering, "<rev>~<n>, e.g. master~3 : A suffix ~<n> to a revision parameter".

mcuadros:
@antham can you tell us, how far are you with the implementation? You did a lot of work!

antham:
I have at the moment implemented almost everything, you can have a look here what we can parse : https://github.com/src-d/go-git/pull/139/files#diff-0e4bd274259ee67fba0f3f2541704aefR78 , I used what was defined in git revision page to do this test. There are some missing like the one where we need to parse a relative date, I can do this later. I still need to complete parse method to check validity of a whole revision. We need to discuss how to integrate this thing.

mcuadros:
Don't worry we can miss some of the features, the whole library doesn't implement all the arcane features, we implement only the more relevant and well known.
 
@antham can you elaborate the cases, I can't understand it sorry.

antham:
Yep I get it, but it's better to have all bases even if it's no fully completed and not used yet, someone can resume the job if needed instead of starting from scratch which is painful : it's what I wanted to do at least.

I'm sorry but I don't understand what you don't understand :-) @mcuadros

mcuadros:
> I still need to complete parse method to check validity of a whole revision for some cases. We need to discuss how to integrate this thing.

antham:
Ah ok.
 
> I still need to complete parse method to check validity of a whole revision for some cases

I implemented every chunk parser (e.g.: ^3) and I'm almost done with the whole parser (e.g.: master^2~3), we miss some case check to have something complete, it what remains to do.

> We need to discuss how to integrate this thing.

Parser extract from a revision string like this  :  "master~1" a list of chunk like that `[]revisioner{ref("master"),tildePath{1}}`.

Do you have any precise idea how to integrate revision in your system with that ? 
Could we provide the ability for someone to parse a revision string in chunk or do you want to hide this in the system (I think it's a pity) ?  

@mcuadros

antham:
For the parser I think I'm done, we can improve gradually now. 

A review work remains to have everything consistent like error message for instance.

mcuadros:
You deleted, your branch?

All you method and structs are private, you I can recognize any interface, can you document the library and make public the methods.

At the level of the public API, for sure we will need:
- `Repository.ResolveRevision(*Revision) *plumbing.Hash`
- `Repository.ResolveRange(*Range) []*plumbing.Hash`

(the name of the functions are placeholders)

This at least the minimum implementation, and with this we can merge this work

antham:
Nop still here but I rebased it.

There is one interface : revisioner. 

Yep but it would be nice to give the ability to have the result of parsing a revision, it could be useful.

antham:
I forgot to precise but I only implemented revision parser not range parser yet.

mcuadros:
@antham if you create the method `Repository.ResolveRevision(*Revision) *plumbing.Hash` we can review the commit

antham:
Yep I will do that.

antham:
@mcuadros I added ResolveRevision with tilde, caret, regexp and date reference solver. Do you want me to add more or it could be enough for a first go ?

internal/revision/scanner.go
```diff
@@ -0,0 +1,124 @@
...
+		return s.scanWord()
+	}
+
+	if unicode.IsNumber(ch) {
+		s.unread()
+
+		return s.scanNumber()
+	}
+
+	return char, string(ch)
```
alcortesm: This needs some explanation, what is the difference between a `char` here and a `word` (from `scanWord`) and why is char the default case for this function (I would expect an error or something in the default case)?

Maybe you can answer all these question by just explaining in a comment the possible cases of your scanner at the very beginning so people understand the problem before reading your solution: Revision are very varied, and people will need to spend a lot of time reading the spec to understand all the supported types. You already spend all that time when you were writing this file so maybe you can save some people time by explaining that yourself.

internal/revision/scanner.go
```diff
@@ -0,0 +1,124 @@
...
+}
+
+// scanNumber return number token
+func (s *scanner) scanNumber() (token, string) {
+	var data string
+
+	for c := s.read(); c != zeroRune; c = s.read() {
+		if unicode.IsNumber(c) {
+			data += string(c)
```
alcortesm: This is quite expensive and harmful for the garbage collector.

Please refactor this to a nice byte slice and return its string version at the end.

internal/revision/scanner.go
```diff
@@ -0,0 +1,124 @@
...
+			s.unread()
+			return number, data
+		}
+	}
+
+	return number, data
+}
+
+// scanWord return a word token
+func (s *scanner) scanWord() (token, string) {
```
alcortesm: this function is the same code as scanNumber. I wonder if you can avoid repeating the code somehow.

internal/revision/scanner.go
```diff
@@ -0,0 +1,124 @@
...
+	"bufio"
+	"io"
+	"unicode"
+)
+
+var zeroRune = rune(0)
+
+// scanner represents a lexical scanner.
+type scanner struct {
+	r *bufio.Reader
```
alcortesm: I think using a bufio.Reader is the wrong approach here.

The only functionality your are using from a bufio.Reader is the `readRune` and `unreadRune` so using a full bufio.Reader cannot be justified.

Why are you not using a `text.Scanner` instead? (The peek and error reporting seems very helpful to achieve your goals here).

plumbing/revision.go
```diff
@@ -0,0 +1,8 @@
+package plumbing
+
+// Revision represents a git revision
+type Revision string
```
alcortesm: It will be nice to have a pointer here to what a git revision is, maybe an URL.

repository.go
```diff
@@ -453,3 +454,129 @@ func (r *Repository) Reference(name plumbing.ReferenceName, resolved bool) (
 func (r *Repository) References() (storer.ReferenceIter, error) {
 	return r.s.IterReferences()
 }
+
+// ResolveRevision resolves revision to corresponding hash
+func (r *Repository) ResolveRevision(rev plumbing.Revision) (*plumbing.Hash, error) {
+	p := revision.NewParserFromString(string(rev))
```
alcortesm: this does not make a lot of sense.
You have a revision, turn it into a string, then call NewParserFromString.
Why don't you create a parser from the revision directly: `p := rev.NewParser()`.

alcortesm: Also the fact that a revision (`plumbing.Revision`) is not part of the revision package is very confusing:

The revision package does not define a revision (confusing), it defines a parser for a revision instead (confusing).

Nobody should care about revision parsers, as they should be private code, the revision should have methods resolve itself without forcing you to use its internal parser.

So either the name of the package is wrong (revision should be revisionParser) or things have to be refactored to make sense.

internal/revision/parser.go
```diff
@@ -0,0 +1,581 @@
...
+	return
+}
+
+// unscan pushes the previously read token back onto the buffer.
+func (p *Parser) unscan() { p.buf.n = 1 }
+
+// Parse explode a revision string into revisioner chunks
+func (p *Parser) Parse() ([]Revisioner, error) {
+	var rev Revisioner
+	var revs []Revisioner
```
alcortesm: What is the difference between these two values (rev and revs)?

internal/revision/parser.go
```diff
@@ -0,0 +1,581 @@
...
+	var err error
+
+	for {
+		tok, _ := p.scan()
+
+		switch tok {
+		case at:
+			p.unscan()
```
alcortesm: Why are you unreading the `@` here and then discarding it at `parseAt()`? (the same for all the parse... functions below)

internal/revision/parser.go
```diff
@@ -0,0 +1,581 @@
...
+		case eof:
+			err = p.validateFullRevision(&revs)
+
+			if err != nil {
+				return []Revisioner{}, err
+			}
+
+			return revs, nil
+		default:
+			p.unscan()
```
alcortesm: Why not use the string from the p.scan call instead of unscaning the whole thing and reading it back again in parse Ref?

internal/revision/parser.go
```diff
@@ -0,0 +1,581 @@
...
+
+			return &ErrInvalidRevision{`":" statement is not valid, could be : <revision>:<path>`}
+		case ColonStagePath:
+			if len(*chunks) == 1 {
+				return nil
+			}
+
+			return &ErrInvalidRevision{`":" statement is not valid, could be : :<n>:<path>`}
+		}
+	}
```
alcortesm: This is quite thoughtful, thanks for the detailed work involved in making this method!

repository.go
alcortesm: Also all this code should not be part of repository, revisions should have a way to resolve themselves (passing them a repository).

Should this resolver return a slice of hashes instead to cope with range revision (when they are supported in the future)?

internal/revision/scanner.go
```diff
@@ -0,0 +1,124 @@
...
+import (
+	"bufio"
+	"io"
+	"unicode"
+)
+
+var zeroRune = rune(0)
+
+// scanner represents a lexical scanner.
+type scanner struct {
```
antham: I'm not sure to understand what you mean, only parser use a scanner all thoses functions are internals they are not intended to be used outside this package.

internal/revision/parser.go
antham: It's an iterator on a component revision list to validate a full revision, you want me to add more details in function comment ?

internal/revision/parser.go
antham: rev is a single revision component and revs is a list of revision components

internal/revision/parser.go
```diff
@@ -0,0 +1,581 @@
...
+// ErrInvalidRevision is emitted if string doesn't match valid revision
+type ErrInvalidRevision struct {
+	s string
+}
+
+func (e *ErrInvalidRevision) Error() string {
+	return "Revision invalid : " + e.s
+}
+
+// Revisioner represents a revision component
```
antham: A revision component is a part of revision, ( we can see how it works here : https://github.com/src-d/go-git/pull/139/files/e29e554853a1850f42f649b96797c5394d65f2fa#diff-0e4bd274259ee67fba0f3f2541704aefR86), so it's all structs defined after this interface (Ref, TildePath, and so on)

internal/revision/parser.go
```diff
@@ -0,0 +1,581 @@
...
+// Parser represents a parser
+// use to tokenize and transform to revisioner chunks
+// a given string
+type Parser struct {
+	s   *scanner
+	buf struct {
+		tok token
+		lit string
+		n   int
+	}
```
antham: buf stands for buffer so I don't really get the point about this name it's rather comment that is not clear at the beginning.

You are write about this value it's a boolean.

internal/revision/parser.go
antham: It's easier for composition and understanding in my opinion to have functions parsing whole chunks, when I call parseTilde for instance it makes much sense to me to give that "~3" instead of "3"

internal/revision/scanner.go
antham: I didn't because it was small logic that won't move so duplicated this function seems the easiest thing to do

internal/revision/token.go
```diff
@@ -0,0 +1,28 @@
+package revision
+
+// token represents a entity extracted from string parsing
+type token int
+
+const (
+	eof token = iota
+
```
antham: It was only to separate this case from others as a special one

plumbing/revision.go
antham: You mean in comment ?

internal/revision/scanner.go
antham: I will

internal/revision/scanner.go
```diff
@@ -0,0 +1,124 @@
...
+	r *bufio.Reader
+}
+
+// newScanner returns a new instance of scanner.
+func newScanner(r io.Reader) *scanner {
+	return &scanner{r: bufio.NewReader(r)}
+}
+
+// read reads the next rune from the bufferred reader.
+// Returns the rune(0) if an error occurs (or io.EOF is returned).
```
antham: read wraps the error and through scan, stop if there is a trouble, but indeed we can add more context returning the error that was trigger from scan if something different than EOF occured.

internal/revision/scanner.go
antham: I thougth it was narrowed to go language, I will have a look

internal/revision/parser.go
alcortesm: I'm sorry, I was asking the question so you answer it in the code, not here in a comment.

internal/revision/parser.go
alcortesm: Give `buf` a good name that explains why is it there, what are you going to use it and how, so everybody understand, at a glance what `buf` is about, without having to read further into the file, otherwise your code is not clear.

If you cannot find a good name, then please comment all these things.

alcortesm: For instance, let's say you are using buf to store the last scanned value, with the intention of having a unscan() method that allows to simulate 'inserting back the last scanned value into the stream'.

The reader of your code, when he reads buf, does not know about unscanning (something that is mentioned later) or how are you going to implement it (with the actual buffer).

So a better name for the `buf` thing will be something like this:

```
type Parser struct {
￼     s   *scanner
￼     lastResult struct { // memoization for unscanning
￼         tok token
￼         lit string
￼         valid bool
￼ }
```
This explains your intention and how are you going to use it from the very beginning, without having to read further into the file. It also helps to detect bugs, as now, everybody will see that valid is clearly a bool, not an int.

internal/revision/parser.go
alcortesm: of course, that is pretty clear by its name, what it is not clear is their purpose, how are you going to use them, etc.

By calling something rev, I can imaging it is a Revision, and by calling something revs I can imaging it is a collections of revisions. But why do they do? for instance, it looks like `revs` is swhat you are planning to return from this method, so maybe you should call it `ret` because line 171 is so far away that nobody will see it when reading the revs daclaration.

`rev` is just a temporal value where you store the current revisioner being returned, so there is not much to explain, but declaring it outside of the loop is very confusing (the same happens with err) and why is rev outside the loop but tok is declared inside every iteration? all those things are confusing, so you have to explain them or do them in a way that are easy to understand without explanations.

internal/revision/parser.go
alcortesm: You are right, a `parseTilde` function makes more sense if it parse "~3" than just "3".

But that does not justify the extra work and complexity and the surprise everybody get when they realize you are inserting back a character you just extracted, just to immediately extract it again.

The correct solution is not to call `parseTilde` at all, you already know you are in a "tilde" thing, because you already read a "~", so you just call `parseInteger` instead to read the remaining "3".

internal/revision/parser.go
alcortesm: No, the function is just nice as it is, I was just complimenting your work here.

internal/revision/scanner.go
alcortesm: The parser only use `newScanner()`, `scanner.scan()` and `scanner.unscan()` and the rest of the scanner functions (`scanner.read`, `scanner.unread`, `scanner.scanWord()`...) are helper functions used internally by scanner, but not by parser or any other type in the code.

This means you wrote `scanner` as to have an API composed of three functions:
- newScanner
- scan
- unscan
Which are the ones that are going to be used from the "outside" (in this case the parser). The rest of the functions are only going to be used from the inside and don't need so much error checks, for example.

But nobody knows that until the expend several minutes studying your code, which is wrong.

You can clarify those aspects by giving scanner an API, maybe even making it an internal package, so I don't have to spend time checking if a private method is doing extensive checking of its arguments, because clearly is just an internal thing.

internal/revision/scanner.go
alcortesm: Yes please, don't hide errors, otherwise, how are we going to know what went wrong when the error occurs?

internal/revision/scanner.go
alcortesm: yes, you are right. It is always difficult to make those judgement calls.

plumbing/revision.go
alcortesm: yes.

internal/revision/scanner.go
alcortesm: Check out bytes.Reader also, it implements `io.RuneScanner` and looks pretty nice.

internal/revision/parser.go
antham: I don't see the point here about revs, revs is declared at the top of the functions close to the return definition it's easy to understand it would be returned at the end, it's subjective but this function is not that big you can easily check the function in one look, I think it's quite understandable.

internal/revision/parser.go
antham: I started step by step and with the idea to reuse component but that has no sense so we can drop those things indeed. But you won't call parseTilde, parseInteger, because in this function you parse and return something related to tilde and a tilde could be follow by nothing.

internal/revision/parser.go
antham: Ah ok sorry I though you mean the opposite.

internal/revision/scanner.go
```diff
@@ -0,0 +1,124 @@
...
+	if err != nil {
+		return zeroRune
+	}
+	return ch
+}
+
+// unread places the previously read rune back on the reader.
+func (s *scanner) unread() { _ = s.r.UnreadRune() }
+
+// Scan extract tokens from an input
```
antham: I updated the comment.

About string value it's used at several places :

https://github.com/src-d/go-git/pull/139/files#diff-baaf5fc3681a156b2bf66e6ae6a30c78R273
https://github.com/src-d/go-git/pull/139/files#diff-baaf5fc3681a156b2bf66e6ae6a30c78R275
https://github.com/src-d/go-git/pull/139/files#diff-baaf5fc3681a156b2bf66e6ae6a30c78R309
https://github.com/src-d/go-git/pull/139/files#diff-baaf5fc3681a156b2bf66e6ae6a30c78R364
and so on

internal/revision/parser.go
antham: For others expressions it's ok because as you enter in a sub function you know what kind of object you will have at the end and you only have to analyze the rest of the expression, unscanning the character is unecessary, but here the current token is part of the whole expression you have to unscan it.

internal/revision/parser.go
alcortesm: OK.

alcortesm:
@antham I see you are fixing some of the issues, but I cannot tell if you have finished. Please let me know when you are finished and your code is ready for another review.

antham:
@alcortesm sorry I'm quite slow, I wanted to do this, i will ping you when everything is over

alcortesm:
@antham , you are doing an excellent job here and we are so grateful for it. Take your time, work at your own pace, there is not hurry.

Just let me know when it is ready for another review, so I don't have to check every morning to see if it is ready or not.

antham:
Ok thank you that's nice. 

I will but maybe you would be pinged if I answer to your comments spread over this PR as I go so don't be surprised to receive notifications. I will ping you when everything will be done.

alcortesm:
:+1:

internal/revision/scanner.go
antham: text.Scanner is too high level to do this job adn probably to tied to go langage

antham: I need more control than what text.Scanner offers to me

antham: To spend more time working on that in need a bette argument than : 

> The only functionality your are using from a bufio.Reader is the readRune and unreadRune so using a full bufio.Reader cannot be justified.

If it's performance issue I'm ok with that, understandability as well but if it's only because I use this object for a subset of its functionalities what's wrong with that ? 

But I agree with the fact to rewrite that to have something better as with points you mentioned elsewhere

repository.go
antham: I followed instruction from mcuadro :

> Place all the parser, tokenizer, etc in an internal package called revision at the root of the package. This package deals with strings, not with the type Revision

Something is not clear, can you describe more how you want to organize revision ?

I added this method to repository cause I don't see why you would have to give a repository to solve a revision to a third part method, they are related closely to repository, like references. It seems strange to me to call it somewhere else, or at least if you do so it's great to have something to resolve them directly from repository as a convenience.

antham:
@alcortesm can you come back to have a look please ? I re-reviewed all your comments to be sure I adressed them but it's possible I missed some, let me know. Sorry for this big delay :-(

alcortesm:
Thanks for the notice.

According to Github, your code has conflicts with the master branch and can not be merged. I will review your changes once you solve the conflicts.

mcuadros:
@antham can you rebases, sorry but we did a huge change (as you can see)

antham:
@alcortesm done

alcortesm:
I'm taking a look at it.

repository.go
alcortesm: I'm sorry, I don't understand @mcuadros comment either.

Maybe he has something in mind about how this is going to be used in the future.

Apparently he wants the `ResolveRevision` method from `Repository` to receive a plain string, not a `plumbing.Revision`. I'm not sure if he wants the method to return also strings instead of a `plumbing.Hash`.

Let's invoke him, so he can solve the problem: @mcuadros, please check this!. You should do what he says if you want your PR to be merged.

PS: independently of that, if a revision will represents revisons ranges in the future, this method should return a slice of strings or of hashes, just to take that into account future improvements.

internal/revision/scanner.go
```diff
@@ -0,0 +1,134 @@
...
+				}
+
+				return word, string(data), nil
+			}
+		}
+
+		return word, string(data), nil
+	}
+
+	if unicode.IsNumber(ch) {
```
antham: I was thinking to do a thing like that, but I wa thinking at the end this function won't move much so I changed my mind, I will do this.

repository.go
antham: So you want to keep me thing like they are but add the function he mentionned ?

antham:
@alcortesm I wanted to make changes you requested  (except my comment) but I have trouble with billy, I opened an issue on the repository.

antham:
Ok @alcortesm with the trick of  @mcuadros I was able to have everything working I rewrote scanner according to what you stated. For the second comment, I wait your answer.

alcortesm:
Thanks @antham , I'm working on your revision right now.

mcuadros:
Just merged, will be great having the range revision :+1:

antham:
It was long job but we finally succeeded to do it.

Yep now we can work on range revision, is it possible to open an issue to have some spec about what you want  at the end.

antham:
@mcuadros have you seen my last comment ?

mcuadros:
Sorry about the radio silence, nowadays I don't have many time to thing about the API, but you can raise a issue or a PR, with your suggest implementation, if you are still interested on contribute.

antham:
@mcuadros sure I'm still interested, I was thinking to open an issue in next few days. I'm facing currently an issue with range revision, so I will open an issue when I will have something interesting as starting point.

Issue #126, opened 2016-11-15
pulling on existing repository with no changes produces new packfile


smola:
Pulling on an existing repository with no changes, using the filesystem storage, writes a new packfile.

mcuadros:
Fixed https://github.com/src-d/go-git/pull/136

PR #140, opened 2016-11-25
storage: filesystem, clean close when the packfile is not used

mcuadros:
This PR try to fix https://github.com/src-d/go-git/issues/126

If not fixed this problem directly, is a step towards.

When the PackWriter was created and not used, the temp file was keep. Also an error was returned because a empty packfile was read

### DIFF
--- a/storage/filesystem/internal/dotgit/writers.go
+++ b/storage/filesystem/internal/dotgit/writers.go
@@ -12,6 +12,12 @@ import (
 	"gopkg.in/src-d/go-git.v4/utils/fs"
 )
 
+// PackWriter is a io.Writer that generates the packfile index simultaneously,
+// a packfile.Decoder is used with a file reader to read the file being written
+// this operation is synchronized with the write operations.
+// The packfile is written in a temp file, when Close is called this file
+// is renamed/moved (depends on the Filesystem implementation) to the final
+// location, if the PackWriter is not used, nothing is written
 type PackWriter struct {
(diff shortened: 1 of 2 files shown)


storage/filesystem/internal/dotgit/writers.go
```diff
@@ -72,34 +78,58 @@ func (w *PackWriter) buildIndex() {
 	w.result <- err
 }
 
-func (w *PackWriter) Write(p []byte) (n int, err error) {
+// waitBuildIndex waits until buildIndex function finish, this can terminate
```
smola: finish -> finishes

storage/filesystem/internal/dotgit/writers.go
```diff
@@ -72,34 +78,58 @@ func (w *PackWriter) buildIndex() {
...
+// with with a packfile.ErrEmptyPackfile, this means that nothing was written
```
smola: with with -> with

storage/filesystem/internal/dotgit/writers.go
```diff
@@ -72,34 +78,58 @@ func (w *PackWriter) buildIndex() {
...
+	}
+
+	return err
+}
+
+func (w *PackWriter) Write(p []byte) (int, error) {
 	return w.synced.Write(p)
 }
 
+// Close close all the file descriptors and save the final packfile, if nothing
```
smola: close -> closes

PR #143, opened 2016-11-26
protocol/packp: sideband muxer and demuxer

mcuadros:
This is the implementation of the sideband muxer and the demuxer.


plumbing/protocol/packp/sideband/demux.go
```diff
@@ -0,0 +1,137 @@
...
+		return nil, err
+	} else if size > d.max {
+		return nil, ErrMaxPackedExceeded
+	}
+
+	switch Channel(content[0]) {
+	case PackData:
+		return content[1:], err
+	case ProgressMessage:
+		_, err := d.Progress.(io.Writer).Write(content[1:])
```
smola: If Progress needs to be a Writer, shouldn't it be defined as io.ReadWriter instead of just io.Reader?

plumbing/protocol/packp/sideband/demux.go
```diff
@@ -0,0 +1,137 @@
...
+		return nil, err
+	case ErrorMessage:
+		return nil, fmt.Errorf("unexepcted error: %s", content[1:])
```
smola: unexepcted -> unexpected

plumbing/protocol/packp/sideband/demux_test.go
```diff
@@ -0,0 +1,121 @@
+package sideband
+
+import (
+	"bytes"
+	"io/ioutil"
+	"testing"
+
+	. "gopkg.in/check.v1"
+	"gopkg.in/src-d/go-git.v4/plumbing/format/pktline"
```
smola: import group order

ajnavarro: Bad import order?

plumbing/protocol/packp/sideband/demux.go
```diff
@@ -0,0 +1,137 @@
+package sideband
+
+import (
+	"errors"
+	"fmt"
+	"io"
+
```
ajnavarro: remove space

plumbing/protocol/packp/sideband/demux.go
```diff
@@ -0,0 +1,137 @@
...
+		s:        pktline.NewScanner(r),
+		Progress: bytes.NewBuffer(nil),
+	}
+}
+
+// Read reads up to len(p) bytes from the PackData channel into p, an error can
+// be return if an error happends when reading or if a message is sent in the
+// ErrorMessage channel.
+//
+// If a ProgressMessage is read, it's not copied into b, intead of this is
```
ajnavarro: Remove duplicated 'is'

plumbing/protocol/packp/sideband/demux.go
```diff
@@ -0,0 +1,137 @@
...
+}
+
+func (d *Demuxer) nextPackData() ([]byte, error) {
+	content := d.readPending()
+	if len(content) != 0 {
+		return content, nil
+	}
+
+	if !d.s.Scan() {
+		return nil, io.EOF
```
smola: if there is no more input, shouldn't the scanner error be checked instead of always returning io.EOF?

plumbing/protocol/packp/sideband/demux.go
```diff
@@ -0,0 +1,137 @@
...
+}
+
+// Read reads up to len(p) bytes from the PackData channel into p, an error can
+// be return if an error happends when reading or if a message is sent in the
+// ErrorMessage channel.
+//
+// If a ProgressMessage is read, it's not copied into b, intead of this is
+// is stored, can be read through the reader Progress, the n value returned is
+// zero, err is nil unless an error reading happends.
+func (d *Demuxer) Read(b []byte) (n int, err error) {
```
ajnavarro: Rewrite the second paragraph. An example:

> If a ProgressMessage is read, it won't be copied to b. Instead of this, it is stored and can be read through the reader Progress. If the n value returned is zero, err will be nil unless an error reading happens.

plumbing/protocol/packp/sideband/demux.go
mcuadros: Is preventing being written by external sources, also I can to declare the intent that should be used for reading

plumbing/protocol/packp/sideband/demux.go
mcuadros: And Error is always nil, when Scan returns `io.EOF` @alcortesm

plumbing/protocol/packp/sideband/demux_test.go
```diff
@@ -0,0 +1,121 @@
...
+type SidebandSuite struct{}
+
+var _ = Suite(&SidebandSuite{})
+
+func (s *SidebandSuite) TestDecode(c *C) {
+	expected := []byte("abcdefghijklmnopqrstuvwxyz")
+
+	buf := bytes.NewBuffer(nil)
+	e := pktline.NewEncoder(buf)
+	e.Encode(append(PackData.Bytes(), expected[0:8]...))
```
alcortesm: This is awesome.

I think you should write this line as an example in the comment of the `Channel.Byte()` method, so everybody understand, at a single glance, what is its intended use case.

You can also call the method `func (c Channel) WithPayload(b []byte) []byte`, so its usage will be a little more clear:

```
e.Encode(PackData.WithPayload(expected[0:8])
```

And make it return a multireader to save memory.

plumbing/protocol/packp/sideband/demux.go
smola: :+1:

plumbing/protocol/packp/sideband/demux.go
```diff
@@ -0,0 +1,137 @@
...
+	return &Demuxer{
+		t:        t,
+		r:        r,
+		max:      max,
+		s:        pktline.NewScanner(r),
+		Progress: bytes.NewBuffer(nil),
+	}
+}
+
+// Read reads up to len(p) bytes from the PackData channel into p, an error can
```
mcuadros: You have a more detailed explanation of Demux at the struct, I think is clear enough, this function (specially for the user), only reads Packdata into p

plumbing/protocol/packp/sideband/demux.go
alcortesm: No, `Error()` is not always nil and `Scan()` does not return `io.EOF`, it returns a boolean instead: from `pktline.Scanner.Scan()`, plumbing/format/pktline/scanner.go:44 : 
```
// Scan advances the Scanner to the next pkt-line, whose payload will
// then be available through the Bytes method.  Scanning stops at EOF
// or the first I/O error.  After Scan returns false, the Err method
// will return any error that occurred during scanning, except that if
// it was io.EOF, Err will return nil.
func (s *Scanner) Scan() bool {
```

We should definitely check the scanner error in this case and return `io.EOF` if the error is nil.

This is a similar behavior as `bufio.Scanner`, see https://golang.org/pkg/bufio/#example_Scanner_lines for example.

PR #144, opened 2016-11-27
protocol/packp: capabilities new Capability entity and List struct, test improvements

mcuadros:
- `Capabilities` struct moved to a package inside packp, and renamed to `List`
- New type `Capability` and all the consts with the valid capabilities
- Validation of unknown capabilities, capabilities with invalid values and other cases 
- Removed duplicate Decode function at packp


common_test.go
```diff
@@ -98,7 +98,7 @@ func (c *MockFetchPackSession) AdvertisedReferences() (
 
 	h := fixtures.ByURL(c.endpoint.String()).One().Head
 
-	cap := packp.NewCapabilities()
+	cap := capability.NewList()
 	cap.Decode("6ecf0ef2c2dffb796033e5a02219af86ec6584e5 HEADmulti_ack thin-pack side-band side-band-64k ofs-delta shallow no-progress include-tag multi_ack_detailed no-done symref=HEAD:refs/heads/master agent=git/2:2.4.8~dbussink-fix-enterprise-tokens-compilation-1167-gc7006cf")
```
ajnavarro: `Decode` method changes his input param type from `string` to `byte[]`.

plumbing/protocol/packp/capability/list.go
```diff
@@ -0,0 +1,158 @@
...
+			return err
+		}
+	}
+
+	return nil
+}
+
+// Get returns the values for a capability
+func (l *List) Get(capability Capability) []string {
+	return l.m[capability].Values
```
alcortesm: This panics when capability is not found in the map.

plumbing/protocol/packp/capability/list.go
```diff
@@ -0,0 +1,158 @@
...
+
+	if !multipleArgument[c] && len(l.m[c].Values) > 0 {
+		return ErrMultipleArguments
+	}
+
+	l.m[c].Values = append(l.m[c].Values, values...)
+	return nil
+}
+
+func (l *List) validate(c Capability, values []string) error {
```
alcortesm: I think this method is missing the point: why is this a method of the List type?

IMHO it should be a method of the `entry` type or an external function. And it must be public.

There must be a public `List.Validate` method thought, but it should not do what this method is doing, instead it should check things like having incompatible capabilities in the list, like sideband and sideband64, for example.

plumbing/protocol/packp/capability/list.go
```diff
@@ -0,0 +1,158 @@
...
+	ErrEmtpyArgument = errors.New("empty argument")
+	// ErrMultipleArguments multiple argument given to a capabilities that not
+	// support it
+	ErrMultipleArguments = errors.New("multiple arguments not allowed")
+)
+
+// List represents a list of capabilities
+type List struct {
+	m map[Capability]*entry
+	o []string
```
alcortesm: What does `o` means?

What is it purpose? I cannot see how is this useful. If it is here only for easier sorting in the `String` method, it should be removed.

plumbing/protocol/packp/capability/list.go
```diff
@@ -0,0 +1,158 @@
...
+	if !l.Supports(c) {
+		l.m[c] = &entry{Name: c}
+		l.o = append(l.o, c.String())
+	}
+
+	if len(values) == 0 {
+		return nil
+	}
+
+	if !multipleArgument[c] && len(l.m[c].Values) > 0 {
```
alcortesm: this is already checked at line 88.

plumbing/protocol/packp/capability/list.go
```diff
@@ -0,0 +1,158 @@
...
+	ErrMultipleArguments = errors.New("multiple arguments not allowed")
+)
+
+// List represents a list of capabilities
+type List struct {
+	m map[Capability]*entry
+	o []string
+}
+
+type entry struct {
```
alcortesm: This must be public, as it is what the clients would like to use to add new capabilities, instead of a `Capability` plus some values.

I would rename `entry` to `Capability` and rename `Capability` to `Type`.

plumbing/protocol/packp/capability/list.go
```diff
@@ -0,0 +1,158 @@
...
+	var o string
+	for _, key := range l.o {
+		cap := l.m[Capability(key)]
+		if len(cap.Values) == 0 {
+			o += key + " "
+			continue
+		}
+
+		for _, value := range cap.Values {
+			o += fmt.Sprintf("%s=%s ", key, value)
```
alcortesm: strings should be quoted here to prevent name collisions. Also on the key.

plumbing/protocol/packp/capability/list.go
```diff
@@ -0,0 +1,158 @@
...
+		}
+	}
+
+	return strings.Trim(o, " ")
```
alcortesm: use a proper prefix separator, initialized to the empty string on the first iterarion, instead.

plumbing/protocol/packp/capability/list.go
```diff
@@ -0,0 +1,158 @@
+package capability
```
alcortesm: The `List` type will need some more methods in the future:
 - Complementary set: to check the validity of client requests: requested caps not present in the announcement
 - Announced-requested validity: to check client/server compatibility. Example: PushCert without AllowReachableSHA1InWant

plumbing/protocol/packp/capability/list.go
mcuadros: Quoted? Why?

plumbing/protocol/packp/capability/list.go
mcuadros: This is a basic validation preventing  values where should not be. The other validation is requested dependant and should be on the request, not here.

plumbing/protocol/packp/capability/list.go
mcuadros: This is not in the scope of this object this should be checked in the request since are request dependant

plumbing/protocol/packp/capability/list.go
mcuadros: I am not agree, is what I removed. The values are vary rarely use only two. So this proposal is worse for the majority of the cases

plumbing/protocol/packp/capability/list.go
mcuadros: True

plumbing/protocol/packp/capability/list.go
mcuadros: I will use a slice and implode it

plumbing/protocol/packp/capability/list.go
mcuadros: is not the same, this check, if has previous values stored, the other function validate the values you are trying to add

PR #145, opened 2016-11-28
transport: add local transport

smola:
* transport: move common packp protocol out of ssh transport.
* fixtures: add fixture for empty repository.
* transport: add local transport
* transport/client: add local handler for file://


plumbing/transport/client/client.go
```diff
@@ -21,7 +23,7 @@ func InstallProtocol(scheme string, c transport.Client) {
 }
 
 // NewClient returns the appropriate client among of the set of known protocols:
-// HTTP, SSH. See `InstallProtocol` to add or modify protocols.
+// HTTP, SSH or local. See `InstallProtocol` to add or modify protocols.
```
mcuadros: local or file?

smola:
Updated PR with more docs about internal/common.

PR #150, opened 2016-11-30
file: return error on File.Contents() if reader fails

smola:


smola:
For go-git v3: https://github.com/src-d/go-git/pull/149

PR #154, opened 2016-12-04
Edit example in README.md for v4 

daehyeok:
example code in README looks still remained in v3. I changed with new API


mcuadros:
Yep, sorry. We didn't update the readme because v4 is still work in progress, as soon as we release something stable we will update the readme, thx

mcuadros:
A more completed README.md done at #181, thanks!

PR #156, opened 2016-12-04
remote: sideband support

mcuadros:
This is the support of `Remote` for sideband capable servers.

When `Remote. Fetch` is called If the request contained a `side-band-64k` or `side-band` a `sideband.Demux` is used inside of `Remote` to read the packfile data. 

If the human readable progress is needed a io.Reader/io.Writer should be provided to `Repository.Progress` field, this pointer will be pass to the remotes and be used when the packfile is read.  If the `Repository.Progress` is not set the `no-progress` capability is set if supported.

Also a couple of bugs are solved at transport/client: 
- Content-Length is correctly encoded using `strconv.Itoa` instead of just casting a integer to string. 
- Request encoding is done now using the encoders provide by the `protocol/packp package

### DIFF
--- a/plumbing/protocol/packp/sideband/demux.go
+++ b/plumbing/protocol/packp/sideband/demux.go
@@ -20,17 +20,17 @@ type Progress interface {
 // Demuxer demultiplex the progress reports and error info interleaved with the
 // packfile itself.
 //
-// A sideband has three diferent channels the main one, call PackData, contains
+// A sideband has three different channels the main one, call PackData, contains
 // the packfile data, the ErrorMessage channel, that contains server errors and
 // the last one ProgressMessage channel containing information about the ongoing
-// tast happening in the server (optinal, can be suppressed sending NoProgress
+// task happening in the server (optional, can be suppressed sending NoProgress
 // or Quiet capabilities to the server)
 //
 // In order to demultiplex the data stream, method `Read` should be called to
 // retrieve the PackData channel, the incoming data from the ProgressMessage is
 // written at `Progress` (if any), if any message is retrieved from the
 // ErrorMessage channel an error is returned and we can assume that the
-// conection has been closed.
+// connection has been closed.
 type Demuxer struct {
 	t Type
 	r io.Reader
@@ -59,12 +59,11 @@ func NewDemuxer(t Type, r io.Reader) *Demuxer {
 }
 
 // Read reads up to len(p) bytes from the PackData channel into p, an error can
-// be return if an error happends when reading or if a message is sent in the
+// be return if an error happens when reading or if a message is sent in the
 // ErrorMessage channel.
 //
-// If a ProgressMessage is read, it won't be copied to b. Instead of this,
-// Progress was set will be stored there. If the n value returned is zero, err
-// will be nil unless an error reading happens.
+// When a ProgressMessage is read, is not copy to b, instead of this is written
+// to the Progress
 func (d *Demuxer) Read(b []byte) (n int, err error) {
 	var read, req int
 
--- a/plumbing/protocol/packp/sideband/demux.go
+++ b/plumbing/protocol/packp/sideband/demux.go
@@ -1,7 +1,6 @@
 package sideband
 
 import (
-	"bytes"
 	"errors"
 	"fmt"
 	"io"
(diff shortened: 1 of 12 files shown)


plumbing/protocol/packp/sideband/demux.go
```diff
@@ -15,21 +14,22 @@ var ErrMaxPackedExceeded = errors.New("max. packed size exceeded")
...
 
 // Demuxer demultiplex the progress reports and error info interleaved with the
 // packfile itself.
 //
-// A sideband has three diferent channels the main one call PackData contains
+// A sideband has three diferent channels the main one, call PackData, contains
 // the packfile data, the ErrorMessage channel, that contains server errors and
 // the last one ProgressMessage channel containing information about the ongoing
 // tast happening in the server (optinal, can be suppressed sending NoProgress
 // or Quiet capabilities to the server)
```
alcortesm: English grammar and typos, consider this alternate version:
```
Sideband information comes through 3 different channels:
- PackData: the packfile data itself.
- ErrorMessage: server errors.
- ProgressMessage: progress information about the packfile download process. When the NoProgress and/or Quiet capabilities are in use, this channel remains silent.
```

plumbing/protocol/packp/sideband/demux.go
```diff
@@ -15,21 +14,22 @@ var ErrMaxPackedExceeded = errors.New("max. packed size exceeded")
...
 //
 // In order to demultiplex the data stream, method `Read` should be called to
 // retrieve the PackData channel, the incoming data from the ProgressMessage is
-// stored and can be read from `Progress` field, if any message is retrieved
-// from the ErrorMessage channel an error is returned and we can assume that the
+// written at `Progress` (if any), if any message is retrieved from the
+// ErrorMessage channel an error is returned and we can assume that the
 // conection has been closed.
 type Demuxer struct {
```
alcortesm: This is confusing as the Demuxer type does not Demultiplex the channels and it is not clear how are errors messages sent back to the user.  Also a connection is mentioned in the comment, and some assumptions about it being closed, but it is not clear how this can be relevant for anything.


I think a better comment would be something like this:

```
[Foo] is a reader to the packfile data in the PackData channel.  The progress information
is writen to the the Progress field while reading.  If any message is received through the
ErrorMessage channel, the `Read` method will return it as an error and it will assume the
connection has been closed.
```

Although I am still not sure about the connection thing.

I would rather use a real demuxer though or even simply a method that returns 3 readers (or even better, 3 scanners): one for each channel, treating them as equal, with the same usage semantics.  This will have a much simpler API and will let the user decide how to use each channel.

Another approach would be to use a single scanner, returning channel and data when read.

plumbing/protocol/packp/sideband/demux.go
```diff
@@ -51,21 +51,20 @@ func NewDemuxer(t Type, r io.Reader) *Demuxer {
...
-		Progress: bytes.NewBuffer(nil),
+		t:   t,
+		r:   r,
+		max: max,
+		s:   pktline.NewScanner(r),
 	}
 }
 
 // Read reads up to len(p) bytes from the PackData channel into p, an error can
 // be return if an error happends when reading or if a message is sent in the
```
alcortesm: s/sent/received/

alcortesm: s/in/over/

plumbing/protocol/packp/sideband/demux.go
```diff
@@ -51,21 +51,20 @@ func NewDemuxer(t Type, r io.Reader) *Demuxer {
...
 // ErrorMessage channel.
 //
-// If a ProgressMessage is read, it won't be copied to b. Instead of this, it is
-// stored and can be read through the reader Progress. If the n value returned
-// is zero, err will be nil unless an error reading happens.
+// If a ProgressMessage is read, it won't be copied to b. Instead of this,
+// Progress was set will be stored there. If the n value returned is zero, err
+// will be nil unless an error reading happens.
```
alcortesm: rewrite.

plumbing/protocol/packp/sideband/demux.go
```diff
@@ -126,13 +125,17 @@ func (d *Demuxer) nextPackData() ([]byte, error) {
 	case PackData:
 		return content[1:], nil
 	case ProgressMessage:
-		_, err := d.Progress.(io.Writer).Write(content[1:])
-		return nil, err
+		if d.Progress != nil {
```
alcortesm: I think `NewDemuxer` should has a `Progress` as an argument.

mcuadros: Is a optional argument, and not the most common behavior, so that's why is not part of the New method

remote_test.go
```diff
@@ -19,6 +19,8 @@ import (
 	"gopkg.in/src-d/go-git.v4/storage/memory"
 	osfs "gopkg.in/src-d/go-git.v4/utils/fs/os"
 
+	"bytes"
```
smola: import order

PR #158, opened 2016-12-05
plumbing/protocol: packp avoid duplication of haves, wants and shallow

mcuadros:
This avoid to have any duplicated hash in haves, wants or shallow lists in different messages.
Additionally `plumbing.HashesSort` has been added to easy hash list sorting

### DIFF
--- a/plumbing/protocol/packp/ulreq_encode.go
+++ b/plumbing/protocol/packp/ulreq_encode.go
@@ -1,9 +1,9 @@
 package packp
 
 import (
+	"bytes"
 	"fmt"
 	"io"
-	"sort"
 	"time"
 
 	"gopkg.in/src-d/go-git.v4/plumbing"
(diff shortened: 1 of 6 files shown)


plumbing/protocol/packp/ulreq_encode.go
```diff
@@ -78,23 +68,38 @@ func (e *ulReqEncoder) encodeFirstWant() stateFn {
...
+		last = w
 	}
 
 	return e.encodeShallows
 }
 
 func (e *ulReqEncoder) encodeShallows() stateFn {
-	sorted := sortHashes(e.data.Shallows)
-	for _, s := range sorted {
+	plumbing.HashesSort(e.data.Shallows)
```
alcortesm: This is not the correct way to solve this.

If we are going to create a new `plubing.HashesSort([]Hash)`, why not create a `plumbing.UniqueHashes([]Hash) []Hash` in addition, which is really what we want to do?

That would avoid duplicating code, improve code legibility and communicate intention better.

PR #162, opened 2016-12-06
protocol/packp: add report status message.

smola:


plumbing/protocol/packp/report_status.go
```diff
@@ -0,0 +1,145 @@
...
+			break
+		}
+
+		if err := s.decodeCommandStatus(b); err != nil {
+			return err
+		}
+	}
+
+	if !flushed {
+		return fmt.Errorf("missing flush")
```
mcuadros: how can you reach this point without `flushed` being true?  (with this code)

smola: when scan.Scan() = false in the first iteration.

PR #171, opened 2016-12-09
fixtures: fix empty repository fixture .git.

smola:
Previous fixture for empty.git was created with clone without --bare.
That was not valid for push tests. Replaced with a .git created with
git init --bare.


fixtures/fixtures.go
```diff
@@ -268,6 +268,7 @@ func (s *Suite) SetUpSuite(c *check.C) {
 
 func (s *Suite) TearDownSuite(c *check.C) {
 	for f := range folders {
+		continue
```
mcuadros: what?

alcortesm: test leftover ?

Issue #50, opened 2016-05-19
Implement client for git schema


mcuadros:
https://git-scm.com/book/ch4-1.html#The-Git-Protocol

mcuadros:
@ferhatelmas maybe you can take care of this as well, since you already know a bit about clients

ferhatelmas:
@mcuadros Sure, my pleasure.

smola:
Merged.

PR #175, opened 2016-12-09
transport: new git protocol

mcuadros:
This is the implementation of fetch-pack command over git protocol, the send-pack protocol is extremely undesirable, so I think doesn't work to implement it.

The communication is done by a plain TCP connection to the default port `9418` and using pkt-line as communication transport.

https://github.com/src-d/go-git/issues/50


alcortesm:
We should really stop naming files `common.go`.

PR #177, opened 2016-12-11
config: Config, bare flag

mcuadros:
This is the `core.bare` config flag, this PR implemente the encoding/decoding of the flag, and also the `Repository.Clone` makes use of it

### DIFF
--- a/storage/filesystem/config.go
+++ b/storage/filesystem/config.go
@@ -1,9 +1,8 @@
 package filesystem
 
 import (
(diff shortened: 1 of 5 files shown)


storage/filesystem/config.go
```diff
@@ -3,15 +3,19 @@ package filesystem
 import (
 	"os"
```
smola: extra space

PR #178, opened 2016-12-12
remote: add Push

smola:
* add Push method to Remote.
* add method Push to Repository.
* examples: add push example.

### DIFF
--- a/examples/common_test.go
+++ b/examples/common_test.go
@@ -2,6 +2,7 @@ package examples
 
 import (
 	"flag"
+	"fmt"
 	"go/build"
 	"io/ioutil"
 	"os"
@@ -20,6 +21,16 @@ var args = map[string][]string{
 	"clone":       []string{defaultURL, tempFolder()},
 	"progress":    []string{defaultURL, tempFolder()},
 	"open":        []string{filepath.Join(cloneRepository(defaultURL, tempFolder()), ".git")},
+	"push": []string{
+		func() string {
+			remote := createBareRepository(tempFolder())
+			local := filepath.Join(cloneRepository(defaultURL, tempFolder()), ".git")
+			setRemote(local, fmt.Sprintf("file://%s", remote))
+			return local
+		}(),
+		"origin",
+		"refs/heads/*:refs/heads/*",
+	},
 }
 
 var ignored = map[string]bool{
@@ -83,6 +94,21 @@ func cloneRepository(url, folder string) string {
 	CheckIfError(err)
 
 	return folder
+}
+
+func createBareRepository(dir string) string {
+	cmd := exec.Command("git", "init", "--bare", dir)
+	err := cmd.Run()
+	CheckIfError(err)
+
+	return dir
+}
+
+func setRemote(local, remote string) {
+	cmd := exec.Command("git", "remote", "set-url", "origin", remote)
+	cmd.Dir = local
+	err := cmd.Run()
+	CheckIfError(err)
 }
 
 func testExample(t *testing.T, name, example string) {
--- a/examples/common_test.go
+++ b/examples/common_test.go
@@ -21,16 +21,7 @@ var args = map[string][]string{
 	"clone":       []string{defaultURL, tempFolder()},
 	"progress":    []string{defaultURL, tempFolder()},
 	"open":        []string{filepath.Join(cloneRepository(defaultURL, tempFolder()), ".git")},
-	"push": []string{
-		func() string {
-			remote := createBareRepository(tempFolder())
-			local := filepath.Join(cloneRepository(defaultURL, tempFolder()), ".git")
-			setRemote(local, fmt.Sprintf("file://%s", remote))
-			return local
-		}(),
-		"origin",
-		"refs/heads/*:refs/heads/*",
-	},
+	"push":        []string{setEmptyRemote(filepath.Join(cloneRepository(defaultURL, tempFolder()), ".git"))},
 }
 
 var ignored = map[string]bool{
@@ -101,6 +92,12 @@ func createBareRepository(dir string) string {
 	err := cmd.Run()
 	CheckIfError(err)
 
+	return dir
+}
+
+func setEmptyRemote(dir string) string {
+	remote := createBareRepository(tempFolder())
+	setRemote(dir, fmt.Sprintf("file://%s", remote))
 	return dir
 }
 
--- a/examples/push/main.go
+++ b/examples/push/main.go
@@ -1,0 +1,22 @@
+package main
+
+import (
+	"fmt"
+	"os"
+
+	"gopkg.in/src-d/go-git.v4"
+	. "gopkg.in/src-d/go-git.v4/examples"
+)
+
+func main() {
+	CheckArgs("<repo path>")
+	repoPath := os.Args[1]
+
+	repo, err := git.NewFilesystemRepository(repoPath)
+	CheckIfError(err)
+
+	err = repo.Push(&git.PushOptions{})
+	CheckIfError(err)
+
+	fmt.Print("pushed")
+}
--- a/options.go
+++ b/options.go
@@ -100,3 +100,27 @@ func (o *FetchOptions) Validate() error {
 
 	return nil
 }
+
+// PushOptions describe how a push should be performed.
+type PushOptions struct {
+	// RemoteName is the name of the remote to be pushed to.
+	RemoteName string
+	// RefSpecs specify what destination ref to update with what source
+	// object. A refspec with empty src can be used to delete a reference.
+	RefSpecs []config.RefSpec
+}
+
+// Validate validate the fields and set the default values
+func (o *PushOptions) Validate() error {
+	if o.RemoteName == "" {
+		o.RemoteName = DefaultRemoteName
+	}
+
+	for _, r := range o.RefSpecs {
+		if !r.IsValid() {
+			return ErrInvalidRefSpec
+		}
+	}
+
+	return nil
+}
--- a/options.go
+++ b/options.go
@@ -116,6 +116,12 @@ func (o *PushOptions) Validate() error {
 		o.RemoteName = DefaultRemoteName
 	}
 
+	if len(o.RefSpecs) == 0 {
+		o.RefSpecs = []config.RefSpec{
+			config.RefSpec(config.DefaultPushRefSpec),
+		}
+	}
+
 	for _, r := range o.RefSpecs {
 		if !r.IsValid() {
 			return ErrInvalidRefSpec
--- a/remote.go
+++ b/remote.go
@@ -70,10 +70,6 @@ func (r *Remote) Push(o *PushOptions) (err error) {
 		return fmt.Errorf("remote names don't match: %s != %s", o.RemoteName, r.c.Name)
 	}
 
-	if len(o.RefSpecs) == 0 {
-		return fmt.Errorf("no refspecs given")
-	}
-
 	s, err := newSendPackSession(r.c.URL)
 	if err != nil {
 		return err
--- a/remote.go
+++ b/remote.go
@@ -8,9 +8,11 @@ import (
 	"gopkg.in/src-d/go-git.v4/config"
 	"gopkg.in/src-d/go-git.v4/plumbing"
 	"gopkg.in/src-d/go-git.v4/plumbing/format/packfile"
+	"gopkg.in/src-d/go-git.v4/plumbing/object"
 	"gopkg.in/src-d/go-git.v4/plumbing/protocol/packp"
 	"gopkg.in/src-d/go-git.v4/plumbing/protocol/packp/capability"
 	"gopkg.in/src-d/go-git.v4/plumbing/protocol/packp/sideband"
+	"gopkg.in/src-d/go-git.v4/plumbing/revlist"
 	"gopkg.in/src-d/go-git.v4/plumbing/storer"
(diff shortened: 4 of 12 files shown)


examples/push/main.go
```diff
@@ -0,0 +1,47 @@
...
+import (
+	"fmt"
+	"os"
+
+	"gopkg.in/src-d/go-git.v4"
+	"gopkg.in/src-d/go-git.v4/config"
+)
+
+func main() {
+	if len(os.Args) != 4 {
```
mcuadros: I created a helper function for the args parse:
```
CheckArgs("<directory>", "<remote-url>", "<refspec>")
directory := os.Args[1]
remoteUrl := os.Args[2]
```

examples/push/main.go
```diff
@@ -0,0 +1,47 @@
...
+	remoteUrl := os.Args[2]
+	refspec := config.RefSpec(os.Args[3])
+
+	if !refspec.IsValid() {
+		fmt.Printf("invalid refspec: %s\n", refspec)
+		os.Exit(1)
+	}
+
+	repo, err := git.NewFilesystemRepository(repoPath)
+	if err != nil {
```
mcuadros: The err are checked using `CheckIfError(err)`

examples/push/main.go
```diff
@@ -0,0 +1,47 @@
...
+	}
+
+	if _, err := repo.CreateRemote(&config.RemoteConfig{
+		Name: remoteName,
+		URL:  remoteUrl,
+	}); err != nil {
+		panic(err)
+	}
+
+	if err := repo.Push(&git.PushOptions{
```
mcuadros: And is supposed to illustrate the code with his homologue at git 
`Info("git clone %s %s", url, directory)`

remote.go
```diff
@@ -17,33 +17,45 @@ import (
...
+	NoErrAlreadyUpToDate = errors.New("already up-to-date")
+	ErrAlreadyConnected  = errors.New("remote already connected")
+	ErrFetchNotConnected = errors.New("fetch not connected")
+	ErrPushNotConnected  = errors.New("push not connected")
+)
 
 // Remote represents a connection to a remote repository
 type Remote struct {
 	c *config.RemoteConfig
+	r *Repository
```
mcuadros: if you add the Repository then remove the Storer

remote.go
```diff
@@ -17,33 +17,45 @@ import (
...
 
 // Config return the config
 func (r *Remote) Config() *config.RemoteConfig {
 	return r.c
 }
 
-// Connect with the endpoint
-func (r *Remote) Connect() error {
+// ConnectFetch connects to the remote and starts a fetch session.
+func (r *Remote) ConnectFetch() error {
```
mcuadros: make private this methods

remote.go
```diff
@@ -137,6 +173,99 @@ func (r *Remote) Fetch(o *FetchOptions) (err error) {
...
+
+	if o.RemoteName != r.c.Name {
+		return fmt.Errorf("remote names don't match: %s != %s", o.RemoteName, r.c.Name)
+	}
+
+	if len(o.RefSpecs) == 0 {
+		return fmt.Errorf("no refspec given")
+	}
+
+	if err := o.Validate(); err != nil {
```
mcuadros: the validation should be the first thing, doesn't worth to make any operation if the o are not valid

remote.go
```diff
@@ -137,6 +173,99 @@ func (r *Remote) Fetch(o *FetchOptions) (err error) {
...
+	}
+
+	req := packp.NewReferenceUpdateRequestFromCapabilities(r.advRefs.Capabilities)
+	for _, rs := range o.RefSpecs {
+		iter, err := r.s.IterReferences()
+		if err != nil {
+			return err
+		}
+
+		if err := iter.ForEach(func(ref *plumbing.Reference) error {
```
mcuadros: I prefer not having the if and the ForEach with the closure in the same line

remote.go
```diff
@@ -318,10 +511,19 @@ func (r *Remote) References() (storer.ReferenceIter, error) {
 	return r.refs.IterReferences()
 }
 
-// Disconnect from the remote and save the config
+// Disconnect from the remote server. Calling Disconnect on a disconnected
+// remote has no effect.
 func (r *Remote) Disconnect() error {
```
mcuadros: make it private

remote.go
```diff
@@ -294,6 +423,70 @@ func (r *Remote) buildFetchedTags() error {
 	})
 }
 
+func (r *Remote) getHaves(refs memory.ReferenceStorage) []plumbing.Hash {
```
alcortesm: why is this a method of `Remote` instead of a free function?

repository.go
```diff
@@ -275,6 +269,25 @@ func (r *Repository) Pull(o *PullOptions) error {
 	return r.createReferences(head)
 }
 
+// Push pushes changes to a remote.
+func (r *Repository) Push(o *PushOptions) error {
+	if o.RemoteName == "" {
```
alcortesm: Again, this was supposed to be part of Validate.

repository.go
```diff
@@ -275,6 +269,25 @@ func (r *Repository) Pull(o *PullOptions) error {
...
+		o.RemoteName = DefaultRemoteName
+	}
+
+	remote, err := r.Remote(o.RemoteName)
+	if err != nil {
+		return err
+	}
+
+	if err := remote.ConnectPush(); err != nil {
```
alcortesm: I think Connect should not be needed here, unless you want to do other things over the connection in addition to Push.

options.go
```diff
@@ -94,3 +94,27 @@ func (o *FetchOptions) Validate() error {
...
+// PushOptions describe how a push should be performed.
+type PushOptions struct {
+	// RemoteName is the name of the remote to be pushed to.
+	RemoteName string
+	// RefSpecs specify what destination ref to update with what source
+	// object. A refspec with empty src can be used to delete a reference.
+	RefSpecs []config.RefSpec
+}
+
+// Validate validate the fields and set the default values
```
alcortesm: Based on its name, I find weird that Validate set anything, I would expect it to have a non-pointer receiver and just return true or false.

smola: I'm following the same convention that we are using for a lot of objects in go-git (check other Validate methods). We can discuss changing them as a separate issue, not related to this PR.

options.go
```diff
@@ -94,3 +94,27 @@ func (o *FetchOptions) Validate() error {
...
+func (o *PushOptions) Validate() error {
```
smola: See the above comment, I'm using the same conventions for Validate as in the rest of the code base. The example doesn't use Validate, the user doesn't need to call it, since it's always called by porcelain methods when they get options.

examples/push/main.go
```diff
@@ -0,0 +1,47 @@
...
+		fmt.Printf("invalid refspec: %s\n", refspec)
+		os.Exit(1)
+	}
+
+	repo, err := git.NewFilesystemRepository(repoPath)
+	if err != nil {
+		panic(err)
+	}
+
+	if _, err := repo.CreateRemote(&config.RemoteConfig{
```
smola: Yeah, I'm changing the example to do a more standard/intuitive thing.

remote.go
```diff
@@ -17,33 +17,45 @@ import (
...
+	r *Repository
 	s Storer
 	p sideband.Progress
 
 	// cache fields, there during the connection is open
 	endpoint     transport.Endpoint
 	client       transport.Client
 	fetchSession transport.FetchPackSession
+	sendSession  transport.SendPackSession
+	session      transport.Session
```
smola: session can be either fetchSession / sendSession. It's used for some common operations.

remote.go
```diff
@@ -137,6 +173,99 @@ func (r *Remote) Fetch(o *FetchOptions) (err error) {
...
 }
 
+// Push performs a push to the remote. Returns NoErrAlreadyUpToDate if the
+// remote was already up-to-date.
+//
+// TODO: Support deletes.
+// TODO: Support pushing tags.
+// TODO: Check if force update is given, otherwise reject non-fast forward.
+func (r *Remote) Push(o *PushOptions) (err error) {
+	if o.RemoteName == "" {
```
smola: Right. Changing it.

remote.go
```diff
@@ -137,6 +173,99 @@ func (r *Remote) Fetch(o *FetchOptions) (err error) {
...
+
+	if len(o.RefSpecs) == 0 {
+		return fmt.Errorf("no refspec given")
+	}
+
+	if err := o.Validate(); err != nil {
+		return err
+	}
+
+	if r.sendSession == nil {
```
smola: We should avoid it entirely, Push and Fetch should be atomic operations without public connect/disconnect methods. However, changing this requires significant changes to how Fetch is used by Repository.

remote.go
smola: Changed.

remote.go
smola: Done.

remote.go
smola: Not until we make Fetch atomic, which currently it is not.

repository.go
smola: I did it this way to keep symmetry with the Fetch implementation. Although both should be changed to have atomic operations without Connect/Disconnect at all.

remote.go
smola: This will go away if we merge https://github.com/src-d/go-git/pull/185 first.

remote.go
smola: it will not be present in the next version (after rebase).

remote.go
smola: they are gone after the next rebase

examples/push/main.go
```diff
@@ -0,0 +1,33 @@
+package main
+
+import (
+	"fmt"
+	"os"
+
+	"gopkg.in/src-d/go-git.v4"
+	"gopkg.in/src-d/go-git.v4/config"
+	. "gopkg.in/src-d/go-git.v4/examples"
```
smola: why? they are all internal imports.

options.go
```diff
@@ -94,3 +94,27 @@ func (o *FetchOptions) Validate() error {
 
 	return nil
 }
+
+// PushOptions describe how a push should be performed.
+type PushOptions struct {
+	// RemoteName is the name of the remote to be pushed to.
+	RemoteName string
```
alcortesm: :+1:

options.go
alcortesm: :+1:

options.go
alcortesm: :+1:

remote.go
alcortesm: :+1:

remote.go
alcortesm: :+1:

repository.go
alcortesm: :+1:

examples/push/main.go
ajnavarro: oops, you're right, I just saw the period and i thought that was go-check or something... :+1:

mcuadros:
@smola rebase

examples/common_test.go
```diff
@@ -20,6 +25,16 @@ var args = map[string][]string{
 	"clone":       []string{defaultURL, tempFolder()},
 	"progress":    []string{defaultURL, tempFolder()},
 	"open":        []string{filepath.Join(cloneRepository(defaultURL, tempFolder()), ".git")},
+	"push": []string{
```
mcuadros: can you create a function?

examples/common_test.go
```diff
@@ -20,6 +25,16 @@ var args = map[string][]string{
...
+		func() string {
+			remote := createBareRepository(tempFolder())
+			local := filepath.Join(cloneRepository(defaultURL, tempFolder()), ".git")
+			setRemote(local, fmt.Sprintf("file://%s", remote))
+			return local
+		}(),
+		"origin",
```
mcuadros: the default RemoteName is "origin" should to omit this arg to simplify the example?

options.go
```diff
@@ -100,3 +100,27 @@ func (o *FetchOptions) Validate() error {
...
+	// RefSpecs specify what destination ref to update with what source
+	// object. A refspec with empty src can be used to delete a reference.
+	RefSpecs []config.RefSpec
```
mcuadros: We can have a default RefSpec?

remote.go
```diff
@@ -221,6 +376,27 @@ func getWants(spec []config.RefSpec, localStorer Storer, remoteRefs storer.Refer
 	return result, nil
 }
 
+func (r *Remote) referenceNeedsUpdate(wantTags bool, specs []config.RefSpec,
```
mcuadros: function not tested, or codecov failing

repository.go
```diff
@@ -345,6 +345,20 @@ func (r *Repository) Fetch(o *FetchOptions) error {
 	return remote.Fetch(o)
 }
 
+// Push pushes changes to a remote.
+func (r *Repository) Push(o *PushOptions) error {
```
mcuadros: function not tested

examples/common_test.go
smola: done

examples/common_test.go
smola: done, reduced to just push origin with default refspec.

options.go
smola: yes: refs/heads/*:refs/heads/*. I've just added it. Note that git has different behaviours for default push refspec depending on config.

repository.go
smola: done.

PR #180, opened 2016-12-13
storage: shallow storage

mcuadros:
This PR implements the shallow storage, when a Repository is cloned using Depth != 0, the shallow commits should be saved

### DIFF
--- a/plumbing/storer/shallow.go
+++ b/plumbing/storer/shallow.go
@@ -2,8 +2,8 @@ package storer
 
 import "gopkg.in/src-d/go-git.v4/plumbing"
 
-// ShallowStorer storage of shallow commits, meaning that it does not have the
-// parents of a commit (explanation from git documentation)
+// ShallowStorer storage of references to shallow commits by hash, meaning that
+// these commits have missing parents because of a shallow fetch.
 type ShallowStorer interface {
 	SetShallow([]plumbing.Hash) error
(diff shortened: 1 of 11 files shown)


plumbing/storer/shallow.go
```diff
@@ -0,0 +1,10 @@
+package storer
+
+import "gopkg.in/src-d/go-git.v4/plumbing"
+
+// ShallowStorer storage of shallow commits, meaning that it does not have the
+// parents of a commit (explanation from git documentation)
+type ShallowStorer interface {
```
alcortesm: Why is there a `storer.ShallowStorer` interface when we already have an `storer.ObjectStorer`?

alcortesm: Why the interfaces repeat the word `Storer` in their name when it is already part of the package name?

storage/filesystem/shallow.go
```diff
@@ -0,0 +1,52 @@
+package filesystem
+
+import (
+	"fmt"
+
```
smola: extra space ;)

plumbing/storer/shallow.go
smola: I'd propose something along these lines:
```
// ShallowStorer storage of references to shallow commits by hash, meaning that these commits have
// missing parents because of a shallow fetch.
```
?

PR #190, opened 2016-12-19
server: add git server implementation

smola:


smola:
Depends on https://github.com/src-d/go-git/pull/178

examples/git-receive-pack/main.go
```diff
@@ -0,0 +1,23 @@
...
+	"gopkg.in/src-d/go-git.v4/plumbing/transport/file"
+)
+
+func main() {
+	if len(os.Args) != 2 {
+		fmt.Fprintln(os.Stderr, "usage: git receive-pack <git-dir>")
+		os.Exit(129)
+	}
+
+	if err := file.DefaultServer.Serve(
```
mcuadros: Use CheckIfError from the example library

plumbing/protocol/packp/capability/list.go
```diff
@@ -167,6 +167,18 @@ func (l *List) Delete(capability Capability) {
 	}
 }
 
+// Check checks that all capabilities set in the given list are supported
+// in this list.
+func (l *List) Check(ol *List) error {
```
mcuadros: Maybe instead of doing this you can turn `Supports` in variadic

plumbing/protocol/packp/uppackresp.go
```diff
@@ -17,8 +18,8 @@ var ErrUploadPackResponseNotDecoded = errors.New("upload-pack-response should be
 type UploadPackResponse struct {
 	ShallowUpdate
 	ServerResponse
-	r io.ReadCloser
 
+	Packfile   io.ReadCloser
```
mcuadros: why?

plumbing/transport/file/git-receive-pack
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+# Simulates the git-upload-pack command.
+# For testing purposes only.
+# See examples/git-upload-pack.
+
+path="$(dirname $0)"
+exec go run "$path/../../../examples/git-receive-pack/main.go" "$@"
```
mcuadros: what is this?

mcuadros: if is for testing write it with the test, or something, or put in a fixture folder

plumbing/transport/file/server.go
```diff
@@ -0,0 +1,82 @@
...
+	"gopkg.in/src-d/go-git.v4/plumbing/transport/server"
+	"gopkg.in/src-d/go-git.v4/storage/filesystem"
+	"gopkg.in/src-d/go-git.v4/utils/ioutil"
+
+	osfs "srcd.works/go-billy.v1/os"
+)
+
+var DefaultServer = NewServer(server.DefaultHandler)
+
+type Server struct {
```
mcuadros: having the server at this level don't allow to you to having a server of multiple repositories, I was thinking on having a server package at porcelain level

plumbing/transport/file/upload_pack_test.go
```diff
@@ -0,0 +1,98 @@
...
+
+	. "gopkg.in/check.v1"
+)
+
+type UploadPackSuite struct {
+	fixtures.Suite
+	Server        *Server
+	Path          string
+	URL           string
+	UploadPackBin string
```
mcuadros: Just bin, since the test is for uploadpack

plumbing/transport/server/server.go
```diff
@@ -0,0 +1,394 @@
...
+
+func (s *rpSession) setStatus(ref plumbing.ReferenceName, err error) {
+	s.cmdStatus[ref] = err
+	if s.firstErr != nil && err != nil {
+		s.firstErr = err
+	}
+}
+
+func (s *rpSession) reportStatus() *packp.ReportStatus {
+
```
mcuadros: extra space

plumbing/transport/server/server.go
```diff
@@ -0,0 +1,394 @@
...
+func (*rpSession) setSupportedCapabilities(c *capability.List) error {
+	if err := c.Set(capability.Agent, capability.DefaultAgent); err != nil {
+		return err
+	}
+
+	if err := c.Set(capability.OFSDelta); err != nil {
+		return err
+	}
+
+	if err := c.Set(capability.ReportStatus); err != nil {
```
mcuadros: this can be returned directly

plumbing/transport/server/server.go
```diff
@@ -0,0 +1,394 @@
...
+		return err
+	}
+
+	if ref.Type() == plumbing.SymbolicReference {
+		if err := ar.AddReference(ref); err != nil {
+			return nil
+		}
+
+		ref, err = storer.ResolveReference(s, ref.Target())
+		if err == plumbing.ErrReferenceNotFound {
```
mcuadros: duplicate code, function?

examples/git-receive-pack/main.go
smola: I have not used it here so that the example behaves exactly the same as the official git-receive-pack.

plumbing/protocol/packp/uppackresp.go
smola: So that it can be set by the server when building the response.

plumbing/transport/file/git-receive-pack
smola: ok, writing it in SetUpSuite

plumbing/transport/file/server.go
smola: You can call Serve multiple times with different arguments for different repositories.
I am thinking of modularizing it further by decoupling repository finding/loading from the server itself, just as with the server.Handler. Something like server.StorerLoader. The default one would load repositories by file system path and could be shared by default implementations, but the user could pass a loader getting repositories from a database. What do you think?

plumbing/transport/file/upload_pack_test.go
smola: ok

plumbing/transport/server/server.go
smola: ok

plumbing/transport/server/server.go
smola: ok

plumbing/transport/server/server.go
smola: ok

examples/git-receive-pack/main.go
mcuadros: but is a example so we don't need to replicate nothing

plumbing/transport/server/loader.go
```diff
@@ -0,0 +1,40 @@
...
+
+// DefaultLoader is a filesystem loader ignoring host and resolving paths to /.
+var DefaultLoader = NewFilesystemLoader(os.New("/"))
+
+// Loader loads repository's storer.Storer based on an optional host and a path.
+type Loader interface {
+	// Load loads a storer.Storer given an optional host and a path.
+	// Returns transport.ErrRepositoryNotFound if the repository does not
+	// exist.
+	Load(host, path string) (storer.Storer, error)
```
mcuadros: why not a Endpoint entity?

cli/go-git/main.go
```diff
@@ -0,0 +1,63 @@
...
+
+		os.Exit(1)
+	}
+}
+
+type cmd struct {
+	Verbose bool `short:"v" description:"Activates the verbose mode"`
+}
+
+func (c *cmd) print(format string, a ...interface{}) {
```
smola: Yes. Removed.

cli/go-git/main.go
```diff
@@ -0,0 +1,63 @@
...
+	if !c.Verbose {
+		return
+	}
+
+	fmt.Printf(format, a...)
+}
+
+func resolvePath(path string) (string, error) {
```
smola: Done.

plumbing/transport/file/server_test.go
```diff
@@ -0,0 +1,109 @@
...
+	c.Assert(cmd.Start(), IsNil)
+
+	go func() {
+		b, err := ioutil.ReadAll(sout)
+		if err != nil {
+			outDoneChan <- err
+		}
+
+		outChan <- string(b)
+		outDoneChan <- nil
```
smola: I've moved this code to a function, which should be more legible. Also changed the error channel to always return a value (either error or nil), just as if it was a regular error return value.

plumbing/transport/server/loader.go
```diff
@@ -0,0 +1,51 @@
...
+func (l *fsLoader) Load(ep transport.Endpoint) (storer.Storer, error) {
+	fs := l.base.Dir(ep.Path)
+	if _, err := fs.Stat("config"); err != nil {
+		return nil, transport.ErrRepositoryNotFound
+	}
+
+	return filesystem.NewStorage(fs)
+}
+
+type MapLoader map[transport.Endpoint]storer.Storer
```
smola: I've added further comments to explain each behaviour.

A Loader loads a repository (actually, a storer.Storer, which is a low-level representation of a repository) given an endpoint.

The fsLoader, as documented in NewFilesystemLoader, loads repositories from a file system. That is, it gets an endpoint's path, looks it up in the base file system, and if it's a repository, creates a storer for it. The DefaultLoader is a file system loader using the native OS filesystem.

MapLoader uses a lookup map to return storers that were already created by the user. This is useful for testing (e.g. registering a memory repository with a given path), but also for actual production applications, such as in-process pushes (i.e. source and destination repositories are both loaded in the same go-git process).

The relation in MapLoader is not bijective.

With respect to the name, feel free to suggest any alternative. `Load` takes an endpoint and it tries to build a storer for it if the repository exists.

cli/go-git/main.go
alcortesm: :+1:

cli/go-git/main.go
alcortesm: :+1:

plumbing/transport/file/server_test.go
alcortesm: :+1:

plumbing/transport/server/loader.go
alcortesm: I don't think I can provide a better name for `Load` right now.

I don't know what problems are you trying to solve with this `loader` thing. The description of the PR is not enough to understand why a `Loader` is needed and how we were doing that before having a `loader`. I think I am too disconnected from the project to give my opinion.

From my point of view: I don't understand some of the names here, they are too generic:

- 'storer.Storer': of what?, you say it is actually a repository in the previous answer but the name does not give any hints about it. By its definition it is anything capable of holding encoded objects and references (which looks as a classical git repository on disk to me, this is a packfile + references).

- 'server.Loader': of what?, it does not actually loads a server as its name implies, but apparently it is a database capable of returning the encoded objects and references of a repository from its endpoint. There is also a comment still says it looks at an (optional) host and path, that may be outdated. It looks as a resolver to me: given the name of a resource, returns the resource.

I don't think any of these is wrong, it is just that I am very lost and I have not found a description of the problem and how we have decided to solve it, so it is very difficult to understand your solution and think about it.

alcortesm: Maybe we can talk today about it.

alcortesm: Please, feel free to merge it none of these make sense to you. We can talk about it after merging.

smola: As discussed before, we'll proceed with current naming. We agreed that it's correct that `Loader` is in the `server` package, since there are no uses out of it. We can revisit this if a use case comes up for it.

With respect to naming of both `Loader` and `Storer`, the discussion is still open, but we'll not stop the PR.

PR #201, opened 2017-01-05
packfile/scanner: reset zlib reader instead of new one

ajnavarro:


plumbing/format/packfile/scanner.go
```diff
@@ -275,19 +276,28 @@ func (s *Scanner) NextObject(w io.Writer) (written int64, crc32 uint32, err erro
 // ReadRegularObject reads and write a non-deltified object
 // from it zlib stream in an object entry in the packfile.
 func (s *Scanner) copyObject(w io.Writer) (int64, error) {
```
alcortesm: The comment on this function starts with a wrong function name.

plumbing/format/packfile/scanner.go
```diff
@@ -275,19 +276,28 @@ func (s *Scanner) NextObject(w io.Writer) (written int64, crc32 uint32, err erro
...
-	zr, err := zlib.NewReader(s.r)
-	if err != nil {
-		return -1, fmt.Errorf("zlib reading error: %s", err)
+	var err error
+	if s.zr == nil {
+		zr, err := zlib.NewReader(s.r)
+		if err != nil {
+			return -1, fmt.Errorf("zlib initialization error: %s", err)
```
alcortesm: Why are we returning -1, this is not C!

`Scanner.NextObject` first return argument is not used anywhere, it is safe to return 0 here instead.

plumbing/format/packfile/scanner.go
ajnavarro: I think this change is not related to this PR, Could fix this naming issue into another PR?

plumbing/format/packfile/scanner.go
ajnavarro: I was just replying the previous behavior. I can change it.

plumbing/format/packfile/scanner.go
smola: Yes, please. Avoid unrelated changes here.

alcortesm: Sure, change it on another PR.

Issue #202, opened 2017-01-05
Fatal panic : Slice out of bounds


icambridge:
I get this fatal panic. When getting the commits for https://github.com/vkbot/TrinityCore-1

```panic: runtime error: slice bounds out of range

goroutine 3093 [running]:
panic(0x773b60, 0xc42000c0f0)
        /usr/local/go/src/runtime/panic.go:500 +0x1a1
gopkg.in/src-d/go-git.v4/plumbing/object.(*Signature).Decode(0xc42f9ecc58, 0xc42b74f065, 0x32, 0xf9b)
        /home/icambridge/golang/src/gopkg.in/src-d/go-git.v4/plumbing/object/object.go:93 +0x225
gopkg.in/src-d/go-git.v4/plumbing/object.(*Commit).Decode(0xc42f9ecc40, 0x98a900, 0xc4309fcb00, 0x0, 0x0)
        /home/icambridge/golang/src/gopkg.in/src-d/go-git.v4/plumbing/object/commit.go:147 +0x5c0
gopkg.in/src-d/go-git.v4/plumbing/object.DecodeCommit(0x6fe9c9844080, 0xc421f78330, 0x98a900, 0xc4309fcb00, 0x0, 0x0, 0xc42f4f41e0)
        /home/icambridge/golang/src/gopkg.in/src-d/go-git.v4/plumbing/object/commit.go:49 +0x9f
gopkg.in/src-d/go-git.v4/plumbing/object.(*CommitIter).Next(0xc430cdc060, 0xbc1909c6334e4c62, 0xc7d43224, 0xc42d6fb0a8)
        /home/icambridge/golang/src/gopkg.in/src-d/go-git.v4/plumbing/object/commit.go:250 +0x75
github.com/icambridge/techfinder/techfinder.GitRepository.Commits(0xc421f78390, 0xc421f78270, 0x2a, 0xc421f78210, 0x24, 0xc421f78240, 0x26, 0xc4225d1da0, 0x0, 0x0, ...)
        /home/icambridge/golang/src/github.com/icambridge/x.go:64 +0x138
```

mcuadros:
fixed

PR #204, opened 2017-01-09
plumbing: fix signature with `>` before `<` parsing

mcuadros:
This is a fix for issue: https://github.com/src-d/go-git/issues/202

Since the fullname is free text, given a `foo>` the result was a panic, now this is handled properly returning  `foo>` as the fullname. Another extra security guards was added to protect the function against malformed signatures.

### DIFF
--- a/plumbing/object/object.go
+++ b/plumbing/object/object.go
@@ -83,9 +83,13 @@ type Signature struct {
 
 // Decode decodes a byte slice into a signature
 func (s *Signature) Decode(b []byte) {
-	open := bytes.IndexByte(b, '<')
-	close := bytes.IndexByte(b, '>')
+	open := bytes.LastIndexByte(b, '<')
+	close := bytes.LastIndexByte(b, '>')
 	if open == -1 || close == -1 {
+		return
+	}
+
+	if close < open {
 		return
 	}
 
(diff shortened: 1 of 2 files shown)


ajnavarro:
If the email address is crap, this crap will be extended to the committer name:

```
`Foo Bar <foo>@bar.com>`: {
	Name:  "Foo Bar",
	Email: "foo>@bar.com",
	When:  time.Time{},
},
// ... obtained string = "Foo Bar <foo"
// ... expected string = "Foo Bar"
`Foo Bar <foo<@bar.com>`: {
	Name:  "Foo Bar",
	Email: "foo<@bar.com",
	When:  time.Time{},
},
// ... obtained string = "Foo Bar <foo"
// ... expected string = "Foo Bar"
```

smola:
We've checked the behavior of git and it's as follows:

* When generating commits, `<` and `>` symbols are removed (no matter their position). Both for name and email.
* When reading, different operations have slightly different behaviours For example, `git log` prints almost the original output (no normalization), but `git log --pretty='%an|%ae'`, which actually fully parses the signature has the following behaviour:
  * `foo <my<@email.com>` is parsed as name `foo` and email `my<@email.com`.
  * `foo <my>@email.com>` is parsed as name `foo` and email `my`.

Here's the original source in git: https://github.com/git/git/blob/3ef7618e616e023cf04180e30d77c9fa5310f964/ident.c#L250
we might want to replicate the same or similar logic for name splitting.

Issue #206, opened 2017-01-10
SSH connection without SSH-agent


skyflyer:
I'm having an issue connecting to git repository over ssh on windows, where I don't have an SSH agent running.

In `ssh/common.go`, the `command.connect` method calls `c.setAuthFromEndpoint()` which in turn tries to connect with SSH agent and fails. The method's comments note:

```
connect connects to the SSH server, unless a AuthMethod was set with
SetAuth method...
```

but for me, that does not work. Here is a brief program that reproduces this issue:

```go
func main() {
	repoURL := "ssh://git@hostname:portnum/testing.git"
	signer := getSSHSigner()
	r := git.NewMemoryRepository()
	err := r.Clone(&git.CloneOptions{
		URL:  repoURL,
		Auth: &gitssh.PublicKeys{User: "git", Signer: signer},
	})
	check(err)
}
```

The error returned is `dial unix: missing address`, which is correct (in a way), since there is no SSH agent environment variable available, so there's nowhere to connect to. The `getSigner` function above retrieves the ssh.Signer from a private key, supplied to the program.

I can prepare a pull request with a fix, if you'd like that; it seems to me, that we would need to reorganise some code, as currently, `Auth` does not get propagated down to the actual transport layer.

mikkeloscar:
I have the same problem as @skyflyer (on linux where I don't want to use the ssh-agent but pass in a privatekey).

mcuadros:
Exactly the `CloneOptions.Auth` field is totally unused, the only viable method to provide an auth config right now is using the endpoint and this is not enough or valid. We will find a solution in the next days.

amlweems:
I've been trying to find a way to implement this and it seems tricky without changing the interfaces significantly.

The general flow for `transport/http` interface is: create the session using `NewFetchPackSession` (which basically initializes a struct with an http.Client), optionally call `session.SetAuth(auth)` if needed, and call `FetchPack()` to execute the actual network connections.

The above flow works great because the initialization doesn't perform any network activity and it follows the `Session` interface in `transport/common.go`. The general flow for the `transport/ssh` interface is: create the session and Dial the SSH server using SSH agent for authentication and call `FetchPack()` to read from stdin/stdout.

I think the SSH transport needs to be broken out so that network connectivity happens during `FetchPack()` instead of in the session init, however, this is a rather large modification to the architecture.

[...]

PR #210, opened 2017-01-14
update transport Command interface to support SetAuth after creation

amlweems:
Reworks `Command` interface to split the session creation into two components: general object create & session connect. This allows an external caller to create the session, call `SetAuth()`, and **then** perform the network connection. This network connection is "baked" in to the other interface functions to keep the external interface the same.

Addresses #206.

### DIFF
--- a/plumbing/transport/internal/common/common.go
+++ b/plumbing/transport/internal/common/common.go
@@ -37,17 +37,13 @@ type Commander interface {
 	// error should be returned if the endpoint is not supported or the
 	// command cannot be created (e.g. binary does not exist, connection
 	// cannot be established).
-	Command(cmd string, ep transport.Endpoint) (Command, error)
+	Command(cmd string, ep transport.Endpoint, auth transport.AuthMethod) (Command, error)
 }
 
 // Command is used for a single command execution.
 // This interface is modeled after exec.Cmd and ssh.Session in the standard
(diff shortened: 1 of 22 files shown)


plumbing/transport/internal/common/common.go
```diff
@@ -46,6 +46,8 @@ type Commander interface {
 type Command interface {
 	// SetAuth sets the authentication method.
 	SetAuth(transport.AuthMethod) error
+	// Perform the actual network connection.
+	Connect() error
```
smola: SetAuth is actually a leftover from a previous implementation that should have been gone. Auth can only be set before connection and we don't support reusing commands or connections at all.

So, instead of adding a new Connect function, I think it would be better to just remove SetAuth from all interfaces (Command, Session) and add transport.AuthMethod as a second parameter to NewUploadPackSession and NewReceivePackSession.

That would keep these interfaces simpler and avoid invalid sequences of function calls.

amlweems: Ah, I had wondered about that. I've moved `AuthMethod` to the `PackSession` functions now.

While doing that, I noticed that server also uses these functions. I've left a `TODO` for now to design / implement `AuthMethod` for servers. I *think* this could be used to verify a client connection to the go-git server in the future.

PR #214, opened 2017-01-18
fixtures: initialize fixtures into separated methods

ajnavarro:
To be able to use fixtures with other test frameworks than go-check,
we created two methods, one to set fixtures path correctly, and
another to remove all the temporal data created when testing.

### DIFF
--- a/fixtures/fixtures.go
+++ b/fixtures/fixtures.go
@@ -264,20 +264,34 @@ func (g Fixtures) Exclude(tag string) Fixtures {
 	return r
 }
 
-type Suite struct{}
-
-func (s *Suite) SetUpSuite(c *check.C) {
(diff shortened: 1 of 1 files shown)


fixtures/fixtures.go
```diff
@@ -264,20 +264,34 @@ func (g Fixtures) Exclude(tag string) Fixtures {
 	return r
 }
 
-type Suite struct{}
-
-func (s *Suite) SetUpSuite(c *check.C) {
+// InitFixtures set the correct path to be able to access to the fixtures files
+func InitFixtures() {
```
smola: Probably `Init` and `Clean` would be better names since we're already in the fixtures packages (e.g. `fixtures.Init()`)

PR #221, opened 2017-01-24
transport/http: fix partial request with haves. Fix #216.

### LINKED ISSUE #216: Partial fetch using http protocol
Partial fetch is working in `file` and `git` protocols, but adding this test into `plumbing/transport/test/upload_pack.go` file:

```
func (s *UploadPackSuite) TestUploadPackPartial(c *C) {
	r, err := s.Client.NewUploadPackSession(s.Endpoint, s.EmptyAuth)
	c.Assert(err, IsNil)
	defer func() { c.Assert(r.Close(), IsNil) }()

	req := packp.NewUploadPackRequest()
	req.Wants = append(req.Wants, plumbing.NewHash("6ecf0ef2c2dffb796033e5a02219af86ec6584e5"))
	req.Haves = append(req.Haves, plumbing.NewHash("918c48b83bd081e863dbe1b80f8998f058cd8294"))

	reader, err := r.UploadPack(req)
	c.Assert(err, IsNil)

	s.checkObjectNumber(c, reader, 4)
}
```
It fails when `plumbing/transport/http/upload_pack_test.go` is executed:

```
FAIL: <autogenerated>:21: UploadPackSuite.TestUploadPackPartial

/plumbing/transport/test/upload_pack.go:210:
    c.Assert(err, IsNil)
... value *packfile.Error = &packfile.Error{reason:"empty packfile", details:""} ("empty packfile")
```

ajnavarro:
Using the http transport, now if we request a partial packfile, (using haves that aren't head ones) the response is not an empty packfile.

Skip this test in server implementation. Revlist is returning more objects than expected.

### DIFF
--- a/plumbing/protocol/packp/uppackreq.go
+++ b/plumbing/protocol/packp/uppackreq.go
@@ -68,7 +68,8 @@ type UploadHaves struct {
 	Haves []plumbing.Hash
 }
 
-// Encode encodes the UploadHaves into the Writer.
+// Encode encodes the UploadHaves into the Writer. If flush is true, a flush
+// command will be encoded at the end of the writer content.
 func (u *UploadHaves) Encode(w io.Writer, flush bool) error {
 	e := pktline.NewEncoder(w)
 
--- a/plumbing/protocol/packp/uppackreq.go
+++ b/plumbing/protocol/packp/uppackreq.go
@@ -69,7 +69,7 @@ type UploadHaves struct {
 }
 
 // Encode encodes the UploadHaves into the Writer.
-func (u *UploadHaves) Encode(w io.Writer) error {
+func (u *UploadHaves) Encode(w io.Writer, flush bool) error {
 	e := pktline.NewEncoder(w)
 
 	plumbing.HashesSort(u.Haves)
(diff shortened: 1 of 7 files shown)


plumbing/transport/server/upload_pack_test.go
```diff
@@ -38,3 +38,8 @@ func (s *UploadPackSuite) TestAdvertisedReferencesNotExists(c *C) {
 	c.Assert(err, Equals, transport.ErrRepositoryNotFound)
 	c.Assert(r, IsNil)
 }
+
+// TODO revList implementation is returning more objects than expected.
```
ajnavarro: I opened an issue (#222) to fix this problem.

plumbing/protocol/packp/uppackreq.go
```diff
@@ -69,7 +69,7 @@ type UploadHaves struct {
 }
 
 // Encode encodes the UploadHaves into the Writer.
```
smola: Add documentation about what does the flush parameter and why.

plumbing/transport/http/upload_pack.go
```diff
@@ -171,10 +171,13 @@ func uploadPackRequestToReader(req *packp.UploadPackRequest) (*bytes.Buffer, err
 		return nil, fmt.Errorf("sending upload-req message: %s", err)
 	}
 
-	if err := req.UploadHaves.Encode(buf); err != nil {
+	if err := req.UploadHaves.Encode(buf, false); err != nil {
```
smola: Here UploadRequest is adding a flush, but UploadHaves is not? This seems a partial fix, since it's still against the HTTP protocol as specified here: https://github.com/git/git/blob/4e59582ff70d299f5a88449891e78d15b4b3fabe/Documentation/technical/http-protocol.txt#L302

Let's be triple check this.

PR #224, opened 2017-01-25
adds Tree method to Tree

alcortesm:
This patch adds a new method to the Tree object that allows you to get a child tree from a parent tree by its relative name.

Before this patch, this was only possible with files, using the `File` method.

The new `Tree` method has a similar signature to the old `File` method for consistency.

### DIFF
--- a/plumbing/object/tree.go
+++ b/plumbing/object/tree.go
@@ -77,7 +77,7 @@ func (t *Tree) File(path string) (*File, error) {
 
 	blob, err := GetBlob(t.s, e.Hash)
 	if err != nil {
-		return nil, err
+		return nil, ErrFileNotFound
 	}
 
 	return NewFile(path, e.Mode, blob), nil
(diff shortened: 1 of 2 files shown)


plumbing/object/tree_test.go
```diff
@@ -48,6 +48,22 @@ func (s *TreeSuite) TestType(c *C) {
 	c.Assert(s.Tree.Type(), Equals, plumbing.TreeObject)
 }
 
+func (s *TreeSuite) TestTree(c *C) {
```
smola: Could you add a test for Tree(path) when path points to a Blob instead of a Tree?

alcortesm: Good catch.
Fixed in a new commit.
Also add the equivalent test for `File`, for consistency.

plumbing/object/tree.go
```diff
@@ -77,7 +77,7 @@ func (t *Tree) File(path string) (*File, error) {
 
 	blob, err := GetBlob(t.s, e.Hash)
 	if err != nil {
-		return nil, err
+		return nil, ErrFileNotFound
```
smola: There might be errors other than file not found (such as decoding or network errors). So if `err == plumbing.ErrObjectNotFound then ErrFileNotFound`, but otherwise just return the error.

plumbing/object/tree.go
```diff
@@ -91,7 +91,12 @@ func (t *Tree) Tree(path string) (*Tree, error) {
 		return nil, ErrDirectoryNotFound
 	}
 
-	return GetTree(t.s, e.Hash)
+	ret, err := GetTree(t.s, e.Hash)
+	if err != nil {
```
smola: Same as my previous comment (check the actual error).

PR #227, opened 2017-01-26
config: marshal and unmarshal done inside of the package, and submodules config file

mcuadros:
This is the implementation of the submodules files, also include the refactor of the marshal and unmashal functions for config. 

The marshalling logic was moved from the `storage.Filesystem` because at the end anyone may want marshal the struct for many other reason than just save it in a filesystem. Also we are implementing the submodule file that use the same format and is not stored in the same way.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -6,7 +6,7 @@ import (
 	"errors"
 	"fmt"
 
-	"gopkg.in/src-d/go-git.v4/plumbing/format/config"
+	format "gopkg.in/src-d/go-git.v4/plumbing/format/config"
 )
 
 const (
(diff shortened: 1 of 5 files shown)


config/modules.go
```diff
@@ -0,0 +1,133 @@
...
+	buf := bytes.NewBuffer(nil)
+	if err := config.NewEncoder(buf).Encode(m.raw); err != nil {
+		return nil, err
+	}
+
+	return buf.Bytes(), nil
+}
+
+// Submodule defines a submodule
+// https://www.kernel.org/pub/software/scm/git/docs/gitmodules.html
```
alcortesm: I see, this link is on the wrong place, move this to the Module type.

config/modules.go
```diff
@@ -0,0 +1,133 @@
...
+type Submodule struct {
```
alcortesm: are zero values safe to use?

are zero values valid (as per Validate method)?

config/config.go
```diff
@@ -56,11 +70,85 @@ func (c *Config) Validate() error {
...
+const (
+	remoteSection = "remote"
+	coreSection   = "core"
+	fetchKey      = "fetch"
+	urlKey        = "url"
+	bareKey       = "bare"
+)
+
+// Unmarshal parses a git-config file and stores it
+func (c *Config) Unmarshal(b []byte) error {
```
mcuadros: the main reason is because I tried to mimic the behavior if the Marshal/Unmarshal in other packages, since the size of the config is small, and we are marshaling the content to a content, I found this interface more familiar than others.

alcortesm: :+1:

storage/filesystem/config.go
```diff
@@ -1,43 +1,21 @@
 package filesystem
 
 import (
-	"fmt"
 	"os"
```
smola: this space shouldn't be here

Issue #225, opened 2017-01-26
Clone repo with worktree


biancalana:
Is it possible to clone a remote repo with worktree ? 

Doing a regular clone like examples:

```golang
       r, err := git.NewFilesystemRepository(directory)
        CheckIfError(err)

        // Clone the given repository to the given directory
        Info("git clone %s %s", url, directory)

        err = r.Clone(&git.CloneOptions{
                URL:   url,
                Depth: 1,
        })

```
Materialize a bare repo

mcuadros:
this feature is still not implemente, is planned and will be released in the v4 version.

biancalana:
I could try to implement that, can you elaborate a little on how to implement that ?

mcuadros:
Your help is highly appreciated but in this case we already started the implementation. We are just waiting to have the new final difftree implementation https://github.com/src-d/go-git/issues/82

PR #229, opened 2017-01-29
Worktree and new Repository Contructors

mcuadros:
This PR includes a very basic implementation of Worktree, the Checkout only can be done when the worktree is completely clean.

Also all the Repository constructors where replaced with a more clean and easy ones. 

Also some fixes discovered during the process where found like: https://github.com/src-d/go-git/commit/22cdc3bd9cd8c60984e3d8cfb89b366c3dd7d902 https://github.com/src-d/go-git/commit/ae86293c30a0e31db2054e64ab1054c345a88cc6

TODO:
- find a way to provide the Progress before Clone
- proper clean up between checkouts


common_test.go
```diff
@@ -4,13 +4,19 @@ import (
 	"fmt"
 	"testing"
 
+	"srcd.works/go-billy.v1/os"
```
smola: This import should go in the third block.

common_test.go
```diff
@@ -74,6 +85,13 @@ func (s *BaseSuite) NewRepositoryFromPackfile(f *fixtures.Fixture) *Repository {
 		panic(err)
 	}
 
+	storer.SetReference(plumbing.NewHashReference(plumbing.HEAD, f.Head))
+
+	r, err := Open(storer, memoryfs.New())
```
smola: Shouldn't this function use Init instead of Open? And delegate the SetReference to Init.

storage/memory/storage.go
```diff
@@ -72,7 +72,7 @@ func (c *IndexStorage) SetIndex(idx *index.Index) error {
 
 func (c *IndexStorage) Index() (*index.Index, error) {
 	if c.index == nil {
-		c.index = &index.Index{}
+		c.index = &index.Index{Version: 2}
```
ajnavarro: Magic number, I think the version number should be a constant.

repository.go
```diff
@@ -3,61 +3,172 @@ package git
...
 	// stored, if nil nothing is stored and the capability (if supported)
 	// no-progress, is sent to the server to avoid send this information
 	Progress sideband.Progress
 }
 
-// NewMemoryRepository creates a new repository, backed by a memory.Storage
-func NewMemoryRepository() *Repository {
-	r, _ := NewRepository(memory.NewStorage())
-	return r
+// Init create an empty git repository, based on the given Storer and worktree.
```
smola: `Init create` -> `Init creates`

worktree.go
```diff
@@ -26,17 +42,24 @@ func (w *Worktree) Checkout(commit plumbing.Hash) error {
 		return err
 	}
 
-	return files.ForEach(w.checkoutFile)
+	idx := &index.Index{Version: 2}
```
ajnavarro: Maybe Index needs a constructor to avoid version problems in a future.

repository.go
```diff
@@ -3,61 +3,172 @@ package git
...
+// The worktree Filesystem is optional, if nil a bare repository is created. If
+// the given storer is not empty ErrRepositoryAlreadyExists is returned
+func Init(s Storer, worktree billy.Filesystem) (*Repository, error) {
+	r := newRepository(s, worktree)
+	_, err := r.Reference(plumbing.HEAD, false)
+	switch err {
+	case plumbing.ErrReferenceNotFound:
+	case nil:
```
smola: `git init` can be run in existing repositories. It _reinitializes_ then. This reinitialization is safe and does not overwrite existing info (e.g. it will create the plumbing.HEAD reference if it does not exist).

repository.go
```diff
@@ -3,61 +3,172 @@ package git
...
+	if worktree == nil {
+		r.setIsBare(true)
+	}
+
+	return r, nil
+}
+
+// Open opens a git repository using the given Storer and worktree filesystem,
+// if the given storer is complete empty ErrRepositoryNotExists is returned.
+// The worktree can be nil when the repository being open is bare, if the
```
smola: `repository being open is bare` -> `epository being opened is bare`
`if a non-bare repository is not provied and worktree is nil` -> `if the repository is not bare and worktree is nil`

worktree.go
```diff
@@ -0,0 +1,294 @@
...
+	"gopkg.in/src-d/go-git.v4/plumbing"
+	"gopkg.in/src-d/go-git.v4/plumbing/format/index"
+	"gopkg.in/src-d/go-git.v4/plumbing/object"
+
+	"srcd.works/go-billy.v1"
+)
+
+var ErrWorktreeNotClean = errors.New("worktree is not clean")
+
+type Worktree struct {
```
smola: Shouldn't we expose a `NewWorktree` so users can have multiple worktrees?

examples/push/main.go
```diff
@@ -1,22 +1,20 @@
...
 	"os"
 
 	"gopkg.in/src-d/go-git.v4"
 	. "gopkg.in/src-d/go-git.v4/examples"
 )
 
 func main() {
-	CheckArgs("<repo path>")
-	repoPath := os.Args[1]
+	CheckArgs("<repository-path>")
```
ajnavarro: Here we can add a comment with an example path. After it must be a path to a .git folder. Now I think it is not necessary.

mcuadros: now it auto detects, what is being open and triggers a error, is documented on the method, I think is enough

storage/filesystem/object.go
```diff
@@ -244,9 +244,9 @@ func (s *ObjectStorage) buildPackfileIters(
 	return iters, nil
 }
 
-type index map[plumbing.Hash]int64
+type idx map[plumbing.Hash]int64
```
mcuadros: Collision with the `index` package, and this is the idx of the packfile and usually is called idx not index.

repository.go
```diff
@@ -3,61 +3,172 @@ package git
...
+		return nil, err
+	}
+
+	return Init(s, wt)
+}
+
+// PlainOpen opens a git repository from the given path. It detects is the
+// repository is bare or a normal one. If the path doesn't contain a valid
+// repository ErrRepositoryNotExists is returned
+func PlainOpen(path string) (*Repository, error) {
```
mcuadros: We already discuss this on the whiteboard, @smola @alcortesm

worktree.go
mcuadros: the custom Worktree, will be advanced feature to be develop in the future, since this require custom `.git` folders

repository.go
mcuadros: I prefer keep it like this, until this feature is required

common_test.go
mcuadros: In this case you are opening a storer that is not empty, so for me Open is correct

repository.go
ajnavarro: I didn't know that, but ok.

examples/push/main.go
ajnavarro: ok

storage/filesystem/object.go
ajnavarro: ok

mcuadros:
Fixes https://github.com/src-d/go-git/issues/225

Issue #82, opened 2016-09-27
increase performance of DiffTree, applying the same algorithm being use at git


mcuadros:

alcortesm:
See https://github.com/alcortesm/git-diff-tree-benchmarks for a description of why our old implementation needs improvements.

alcortesm:
The fix involves taking advantage of the fact that git trees are merkle trees so you don't really need to check all the files in a directory to know if they have change, as the hash of the directory can already tell you if they have change or not.

These patches introduce a generic merkle trie (it is also a n-trie) implementation and an efficient comparator that can be used to quickly compare git trees.

PR #241, opened 2017-01-30
Improved merkletrie, moved to utils.

alcortesm:
This patch adds support for https://github.com/src-d/go-git/issues/82.

The old merkletrie noder, iters and frames need a better implementation to be useful to implement a working difftree:

- The noder lacked support for returning errors when getting their children, which can happen in real life implementations of the noders.
- The iterator did not support getting full paths on its outputs so you didn't know where in the tree were the noders returned by the iterator.
- Also the hash comparison function was not part of the API, so you weren't able to overwrite it.

This patch is the basic support to solve all these issues. It removes the old merkletrie and creates a new one in `utils/merkletrie`. The new package defines the basic interface for the noders in the merkletrie and a reference implementation for tests.

The recommended order to review this patch is:

The noder interface:
- `utils/merkletrie/doc.go`
- `utils/merkletrie/hasher/hasher.go`
- `utils/merkletrie/noder/noder.go`
- `utils/merkletrie/noder/node_test.go`
- `utils/merkletrie/noder/path.go`

A reference implementation of the noder interface for tests:
- `utils/merkletrie/fsnoder/doc.go`
- `utils/merkletrie/fsnoder/dir.go`
- `utils/merkletrie/fsnoder/dir_test.go`
- `utils/merkletrie/fsnoder/file.go`
- `utils/merkletrie/fsnoder/file_test.go`
- `utils/merkletrie/fsnoder/new.go`
- `utils/merkletrie/fsnoder/new_test.go`

PS: the missing iter and frame types will come in future PRs.


alcortesm:
I still have doubts about the `noder.Path` implementation.

The current version in this patch does not have a ctor that asserts the path is not empty, therefore some of the methods will panic with empty paths.

I wrote a version `noder.Path` with a ctor that checks for empty paths, but then the noder slice has to be private and that means having an extra method for retrieving it, which means doing a copy of the noders which can be a problem if a particular implementation needs deep copies.

So in the end I stick with the unsafe implementation. I would love to know your opinions.

smola:
@alcortesm moving to a different package would have been a good case for a separate commit ;)

utils/merkletrie/fsnoder/dir.go
```diff
@@ -0,0 +1,131 @@
+package fsnoder
```
smola: This should probably be just in a _test file or an internal package.

utils/merkletrie/fsnoder/dir.go
```diff
@@ -0,0 +1,131 @@
...
+	return d.name
+}
+
+func (d *dir) IsDir() bool {
+	return true
+}
+
+// returns a copy so nobody can alter the order of its elements from the
+// outside.
+func (d *dir) Children() ([]noder.Noder, error) {
```
smola: Do we really want to do this? It seems that this kind of defensive copy is not that common in our code bases and in Go in general, specially for methods that will be called intensively in our internals. I'm not sure anyway.

utils/merkletrie/fsnoder/doc.go
```diff
@@ -0,0 +1,50 @@
+/*
+Package fsnoder allows to create merkletrie noders that resemble file
+systems, from human readable string descriptions.  Its intended use is
+generating noders in tests in a readable way.
```
smola: So... let's make it available for testing ;)

utils/merkletrie/fsnoder/file.go
```diff
@@ -0,0 +1,73 @@
+package fsnoder
+
+import (
+	"bytes"
+	"fmt"
+	"hash/fnv"
+
+	"gopkg.in/src-d/go-git.v4/utils/merkletrie/noder"
```
smola: Having a noder subpackage seems overdoing it, since noder is not meant to be used independently of merkletrees.

utils/merkletrie/fsnoder/file.go
```diff
@@ -0,0 +1,73 @@
...
+	}, nil
+}
+
+// The hash of a file is just its contents.
+// Empty files will have the fnv64 basis offset as its hash.
+func (f *file) Hash() []byte {
+	if f.hash == nil {
+		h := fnv.New64a()
+		h.Write([]byte(f.contents))
+		f.hash = []byte{}
```
smola: I think this like might not be needed, since you can pass nil to `h.Sum`.

utils/merkletrie/fsnoder/file.go
```diff
@@ -0,0 +1,73 @@
...
+		contents: contents,
+	}, nil
+}
+
+// The hash of a file is just its contents.
+// Empty files will have the fnv64 basis offset as its hash.
+func (f *file) Hash() []byte {
+	if f.hash == nil {
+		h := fnv.New64a()
+		h.Write([]byte(f.contents))
```
smola: No possible errors?

utils/merkletrie/fsnoder/file.go
```diff
@@ -0,0 +1,73 @@
...
+func (f *file) Name() string {
+	return f.name
+}
+
+func (f *file) IsDir() bool {
+	return false
+}
+
+func (f *file) Children() ([]noder.Noder, error) {
+	return noder.NoChildren, nil
```
smola: `noder.NoChildren` seems to suggest that the result is a noder with no children, while it is actually an empty slice of nodes (i.e. `nil`). Shouldn't this be just `return nil, nil`?

utils/merkletrie/fsnoder/new.go
```diff
@@ -0,0 +1,193 @@
+package fsnoder
+
+import (
+	"bytes"
+	"fmt"
+	"io"
+
+	"gopkg.in/src-d/go-git.v4/utils/merkletrie/hasher"
```
smola: Separate package for hasher? I think the same as for `noder`, looks like overdoing it.

utils/merkletrie/fsnoder/new_test.go
```diff
@@ -0,0 +1,337 @@
+package fsnoder
+
+import (
+	. "gopkg.in/check.v1"
+
+	"gopkg.in/src-d/go-git.v4/utils/merkletrie/noder"
```
smola: This should be in reverse order, first a block with go-git imports, then another with external dependencies.

utils/merkletrie/fsnoder/file_test.go
```diff
@@ -0,0 +1,67 @@
+package fsnoder
+
+import (
+	"testing"
+
+	. "gopkg.in/check.v1"
+
+	"gopkg.in/src-d/go-git.v4/utils/merkletrie/noder"
```
smola: This should be in reverse order, first a block with go-git imports, then another with external dependencies.

utils/merkletrie/fsnoder/dir_test.go
```diff
@@ -0,0 +1,364 @@
+package fsnoder
+
+import (
+	"reflect"
+	"sort"
+
+	. "gopkg.in/check.v1"
+
+	"gopkg.in/src-d/go-git.v4/utils/merkletrie/noder"
```
smola: This should be in reverse order, first a block with go-git imports, then another with external dependencies.

utils/merkletrie/noder/noder.go
```diff
@@ -0,0 +1,56 @@
...
+	// complex.
+	NumChildren() (int, error)
+}
+
+// NoChildren represents the children of a noder without children.
+var NoChildren = []Noder{}
+
+// ByName implements sort.Interface for noders, sorting them
+// alphabetically by name.
+type ByName []Noder
```
smola: This should probably be private.

utils/merkletrie/fsnoder/dir.go
alcortesm: Done

utils/merkletrie/fsnoder/dir.go
alcortesm: We don't need it, but tests are easier to build if this is true, and it does not affect the performance of the production code, as the `fsnoder` package should only be used in tests.

utils/merkletrie/fsnoder/doc.go
alcortesm: Done, it is now internal.

utils/merkletrie/fsnoder/file.go
alcortesm: It is needed though, to avoid a circular dependency **in the future**:

- `merkletrie` will import `merkletrie/frame` because the difftree function uses frames.
- `merkletrie/frames` will import `merkletrie` because frames stores noders.

The only way to solve this is to extract the noder to its own package, sadly.

utils/merkletrie/fsnoder/file.go
alcortesm: No, according to its description:
```
It never returns an error.
```
see https://golang.org/pkg/hash/#Hash for more information

utils/merkletrie/fsnoder/file.go
alcortesm: You are totally right! Fixed!

utils/merkletrie/fsnoder/file.go
alcortesm: We need a way for `Children()` to return "there are no childrens": We can use `nil` or `[]noder.Noder{}`.

As `reflect.DeepEqual(nil, []noder.Noder{})` is `false`, we should not mix both of them, so we have to use one or the other.

Nil is not a good options as it is difficult to compare children from one noder with children with the other when one of them is nil.

So the best way to mean "there is no children" is to use an empty slice.

Now, the best way to communicate that we have chosen an empty slice over nil is to define a public var for it. The definition of `noder.NoChildren` explains that:

```
// NoChildren represents the children of a noder without children.
var NoChildren = []Noder{}
```

And it is also explained at the `noder.Noder` interface definition of `Children()`:

```
// Children returns the children of the element.  Note that empty
// directory-like noders and file-like noders will both return
// NoChildren.
        Children() ([]Noder, error) 
```

I would galdly change the name of the var for something better, but nothing comes to my mind right now.

utils/merkletrie/fsnoder/new.go
alcortesm: You are right!, it was a clear case of over-engineering.

Fixed, now `hasher` is on the `noder` package.

utils/merkletrie/fsnoder/new_test.go
alcortesm: :+1:

utils/merkletrie/fsnoder/file_test.go
alcortesm: :+1:

utils/merkletrie/fsnoder/dir_test.go
alcortesm: :+1:

utils/merkletrie/noder/noder.go
alcortesm: Done.

alcortesm: rebased by @mcuadros request.

alcortesm:
Rebased by @mcuadros request.

Issue #246, opened 2017-01-30
Document utils/ioutil


Serabe:
* [x] Document package.
* [x] Try to reword second sentence in `CheckClose` (_ If Close returns an error and the given *error is not nil, *error is set to the error returned by Close._)

Issue #242, opened 2017-01-30
doc improvements at package Plumbing


Serabe:
* [x] Difftree package: document.
* [x] Format:
  * [x] Config:
    * [x] Document `Comment`
    * [x] Document `Config`
      * [x] Document undocumented methods
    * [x] Document `Include` (reword)
    * [x] Document `Options`:
      * [x] Fix format for `Get` method.
    * [x] Document `Section` and its methods.
    * [x] Document `Subsection` and its methods.
  * [x] `idxfile`:
    * [x] Reword `Decoder.Decode` given multiple uses of `idx` to refer to different things. Same for `Encoder.Encode`.
    * [x] Document `EntryList` methods
    * [x] Document `Idxfile.Add` methods
  * [x] Document `index`:
    * [x] Document `ResolveUndo` in a way it makes sense to have the name of the type at the beginning.
  * [x] Document `objfile`:
    * [x] Document package.
  * [x] Document `packfile`:
    * [x] Reword `GetDelta`
    * [x] Clarify `DecodeObjectAt`. Why if offsets are a map from hashes to ints I am required to set them before reading something using an int?
    * [x] `NewEncoder` does no document what the third argument is for.
  * [x] Package `pktline`:
    * [x] In `Encodef`, what is the `arguments suggests`?
* [x] Object
  * [x] Reword `WalkCommitHistory`. Specify what it does, what the inputs are for, what happens if the callback returns an error, etc.
  * [x] Document `DecodeBlob`.
  * [x] `NewBlobIter` fix arguments explanation
  * [x] Document `Commit` fields.
  * [x] In `NewObjectIter`, rephrase input.
  * [x] In `Signature`, what is the field that holds the action that this type represents.
* [x] Package `protocol`
  * [x] Capabilities:
    * [x] Document package
* [x] Package `revlist`:
  * [x] Documentation says that package implements functions in plural. Only one can be found.
* [x] Package `storer`:
  * [x] `ReferenceStorer` type doc update so it reads naturally.
  * [x] `ShallowStorer` type doc update so it reads naturally.
* [x] Package `transport`:
  * [x] Package `client`:
    * [x] Document package.
  * [x] Package `file`:
    * [x] Document package.
  * [x] Package `git`:
    * [x] Document package.
  * [x] Package `ssh`:
    * [x] Document package.

PR #259, opened 2017-02-07
Improve documentation

smola:


smola:
This PR fixes https://github.com/src-d/go-git/issues/246 and partially https://github.com/src-d/go-git/issues/242

plumbing/object/tag.go
```diff
@@ -20,9 +20,13 @@ import (
...
-	Tagger     Signature
+	// Hash of the tag.
+	Hash plumbing.Hash
+	// Name of the tag.
+	Name string
+	// Tagger is the one who created the tag.
+	Tagger Signature
+	// Message is an arbitrary text message
 	Message    string
 	TargetType plumbing.ObjectType
```
ajnavarro: Can you add comments to this two properties too?

plumbing/object/tag.go
```diff
@@ -20,9 +20,13 @@ import (
 //
```
ajnavarro: Add some info about [Lightweight Tags](https://git-scm.com/book/en/v2/Git-Basics-Tagging#_lightweight_tags) and say that this method does not return this kind, can help people that don't know this two kinds of tags.

smola: Done.

plumbing/object/tag.go
smola: Done.

PR #269, opened 2017-02-13
plumbing/storer: add RemoveReference

smola:


storage/filesystem/internal/dotgit/dotgit.go
```diff
@@ -299,6 +310,68 @@ func (d *DotGit) addRefsFromPackedRefs(refs *[]*plumbing.Reference) (err error)
...
+		if _, err := fmt.Fprintln(tmp, line); err != nil {
+			return err
+		}
+	}
+
+	if !found {
+		return nil
+	}
+
+	if err := s.Err(); err != nil {
```
alcortesm: I think this check must go before 350, so in case of error and not found we actually return the error instead of nil.

storage/filesystem/internal/dotgit/dotgit.go
```diff
@@ -299,6 +310,68 @@ func (d *DotGit) addRefsFromPackedRefs(refs *[]*plumbing.Reference) (err error)
...
+		return err
+	}
+
+	backupPackedRefs := packedRefsPath + ".bkp"
```
alcortesm: This function is already too long.

This last part is a very good candidate to be moved to a function called safeRename() or something similar. A good name will also help the reader to identify the purpose of 3 consecutive renames.

storage/filesystem/internal/dotgit/dotgit.go
smola: :+1: done

storage/filesystem/internal/dotgit/dotgit.go
smola: I've just removed this and used billy's Rename directly. Doing our own SafeRename external to billy gets messy really fast (since guarantees depend both on operating system, file system and billy Filesystem implementation). So let's rely on billy implementations to provide guarantees as strong as they can on their own.

plumbing/storer/reference.go
```diff
@@ -18,6 +18,7 @@ type ReferenceStorer interface {
 	SetReference(*plumbing.Reference) error
 	Reference(plumbing.ReferenceName) (*plumbing.Reference, error)
 	IterReferences() (ReferenceIter, error)
+	RemoveReference(plumbing.ReferenceName) error
```
ajnavarro: Some documentation indicating at least that could not be thread safe?

smola: Being thread-safe or not is implementation-dependent. So far, none of our public implementations is thread-safe for any write operation.

PR #292, opened 2017-02-27
plumbing/revlist: input as a slice of hashes instead of commits

### LINKED ISSUE #222: Revlist implementation returns more objects than expected.
`plumbing/revlist/revlist.go` only ignores the hashes in `ignore` param. We should ignore all the objects that are reachable from `ignore` ones.

ajnavarro:
- Now the input of the method Objects inside revlist package is a slice of hashes instead of commits. Hashes can be from Blobs, Trees and Commits objects.
- ObjectStorer now is used to obtain the object content using hashes slice.
- This PR fix #222. Now a test into upload_pack_test.go file is not skipped anymore.
- Remove code from remote.go and server.go that is not necessary.

### DIFF
--- a/plumbing/revlist/revlist.go
+++ b/plumbing/revlist/revlist.go
@@ -3,6 +3,7 @@
 package revlist
 
 import (
+	"fmt"
 	"io"
 
 	"srcd.works/go-git.v4/plumbing"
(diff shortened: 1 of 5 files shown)


plumbing/revlist/revlist.go
```diff
@@ -11,30 +12,56 @@ import (
 )
 
 // Objects applies a complementary set. It gets all the hashes from all
-// the reachable objects from the given commits. Ignore param are object hashes
+// the reachable objects from the given objects. Ignore param are object hashes
 // that we want to ignore on the result. It is a list because is
```
smola: `It is a list because is easier to interact with other porcelain elements, but internally it is converted to a map.`

This comment is about the internal implementation and users do not care too much about this. Could you just remove this part?

PR #295, opened 2017-02-28
improve git package documentation (fix #231)

### LINKED ISSUE #231: doc improvements at package git epic
* [x] Review package doc (remove references to it being a _low-level_ package, etc.)
* [ ] Highlight general steps to work with repositories. Guide the user a bit through the documentation.
* [x] Why `Commit` is treated like a getter while `Push`, `Pull`, etc. are treated like commands?
* [x] Review `References` function documentation:
  * [x] States that it returns a `References` while it does not.
  * [x] Does not specify what the commit input is for.
  * [x] State if the ordering is from newer to older or the other way around.
  * [x] Reword sentence "It stops searching a branch for a file upon reaching the commit were the file was created."
  * [x] Fix _Caveats_ section.
* [x] Document `BlameResult`
* [x] In Blame function:
  * [x] Review output documentation. It states that the last commit is returns, but the return type is not an object.Commit.
  * [x] Review input. It states that there are three inputs, but I can only see two in the signature. What is the commit for? What is the path for? Why is the output documented again after the input with a different version of it?
  * [x] Says that `Line` has a reference to the last commit it was modified on, but the `Line` struct does not seem to have a commit in it.
  * [x] _Each node (line) holds the commit where it was introduced or last modified. To achieve that we use the FORWARD algorithm_. The previous phrase action is _holding a commit_, so I guess we are not using an algorithm to just _hold a commit_ but to find which commit. Rephrase so the pronoun references what is intended to reference.
  * [x] See _General questions about docs_ later in this issue.
* [x] `Line` documentation should be synced with that of `Blame`.
* [x] `XOptions` documentation:
  * [x] Typos fixed in PR #230
  * [x] Explain `Validate`, it seems like a function the user should not care about. If the user should, please state that it is mandatory to be called since it is the one setting the default value.
* [x] `Remote`:
  * [x] ELI5 `Config`.
  * [x] Why do we need to add the remote name to `XOptions` if the action is being done in a Remote?
* [x] `Repository`:
  * [x] Struct doc is messed up.
  * [x] Update `NewXRepository` to reference `go-billy`.
  * [x] Either add the default version (_if you want to use a custom one you need to use the function NewRepository and build you filesystem.Storage_) to all `NewXRepository` methods or, even better, add this to the struct.
  * [x] `NewRepository` is outdated respect to its signature.
  * [x] Unify documentation for getters that return an iterator. Specify clearly that the returned type is an iter and specify the type of the objects being iterated over. In case it is needed, specify the order.
  * [x] What is the difference between `Objects` and `Tags`? Documentation looks the same.
* [x] Storer:
  * [x] Reference where the user can find implementations or create their own. Rewords the part of implementations

## General questions about docs:

* Should we avoid adding TODOs to godocs? IMHO, TODOs are not documentation and the user should not care about them. If anything about them needs to be added to the doc, is which things are not supported.
* Should we avoid stating how things are implemented? IMHO, documentation is not about how something works but why it is there, how the user can take advantage of it, etc. If user needs to know anything about implementation, I would say it is limited to algorithmic complexity.

ajnavarro:


blame.go
```diff
@@ -13,17 +13,21 @@ import (
 	"srcd.works/go-git.v4/utils/diff"
 )
 
+// BlameResult represents the result of a Blame operation
 type BlameResult struct {
-	Path  string
-	Rev   plumbing.Hash
+	// Path is the path of the Blob object that we're blaming
```
smola: We blame files, not blobs.

blame.go
```diff
@@ -13,17 +13,21 @@ import (
...
+	Path string
+	// Rev is the hash of the specified Commit used to generate this result
+	Rev plumbing.Hash
+	// Lines contains the last person that modifies this line, and the content itself
```
smola: Not very clear. Lines contains lines, not person. Maybe reword:
`Lines contains the last person that modifies this line, and the content itself` -> `Lines contrains every line with its authorship`

blame.go
```diff
@@ -98,8 +96,10 @@ func Blame(c *object.Commit, path string) (*BlameResult, error) {
 
 // Line values represent the contents and author of a line in BlamedResult values.
 type Line struct {
-	Author string // email address of the author of the line.
-	Text   string // original text of the line.
+	// email address of the author of the line.
```
smola: Suggested reword: `Author is the email address of the last author that modified the line`

blame.go
```diff
@@ -98,8 +96,10 @@ func Blame(c *object.Commit, path string) (*BlameResult, error) {
...
+	Author string
+	// original text of the line.
```
smola: Add `Text` at the begining, such as `Text is the original text of the line`

blame.go
```diff
@@ -13,17 +13,21 @@ import (
 	"srcd.works/go-git.v4/utils/diff"
 )
 
+// BlameResult represents the result of a Blame operation
```
alcortesm: Add a full stop at the end of the sentence, as it is going to be rendered as English prose by go-doc.

blame.go
```diff
@@ -13,17 +13,21 @@ import (
...
 type BlameResult struct {
-	Path  string
-	Rev   plumbing.Hash
+	// Path is the path of the File that we're blaming
```
alcortesm: full stop.

blame.go
```diff
@@ -13,17 +13,21 @@ import (
...
+	Path string
+	// Rev is the hash of the specified Commit used to generate this result
```
alcortesm: Full stop.

Maybe you should mention that rev is the abbreviation of revision, or change the field name to `Revision`.

blame.go
```diff
@@ -13,17 +13,21 @@ import (
...
+	Rev plumbing.Hash
+	// Lines contains every line with its authorship
```
alcortesm: Full stop.

blame.go
```diff
@@ -13,17 +13,21 @@ import (
...
 }
 
-// Blame returns the last commit that modified each line of a file in a
-// repository.
-//
-// The file to blame is identified by the input arguments: repo, commit and path.
-// The output is a slice of commits, one for each line in the file.
+// Blame returns a BlameResult that contains all the data needed to know the
+// last author of each line of an specified file starting the history from
+// a specified commit. The file to blame is identified by the input arguments:
```
alcortesm: I don't understand what you mean here by "starting the history from a specified commit".

blame.go
```diff
@@ -33,12 +37,6 @@ type BlameResult struct {
 // 2. Then build a graph with a node for every line in every file in
 // the history of the file.
 //
-// Each node (line) holds the commit where it was introduced or
-// last modified. To achieve that we use the FORWARD algorithm
-// described in Zimmermann, et al. "Mining Version Archives for
-// Co-changed Lines", in proceedings of the Mining Software
-// Repositories workshop, Shanghai, May 22-23, 2006.
-//
```
alcortesm: Why did you remove this comment? how are you going to understand the current implementation without the paper describing the algorithm? it is probably the most important comment in the whole file.  I agree it should not be public, though, it should be private as it explains the implementation details.

blame.go
```diff
@@ -98,8 +96,10 @@ func Blame(c *object.Commit, path string) (*BlameResult, error) {
 
 // Line values represent the contents and author of a line in BlamedResult values.
 type Line struct {
-	Author string // email address of the author of the line.
-	Text   string // original text of the line.
+	// Author is the email address of the last author that modified the line
```
alcortesm: Full stop.

references.go
```diff
@@ -10,15 +10,18 @@ import (
...
+// returned. The commits are sorted in commit order, newer to older.
 //
 // Caveats:
+//
 // - Moves and copies are not currently supported.
+//
 // - Cherry-picks are not detected unless there are no commits between them and
-//   therefore can appear repeated in the list.
-//   (see git path-id for hints on how to fix this).
+// therefore can appear repeated in the list. (see git path-id for hints on how to fix this).
```
alcortesm: This comment line looks longer than the rest, it may be longer than 60 or 70 chars (or whatever the line length we are using for comments).

remote.go
```diff
@@ -32,7 +32,7 @@ func newRemote(s storage.Storer, c *config.RemoteConfig) *Remote {
 	return &Remote{s: s, c: c}
 }
 
-// Config return the config
+// Config returns the RemoteConfig object used to instantiate this Remote
```
alcortesm: Full stop.

blame.go
ajnavarro: Regarding to https://github.com/src-d/go-git/issues/231 (General questions) how the things are implemented is not documentation.

I will keep this comment inside the method.

blame.go
ajnavarro: The specified commit is the commit that you are using to call Blame. The BlameResult will change depending of the commit you will use.

blame.go
alcortesm: :+1:

blame.go
alcortesm: Some alternative rewordings:

A) This is correct, but maybe too short and dependent on the BlameResult definition.

```
Blame returns a BlameResult for file `path` in commit `c`.
```

B) This one is more user friendly.
```
Blame returns a BlameResult with the information about the
last author of each line from file `path` at commit `c`.
```

The original wording, written by me, was too verbose, because I was
not familiarized with go-git back then and I feel the need to explain
how the same file in different commits are different entries.  Now I
clearly see being so verbose is no good.

I will also make private the rest of the paragraphs in this comment, as
you did with the one referencing the paper.

ajnavarro: :+1:

PR #296, opened 2017-03-01
git: Repository methods changes

ajnavarro:
To have a more consistent public API, we decided to rename some methods and add others:

- Commit method renamed to CommitObject
- Commits method renamed to CommitObjects
- Tags method renamed to TagObjects
- Tag method renamed to TagObject
- Added method Tags that returns tag references
- Added method Branches that returns branch references
- Added method Notes that returns note references
- Added BlobObject method
- Added BlobObjects method
- Remove Tree method
- Remove Trees method

Also, we added more functionality related to references:
- Added iterator to iterate References with a specific filter

Some notes:
- #298

### DIFF
--- a/plumbing/storer/reference.go
+++ b/plumbing/storer/reference.go
@@ -28,6 +28,67 @@ type ReferenceIter interface {
 	Close()
 }
 
+type ReferenceFilteredIter struct {
+	ff   func(r *plumbing.Reference) bool
+	iter ReferenceIter
+}
+
+// NewReferenceFilteredIter returns a reference iterator for the given reference
+// Iterator. This iterator will iterate only references that accomplish the
+// provided function.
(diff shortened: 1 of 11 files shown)


mcuadros:
Why we are removing Tree?

ajnavarro:
@mcuadros because we talked about this. Get a specific Tree from Porcelain is a little weird, or a list with all the Trees.

I can keep these methods if it's necessary.

plumbing/storer/reference.go
```diff
@@ -28,6 +28,67 @@ type ReferenceIter interface {
 	Close()
 }
 
+type ReferenceFilteredIter struct {
```
smola: let's make this type private.

thefallentree:
👍

PR #302, opened 2017-03-06
Support slash separated branch

hatajoe:
I often use a slash-separated branch name. (e.g, issue/10-blah) 
However, `plumbing.ReferenceName.Short` does not seem to support it.
So how about doing this? (I'm sorry, I'm not very familiar with git)

### DIFF
--- a/plumbing/reference.go
+++ b/plumbing/reference.go
@@ -38,7 +38,7 @@ func (r ReferenceName) String() string {
 // Short returns the short name of a ReferenceName
 func (r ReferenceName) Short() string {
 	parts := strings.Split(string(r), "/")
-	return parts[len(parts)-1]
+	return strings.Join(parts[2:], "/")
 }
 
 const (
--- a/plumbing/reference.go
+++ b/plumbing/reference.go
@@ -16,7 +16,7 @@ const (
 )
 
 var (
-	prefixes = []string{
+	refPrefixes = []string{
 		refHeadPrefix,
 		refTagPrefix,
 		refRemotePrefix,
@@ -54,7 +54,7 @@ func (r ReferenceName) Short() string {
 // refHeadPrefix, refTagPrefix, refRemotePrefix, refNotePrefix and refPrefix
 func (r ReferenceName) short() string {
 	s := string(r)
-	for _, prefix := range prefixes {
+	for _, prefix := range refPrefixes {
 		s = strings.TrimPrefix(s, prefix)
 	}
 	return s
--- a/plumbing/reference.go
+++ b/plumbing/reference.go
@@ -47,12 +47,12 @@ func (r ReferenceName) String() string {
 
 // Short returns the short name of a ReferenceName
 func (r ReferenceName) Short() string {
-	return r.short()
+	return r.removeRefPrefix()
 }
 
 // Instead of hardcoding a number of components, we should remove the prefixes
 // refHeadPrefix, refTagPrefix, refRemotePrefix, refNotePrefix and refPrefix
-func (r ReferenceName) short() string {
+func (r ReferenceName) removeRefPrefix() string {
 	s := string(r)
 	for _, prefix := range refPrefixes {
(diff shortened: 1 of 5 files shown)


smola:
Instead of hardcoding a number of components, we should remove the prefixes refHeadPrefix, refTagPrefix, refRemotePrefix, refNotePrefix and refPrefix (constants defined in reference.go).

unapiedra:
I like this pull request. I've added a unit-test for this. Please add the unit test to this pull request. I also tried the suggestion by @smola.

1. Here is a unit-test:

```
// L21 of plumbing/reference_test.go
func (s *ReferenceSuite) TestReferenceNameWithSlash(c *C) {
	r := ReferenceName("refs/remotes/origin/feature/AllowSlashes")
	c.Assert(r.Short(), Equals, "origin/feature/AllowSlashes")
}
```

The statement code coverage does not increase when using this test, because the statements are already covered by `TestReferenceNameShort`. However, it does test additional functionality. 

Also, I think the reason my next suggestion increases the relative code coverage is because the code is longer than the one before, and all of the added lines are covered through tests. That does not make this better.

2. Here is @smola's suggestion implemented:
```
// Short returns the short name of a ReferenceName
func (r ReferenceName) Short() string {
	/*
	Instead of hardcoding a number of components, we should remove the prefixes
	refHeadPrefix, refTagPrefix, refRemotePrefix, refNotePrefix and refPrefix
	(constants defined in reference.go).
	*/
	s := string(r)
	prefixes := []string{
		refHeadPrefix,
		refTagPrefix,
		refRemotePrefix,
		refNotePrefix,
		refPrefix}
	for _, prefix := range prefixes {
		s = strings.TrimPrefix(s, prefix)
	}
	return s
}
```

3. You could also do @hatajoe's solution in a single line, however I am not saying that this is nicer. This would be it:
```
// Short returns the short name of a ReferenceName
func (r ReferenceName) Short() string {
	return strings.SplitN(string(r), "/", 3)[2]
}
```

unapiedra:
You may find the unittest code here: https://github.com/hatajoe/go-git/pull/1

hatajoe:
@smola @unapiedra Thank you for a nice suggestion. Pushed the fix.

plumbing/reference.go
```diff
@@ -37,8 +37,22 @@ func (r ReferenceName) String() string {
 
 // Short returns the short name of a ReferenceName
 func (r ReferenceName) Short() string {
-	parts := strings.Split(string(r), "/")
-	return parts[len(parts)-1]
+	/*
```
mcuadros: can you move this to a private function? and put the comment there?

plumbing/reference.go
```diff
@@ -37,8 +37,22 @@ func (r ReferenceName) String() string {
...
+		Instead of hardcoding a number of components, we should remove the prefixes
+		refHeadPrefix, refTagPrefix, refRemotePrefix, refNotePrefix and refPrefix
+		(constants defined in reference.go).
+	*/
+	s := string(r)
+	prefixes := []string{
```
mcuadros: can you put the slice initialization outside of the function?

hatajoe:
@mcuadros Thank you for review. I pushed the fix. Is this OK?

plumbing/reference.go
```diff
@@ -37,8 +47,17 @@ func (r ReferenceName) String() string {
...
 // Short returns the short name of a ReferenceName
 func (r ReferenceName) Short() string {
-	parts := strings.Split(string(r), "/")
-	return parts[len(parts)-1]
+	return r.short()
+}
+
+// Instead of hardcoding a number of components, we should remove the prefixes
+// refHeadPrefix, refTagPrefix, refRemotePrefix, refNotePrefix and refPrefix
+func (r ReferenceName) short() string {
```
mcuadros: change to `removeRefPrefix` and is ready to merge

hatajoe:
@mcuadros Thank you!

PR #305, opened 2017-03-12
plumbing/object: add WalkCommitHistoryPost func

### LINKED ISSUE #223: object.WalkCommitHistory should walk parents in reverse order
When you do `git log` and there has been a merge, you see something like:

* merge commit
* merged commit
* rest of history

But the way `WalkCommitHistory` is implemented, the first parent is walked first. This means that the history is traversed as:

* merge commit
* rest of history
* merged commit

I see no advantage in doing it this way. Using it the other way is more intuitive, and should also use less memory as merged histories are often much shorter (e.g. PRs) than the main ones.

mvdan:


mvdan:
As for the test - tried modifying `TestWalkerNext` to use this func as a start, but surprisingly got the wrong number of commits - 10 instead of 8. I presume this is because my implementation doesn't check for duplicates, while the other does. Should it?

Also don't know how to test this func properly. Ideally, both funcs would have a test that verifies the order is correct. In practice, this test suite is a bit obscure and some pointers on how to do that with this mock git repository/history would be of great help.

mvdan:
@smola @mcuadros

mvdan:
A seen map got rid of the len error, so did that.

Perhaps the way to have the tests for the two funcs actually test the order would be to avoid `SortCommits(commits)` and instead check for the expected hashes in the correct order. Are any of those 8 commits merge commits?

smola:
@mvdan for testing, you can call to WalkCommitHistoryPost passing a function that adds commit hashes to a slice, then you just check the slice.

mvdan:
@smola that's what the other test already does (minus the sort). But my question is whether those set of commits contain any merge commits. If they don't, the order obtained from the two functions would be exactly the same and the tests would not be as useful.

mvdan:
ping @smola

smola:
@mvdan Sorry for the delay!

Yes. The `basic` fixture contains 2 merges. There is a copy of it at GitHub: https://github.com/git-fixtures/basic/commits/master where you can check the details.

mvdan:
Awesome thanks - will update the PR with proper tests.

mvdan:
Done, PTAL.

smola:
Thanks @mvdan!

Issue #249, opened 2017-02-01
Bug: bufio: buffer full


icambridge:
I'm currently using master from last night.

I get the following error message. I assume it'll be one of the three repositories that had just started processing. (I clearly need to improve the error logging to include the repository).

```
INFO[0362] Started processing repository https://github.com/zgpmax/Email-Sender 
INFO[0362] Started processing repository https://github.com/mingzong/connman 
INFO[0363] Started processing repository https://github.com/ricardocarneiro/jquery-1 
INFO[0363] Started processing repository https://github.com/flytzen/Autofac 
ERRO[0363] bufio: buffer full                           
```

With the error coming from the following code:

```go
func (gc *GitRepository) Commits() (*[]Commit, error) {

	ref, err := gc.repository.Head()
	if err != nil {
		return nil, err
	}

	// ... retrieves the commit object
	commit, err := gc.repository.Commit(ref.Hash())
	if err != nil {
		return nil, err
	}
	commits, err := commit.History()

	if err != nil {
		return nil, err
	}
	output := []Commit{}
	for _, orgCommit := range commits {

		if err != nil {
			break
		}
		newCommit := convertCommit(*orgCommit)
		output = append(output, newCommit)
	}

	return &output, err
}
```

Which i think comes from https://github.com/src-d/go-git/blob/965b8e48ca111b893de9c133aa5843689a93120c/plumbing/object/tag.go#L85. Since it's the only method that you guys use that returns it that I can see in https://golang.org/pkg/bufio/.

smola:
@icambridge Could you try reproducing it with a single repository or a use case as minimal as possible? Thank you!

mcuadros:
Please let you know if still is happening

hansrodtang:
This happens for me on this [Chromium repository](https://chromium.googlesource.com/chromium/src/), using `object.WalkCommitHistory` on master, the point of the error for me was `r.ReadSlice` in [commit.Decode](https://github.com/src-d/go-git/blob/master/plumbing/object/commit.go#L133). 

Starting commit hash: `b6b96c3de7c6db78762ce660b274514bd18a1c0b`.

In the walker I just placed a simple counter, the last one counted was 451 (hash: `6d3b0113cb987a958bc03faf494a37a3750aff5f`).

`git rev-list --count master` goes to 578829 and if I replace `ReadSlice` with `ReadBytes` as a quick workaround,  my `WalkCommitHistory` test reaches the same number.

lupine:
[...]

PR #314, opened 2017-03-27
plumbing: Use ReadBytes() rather than ReadSlice()

lupine:
Fixes https://github.com/src-d/go-git/issues/249

### DIFF
--- a/plumbing/object/commit.go
+++ b/plumbing/object/commit.go
@@ -130,7 +130,7 @@ func (c *Commit) Decode(o plumbing.EncodedObject) (err error) {
 
 	var message bool
 	for {
-		line, err := r.ReadSlice('\n')
+		line, err := r.ReadBytes('\n')
 		if err != nil && err != io.EOF {
 			return err
 		}
--- a/plumbing/object/commit_test.go
+++ b/plumbing/object/commit_test.go
@@ -2,6 +2,7 @@ package object
 
 import (
 	"io"
+	"strings"
 	"time"
 
 	"github.com/src-d/go-git-fixtures"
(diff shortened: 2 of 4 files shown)


mcuadros:
Can you provide a test for it? To avoid duplicate this bug again? A slice with more than 4096 chars without a newline should be enough for test the issue. Thanks!

hansrodtang:
I am not sure what the original reason for using `ReadSlice()`, but `ReadBytes()` makes a copy of the data read, so won't this change put more pressure on the GC?

Or is this deemed not an issue? I haven't looked into how much the memory usage increases with the change, just a quick observation that might be unimportant.

lupine:
@mcuadros the existing test suite seems to rely  heavily on a set of external fixtures: github.com/src-d/go-git-fixtures

Do I issue a PR against that repo to get a fixture added? 

@hansrodtang it will result in an extra copy per line but we're already doing at least one copy per line, so I don't think it's a major issue - we're not changing the scaling behaviour fundamentally by this change.

lupine:
Ah, never mind, I found a way to add tests without changing fixtures. Wdyt @mcuadros ? I verified that the tests fail without the fix.

lupine:
If accepted, will this fix end up in `gopkg.in/src-d/go-git.v4` automatically? Or is there a process / timeline for getting it added?

mcuadros:
@hansrodtang thanks for pointing the extra memory usage, once we have a stable API, we will take care of the performance reviewing the code base to improve this from the root.

@lupine yep, I need update the v4 branch, but I will do it ASAP.

mcuadros:
@lupine v4 has been updated, https://github.com/src-d/go-git/tree/v4. 
BTW if you don't mind... may I know for what are you using go-git?

lupine:
@mcuadros https://gitlab.com/gitlab-org/es-git-go at the moment. i'm starting to see uses for it elsewhere too, though :D

lupine:
Thanks so much for the merge!

PR #319, opened 2017-03-29
references.go: fix Parents from commit iterator

ajnavarro:
Avoid a panic when we are iterating parents from a commit. There are more than EOF errors, like Parse errors. Now instead of throw a panic we return the error.

### DIFF
--- a/references.go
+++ b/references.go
@@ -76,8 +76,10 @@ func walkGraph(result *[]*object.Commit, seen *map[plumbing.Hash]struct{}, curre
 
 	// optimization: don't traverse branches that does not
 	// contain the path.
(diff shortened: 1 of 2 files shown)


mcuadros:
How hard is implement a test for this?

ajnavarro:
I'll have a look to how to test this. One approximation could be have a correct Commit object that some of his parents don't exists into the ObjectStorer.

alcortesm:
It should be pretty easy if we mock the ObjectStorer.

smola:
You can use an in-memory ObjectStorer for this.

ajnavarro:
@smola @mcuadros @alcortesm Added tests using in-memory ObjectStorer.

PR #324, opened 2017-04-05
object: fix Change.Files() method behavior (fix #317)

### LINKED ISSUE #317: "object not found" error in `object.Change#Files()` when `To` is a submodule
Discovered after updating to the latest version of `gopkg.in/src-d/go-git.v4`

I use `object#DiffTree()` and step through the changes, unconditionally calling `change.Files()` for each. When I come across the addition of a submodule, I get the aforementioned error, when before I would not.

Does it make more sense for `File()` to return `nil` instead for these cases? Presumably there's a similar issue with the `From` side as well.

ajnavarro:
- If 'from' or 'to' are tree entries that aren't files, Files() method will return nil instead of object not found error.
- Added a test checking this using modules fixture.

### DIFF
--- a/plumbing/object/change.go
+++ b/plumbing/object/change.go
@@ -46,6 +46,10 @@ func (c *Change) Files() (from, to *File, err error) {
 
 	if action == merkletrie.Insert || action == merkletrie.Modify {
 		to, err = c.To.Tree.TreeEntryFile(&c.To.TreeEntry)
+		if !c.To.TreeEntry.Mode.IsFile() {
+			return nil, nil, nil
+		}
+
 		if err != nil {
 			return
 		}
@@ -53,6 +57,10 @@ func (c *Change) Files() (from, to *File, err error) {
 
 	if action == merkletrie.Delete || action == merkletrie.Modify {
 		from, err = c.From.Tree.TreeEntryFile(&c.From.TreeEntry)
+		if !c.From.TreeEntry.Mode.IsFile() {
+			return nil, nil, nil
+		}
+
 		if err != nil {
 			return
 		}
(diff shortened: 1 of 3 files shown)


ajnavarro:
Please, merge https://github.com/src-d/go-git-fixtures/pull/4 first.

plumbing/object/change.go
```diff
@@ -61,6 +70,12 @@ func (c *Change) Files() (from, to *File, err error) {
 	return
 }
 
+func isInvalidFileMode(t TreeEntry) bool {
```
mcuadros: Please use the FileMode methods: https://godoc.org/gopkg.in/src-d/go-git.v4/plumbing/filemode

I think that is this method can be replaced with a call to `FileMode.IsRegular()`

ajnavarro: @mcuadros I think `FileMode.IsRegular()` doesn't cover all the cases, like executable files. Maybe @alcortesm could do us some hints.

mcuadros: Then you have `IsFile`. Why the symlink is being skip, is a regular file.

alcortesm: Use `Filemode.IsFile()`, that's what @santi and me agreed when I answered https://github.com/src-d/go-git/issues/317.

smola: @alcortesm It's @smola ;-)

alcortesm: lol, sorry!

smola: Right. I was confused when we talked this morning.
@ajnavarro Symlinks can be returned by Files()

Issue #328, opened 2017-04-06
ssh HostKeyCallback now required


adampointer:
The ssh library has introduced a breaking change where HostKeyCallback is required in ClientConfig whereas before it's absence would simply disable verification altogether. When cloning a repository via SSH you now get this error:

`ssh: must specify HostKeyCallback`

I have forked this and written a very basic callback which will verify host keys against known_hosts. I do not expect my PR to be accepted as it is tbh as its kind of rubbish and needs more thought. But I thought it may be of some use to illustrate the problem.

smola:
@adampointer Thanks! Your solutions seems the right one to me. You could add the following in a PR:

* Apply the same defaults to `ssh.Password` auth method.
* Document in PublicKeys/Password godoc that they use known_hosts.
* Make `clientConfig()` public in `ssh.AuthMethod`.

This way users would be able to easily override this behavior (for example, to use an in-memory map for their own custom known hosts).

taralx:
For now can we put in a stub? Hard to contribute when the test is broken. :disappointed:

mcuadros:
@taralx you can fix golang.org/x/crypto/ssh to a old version previous to this change. Or merge the PR on your local branch

smola:
@taralx As stated in https://github.com/src-d/go-git/issues/332, we will change SSH tests to use a local server. If you want it to be implemented faster, PRs are welcome ;)

kris-nova:
So running this new code, with the new defaults I am getting the following:

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x139851a]

goroutine 1 [running]:
github.com/kris-nova/klone/vendor/golang.org/x/crypto/ssh.publicKeyCallback.auth(0x0, 0xc42017cc40, 0x20, 0x20, 0x0, 0x0, 0x1853260, 0xc420441080, 0x184cda0, 0xc420019500, ...)
	/go/src/github.com/kris-nova/klone/vendor/golang.org/x/crypto/ssh/client_auth.go:187 +0x3a
github.com/kris-nova/klone/vendor/golang.org/x/crypto/ssh.(*connection).clientAuthenticate(0xc420334a80, 0xc42015e240, 0x0, 0xa)
	/go/src/github.com/kris-nova/klone/vendor/golang.org/x/crypto/ssh/client_auth.go:36 +0x369
[...]

PR #329, opened 2017-04-06
Add basic host key verification

adampointer:
Refs #328 

Very basic implementation of host key verification.

### DIFF
--- a/plumbing/transport/ssh/auth_method.go
+++ b/plumbing/transport/ssh/auth_method.go
@@ -1,16 +1,20 @@
 package ssh
 
 import (
+	"bufio"
 	"errors"
 	"fmt"
 	"net"
 	"os"
+	"path/filepath"
+	"strings"
 
 	"golang.org/x/crypto/ssh"
 	"golang.org/x/crypto/ssh/agent"
 )
 
 var ErrEmptySSHAgentAddr = errors.New("SSH_AUTH_SOCK env variable is required")
+var certChecker = &ssh.CertChecker{}
 
 // AuthMethod is the interface all auth methods for the ssh client
 // must implement. The clientConfig method returns the ssh client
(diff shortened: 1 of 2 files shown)


smola:
From https://github.com/src-d/go-git/issues/328#issuecomment-292168389

Your solutions seems the right one to me. You could add the following in a PR:

* Apply the same defaults to `ssh.Password` auth method.
* Document in PublicKeys/Password godoc that they use known_hosts.
* Make `clientConfig()` public in `ssh.AuthMethod`.

This way users would be able to easily override this behavior (for example, to use an in-memory map for their own custom known hosts).

Xfennec:
It's maybe obvious, but while trying this PR, I discovered in `man 8 sshd` that `known_hosts` format can be a bit more complex than that:

- there's up to 5 fields ("markers (optional), hostnames, keytype, base64-encoded key, comment")
- key can be revoked (@revoked marker)
- hostnames are patterns (`*` and `?` act as wild-cards, `!` for negation, `[]` for address with optionnal `:` for port)
- lines may be commented out
- hostnames may be stored in a hashed form (start with `|`, it seems it's the default for [some ?] Debian)

The provided example is scary:
```
        # Comments allowed at start of line
        closenet,...,192.0.2.53 1024 37 159...93 closenet.example.net
        cvs.example.net,192.0.2.10 ssh-rsa AAAA1234.....=
        # A hashed hostname
        |1|JfKTdBh7rNbXkVAQCRp4OQoPfmI=|USECr3SWf1JUPsms5AqfD5QfxkM= ssh-rsa
        AAAA1234.....=
        # A revoked key
        @revoked * ssh-rsa AAAAB5W...
        # A CA key, accepted for any host in *.mydomain.com or *.mydomain.org
        @cert-authority *.mydomain.org,*.mydomain.com ssh-rsa AAAAB5W...
```
Writing a correct parser for this file requires a bit of work. Even a "very basic" parser, like this PR, is probably useless, since it does not work for Debian (7 at least) from my experience.

adampointer:
I agree, I had to fudge my known_hosts to get this to work as of did not like the :22. Feels like there should be a third party library for this. Should we need to reinvent the wheel here? I will have a look around and see of someone had already solved this.

smola:
Other projects are doing their own thing: https://github.com/coreos/fleet/blob/master/ssh/known_hosts.go

It seems fleet does skip the revoked marker, which doesn't sound right. Unless there is a lightweight third-party library for this, I think we should add proper support in our code base.

Xfennec:
If it can be of any help, I've implemented a more complete parser yesterday. It lacks wildcard support, but all other `known_hosts` "features" referenced in my previous post are supported. I think : https://github.com/Xfennec/nosee/blob/19c2994ddb72c0725337120925efff8e9382e1a2/ssh.go#L58-L132

smola:
@Xfennec @a17n Did you check this: https://godoc.org/golang.org/x/crypto/ssh#ParseKnownHosts

Xfennec:
@smola Of course, it's used in my code (and in your Fleet source code link, I think) but it's doing a very small part of the job.

adampointer:
I will look to replicate the way @Xfennec has done this. We don't need to capture every edge case for this as long as of covers the majority of situations. It can always be iterated upon once merged. 

I will get cracking on this next week after the long Easter weekend. 

Thanks for everybody's input!

peefourtee:
FYI a [recent golang.org/x/crypto/ssh/knownhosts commit](https://github.com/golang/crypto/commit/ed779e1bec0180cdfce8135ca6558067b388777b) added https://godoc.org/golang.org/x/crypto/ssh/knownhosts#New for creating a `ssh.HostKeyCallback` from knownhosts

mcuadros:
Closed in favor of https://github.com/src-d/go-git/pull/346

mcuadros:
Thanks everyone

PR #330, opened 2017-04-06
Lazily load object index.

### LINKED ISSUE #327: Open scans object index even if not required
My app opens a repo, checks for a reference, and only does work if the reference is missing. Which means that the object index scan is completely wasted most of the time. Can we add an option to defer it until a later point? Maybe make it lazy.

taralx:
fixes #327


taralx:
Hm, will fix tests.

PR #331, opened 2017-04-06
add git checkout example + housekeeping

marwan-at-work:
Added a `git checkout <ref>` examples. 

One thing that's a bit different from the CLI command, is that when you do `git checkout <ref>` in the git CLI, your repo gets checked out in a detached state. However, go-git keeps you in the same branch with the added difference. Not sure if that's a bug or a feature. Or maybe there's another way to do checkout that mimics the CLI results.

### DIFF
--- a/_examples/checkout/main.go
+++ b/_examples/checkout/main.go
@@ -15,11 +15,10 @@ func main() {
 	url, directory, commitRef := os.Args[1], os.Args[2], os.Args[3]
 
 	// Clone the given repository to the given directory
-	Info("git clone %s %s --recursive", url, directory)
+	Info("git clone %s %s", url, directory)
 
 	r, err := git.PlainClone(directory, false, &git.CloneOptions{
-		URL:               url,
-		RecurseSubmodules: git.DefaultSubmoduleRecursionDepth,
+		URL: url,
 	})
 
 	CheckIfError(err)
--- a/_examples/checkout/main.go
+++ b/_examples/checkout/main.go
@@ -1,0 +1,36 @@
+package main
+
+import (
+	"fmt"
+	"os"
+
+	"gopkg.in/src-d/go-git.v4"
+	. "gopkg.in/src-d/go-git.v4/_examples"
+	"gopkg.in/src-d/go-git.v4/plumbing"
+)
+
+// Basic example of how to checkout a specific commit.
+func main() {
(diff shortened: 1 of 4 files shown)


_examples/common.go
```diff
@@ -6,13 +6,16 @@ import (
 	"strings"
 )
 
+// CheckArgs should be uesed to esnure the right command line arguments are
```
mcuadros: s/uesed/used/

_examples/common.go
```diff
@@ -6,13 +6,16 @@ import (
...
+// passed before executing an example.
 func CheckArgs(arg ...string) {
 	if len(os.Args) < len(arg)+1 {
 		Warning("Usage: %s %s", os.Args[0], strings.Join(arg, " "))
 		os.Exit(1)
 	}
 }
 
+// CheckIfError should be used to naiivly panics if an error is not nil.
```
mcuadros: s/naiivly/naively/

_examples/checkout/main.go
```diff
@@ -0,0 +1,36 @@
...
+	"gopkg.in/src-d/go-git.v4/plumbing"
+)
+
+// Basic example of how to checkout a specific commit.
+func main() {
+	CheckArgs("<url>", "<directory>", "<commit-ref>")
+	url, directory, commitRef := os.Args[1], os.Args[2], os.Args[3]
+
+	// Clone the given repository to the given directory
+	Info("git clone %s %s --recursive", url, directory)
```
mcuadros: why do a --recursive? This makes the example more complex

_examples/checkout/main.go
```diff
@@ -0,0 +1,36 @@
...
+
+	Info("git checkout %s", commitRef)
+
+	w, err := r.Worktree()
+
+	CheckIfError(err)
+
+	CheckIfError(w.Checkout(plumbing.NewHash(commitRef)))
+
+	fmt.Println("voila")
```
mcuadros: Maybe you can print the HEAD commit

marwan-at-work: @mcuadros, the problem as i stated in the pr description, is that the head commit will still be the old one, because go-git does not checkout the commit in a detached state. So you will actually still get the old HEAD commit, but the difference is that you will have the changes from the target-ref applied directly to your current branch . Therefore, printing the HEAD commit would be misleading. Let me know if there's a more proper way to do this. Thanks!

PR #333, opened 2017-04-06
Lazily load object index.

### LINKED ISSUE #327: Open scans object index even if not required

taralx:
fixes #327 

Now with passing tests.

### DIFF
--- a/storage/filesystem/object.go
+++ b/storage/filesystem/object.go
@@ -22,14 +22,18 @@ type ObjectStorage struct {
 
 func newObjectStorage(dir *dotgit.DotGit) (ObjectStorage, error) {
 	s := ObjectStorage{
-		dir:   dir,
-		index: make(map[plumbing.Hash]idx, 0),
+		dir: dir,
 	}
 
-	return s, s.loadIdxFiles()
+	return s, nil
 }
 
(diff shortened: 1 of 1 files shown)


storage/filesystem/object.go
```diff
@@ -22,43 +22,49 @@ type ObjectStorage struct {
...
+
+	s.index = make(map[plumbing.Hash]idx, 0)
 	packs, err := s.dir.ObjectPacks()
 	if err != nil {
 		return err
 	}
 
 	for _, h := range packs {
-		if err := s.loadIdxFile(h); err != nil {
+		idxfile, err := s.dir.ObjectPackIdx(h)
```
mcuadros: can we move this to a function?

taralx: Why? I moved it out of the function since it was a fairly straightforward piece of code...

mcuadros: because with this you can hide some complexity in this function, also this will help to simplify the code, putting the close in a defer, and inline err check.

taralx: Done.

PR #337, opened 2017-04-10
Add Repository.Log() method (fix #298)

### LINKED ISSUE #298: Implement Repository.Log method
This method was planned for this PR: https://github.com/src-d/go-git/pull/296, but it requires a lot of plumbing changes. To be able to do this we need:

- CommitIter should be an interface.
- Commit.History() method should returns CommitIter instead of slice of Commits.
- CommitWalker should be an Iterator.

ajnavarro:
- CommitIter is now an interface
- The old CommitIter implementation is now called StorerCommitIter
- CommitWalker and CommitWalkerPost are now iterators (CommitPreIterator and CommitPostIterator).
- Remove Commit.History() method. There are so many ways to iterate a commit history, depending of the use case. Now, instead of use the History() method, you must use CommitPreIterator or CommitPostIterator.
- Move commitSorterer to references.go because is the only place that it is used, and it must not be used into another place.
- Make References method private, it must only be used into blame logic.
- Added a TODO into references method, where the sortCommits is used to remove it in a near future.

### DIFF
--- a/plumbing/object/commit.go
+++ b/plumbing/object/commit.go
@@ -5,7 +5,6 @@ import (
 	"bytes"
 	"fmt"
 	"io"
-	"sort"
 	"strings"
 
 	"gopkg.in/src-d/go-git.v4/plumbing"
@@ -64,7 +63,7 @@ func (c *Commit) Tree() (*Tree, error) {
 }
 
 // Parents return a CommitIter to the parent Commits.
-func (c *Commit) Parents() *CommitIter {
+func (c *Commit) Parents() CommitIter {
 	return NewCommitIter(c.s,
 		storer.NewEncodedObjectLookupIter(c.s, plumbing.CommitObject, c.parents),
 	)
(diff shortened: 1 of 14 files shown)


plumbing/object/commit.go
```diff
@@ -231,24 +218,31 @@ func indent(t string) string {
...
-type CommitIter struct {
+// CommitIter is a generic closable interface for iterating over commits.
+type CommitIter interface {
+	Next() (*Commit, error)
+	ForEach(func(*Commit) error) error
+	Close()
+}
+
+// StorerCommitIter provides an iterator from commits in an EncodedObjectStorer.
+type StorerCommitIter struct {
```
mcuadros: this object can be private, since we are using through the interface

plumbing/object/commit.go
```diff
@@ -231,24 +218,31 @@ func indent(t string) string {
...
 
-// NewCommitIter takes a storer.EncodedObjectStorer and a
+// NewStorerCommitIter takes a storer.EncodedObjectStorer and a
 // storer.EncodedObjectIter and returns a *CommitIter that iterates over all
 // commits contained in the storer.EncodedObjectIter.
 //
 // Any non-commit object returned by the storer.EncodedObjectIter is skipped.
-func NewCommitIter(s storer.EncodedObjectStorer, iter storer.EncodedObjectIter) *CommitIter {
-	return &CommitIter{iter, s}
+func NewStorerCommitIter(s storer.EncodedObjectStorer, iter storer.EncodedObjectIter) CommitIter {
```
mcuadros: If the implementation is private the method can be calle `NewCommitIter`

plumbing/object/commit.go
```diff
@@ -271,6 +265,10 @@ func (iter *CommitIter) ForEach(cb func(*Commit) error) error {
 	})
 }
 
+func (iter *StorerCommitIter) Close() {
```
mcuadros: not tested

plumbing/object/commit.go
```diff
@@ -292,9 +290,3 @@ func SortCommits(l []*Commit) {
 	s := &commitSorterer{l}
```
mcuadros: not tested

Issue #256, opened 2017-02-06
add and commit support


dtmistry:
Thank you for this awesome library!! I apologize if this question is already answered. 

I see there is support for push using go-git. Is it in the roadmap to support other porcelain commands like 'add' or 'push'?

Also, is there current support to update index or write tree? 

I'm trying to use go-git with the memory storer in a tool that updates an existing repository.

thefallentree:
+1 ,  to do what git add/commit does it pretty heavy-lifting,  I think we can start with a simple CreateCommitWithBlob() sort of thing on Repository struct first.

nobuhito:
👍

Skilgarriff:
I would really love this feature as it would allow me to bundle my application with no external dependencies. Anything I can do to help here? I have some time this weekend, and would love to crank some of this out.

mcuadros:
@Skilgarriff I invite you to take a look to the internals, but is not one weekend task.

The main thing required to start, is implement `Worktree.Status` (`git status`) based on [merkletrie](https://github.com/src-d/go-git/tree/master/utils/merkletrie) to compare the file system tree, staging area and HEAD

Having a proper status, will be easy implement `git add` and `git commit`, those methods aren't more than wiring between other components

mcuadros:
@Skilgarriff I am doing myself the implementation of status. I hope you didn't started

Skilgarriff:
@mcuadros I have not started yet, I was planning on checking it out today. Can I help at all? Or will you ping here when that is completed and we could begin looking at commit and add

mcuadros:
I am opening a PR a soon as I have something, to let you see what is being implemented

You can work in the `add` now if you want, take a look how create objects, is not very evolved and maybe you will need to do some PRs

PR #339, opened 2017-04-11
worktree, status and reset implementation based on merkletrie

mcuadros:
This is a implementation of the status and the reset implementation based on merkletrie.

Things to be agreed:
- Two new implementations of `merkletrie.Noder` have been implemented, those implementations are inside of `utils/merkletrie/<use-case>`, and the original one is at `plumbing/objects/`, worth to put all the implementations in the same place?

- Due to the lack of error in the `noder.Hash` interface method, the [implementation](https://github.com/src-d/go-git/pull/339/files#diff-c7259dcedc0d9f9f14a75f59997c622fR50) contains panic, We should add an error to this method?

NOTE: hard reset is not implemented, I am finishing this and some additional `Worktree.Reset` tests tomorrow

Required by https://github.com/src-d/go-git/issues/256

### DIFF
--- a/options.go
+++ b/options.go
@@ -177,3 +177,55 @@ type SubmoduleUpdateOptions struct {
 	// submodules (and so on). Until the SubmoduleRescursivity is reached.
 	RecurseSubmodules SubmoduleRescursivity
 }
+
+// CheckoutOptions describes how a checkout operation should be performed.
+type CheckoutOptions struct {
+	// Branch to be checked out, if empty uses `master`
+	Branch plumbing.ReferenceName
+	Hash   plumbing.Hash
+	// RemoteName is the name of the remote to be pushed to.
+	Force bool
+}
+
+// Validate validates the fields and sets the default values.
+func (o *CheckoutOptions) Validate() error {
+	if o.Branch == "" {
+		o.Branch = plumbing.Master
+	}
+
+	return nil
+}
+
+type ResetMode int
+
+const (
+	// HardReset resets the index and working tree. Any changes to tracked files
+	// in the working tree are discarded.
+	HardReset ResetMode = iota
+	// MixedReset Resets the index but not the working tree (i.e., the changed
+	// files are preserved but not marked for commit) and reports what has not
+	// been updated. This is the default action.
+	MixedReset
+)
+
+// ResetOptions describes how a reset operation should be performed.
+type ResetOptions struct {
+	// Commit, if commit is pressent set the current branch head (HEAD) to it.
+	Commit plumbing.Hash
+	// Mode
+	Mode ResetMode
+}
+
+// Validate validates the fields and sets the default values.
+func (o *ResetOptions) Validate(r *Repository) error {
+	if o.Commit == plumbing.ZeroHash {
+		ref, err := r.Head()
+		if err != nil {
+			return err
+		}
+
+		o.Commit = ref.Hash()
+	}
+
+	return nil
+}
--- a/options.go
+++ b/options.go
@@ -178,12 +178,15 @@ type SubmoduleUpdateOptions struct {
 	RecurseSubmodules SubmoduleRescursivity
 }
 
-// CheckoutOptions describes how a checkout operation should be performed.
+// CheckoutOptions describes how a checkout 31operation should be performed.
 type CheckoutOptions struct {
-	// Branch to be checked out, if empty uses `master`
+	// Hash to be checked out, if used HEAD will in detached mode. Branch and
+	// Hash are mutual exclusive.
+	Hash plumbing.Hash
+	// Branch to be checked out, if Branch and Hash are empty is set to `master`.
 	Branch plumbing.ReferenceName
-	Hash   plumbing.Hash
-	// RemoteName is the name of the remote to be pushed to.
+	// Force, if true when switching branches, proceed even if the index or the
+	// working tree differs from HEAD. This is used to throw away local changes
 	Force bool
 }
 
@@ -196,23 +199,34 @@ func (o *CheckoutOptions) Validate() error {
 	return nil
 }
 
-type ResetMode int
+// ResetMode defines the mode of a reset operation.
+type ResetMode int8
 
 const (
 	// HardReset resets the index and working tree. Any changes to tracked files
 	// in the working tree are discarded.
 	HardReset ResetMode = iota
-	// MixedReset Resets the index but not the working tree (i.e., the changed
+	// MixedReset resets the index but not the working tree (i.e., the changed
 	// files are preserved but not marked for commit) and reports what has not
 	// been updated. This is the default action.
 	MixedReset
+	// MergeReset resets the index and updates the files in the working tree
+	// that are different between Commit and HEAD, but keeps those which are
+	// different between the index and working tree (i.e. which have changes
+	// which have not been added).
+	//
+	// If a file that is different between Commit and the index has unstaged
+	// changes, reset is aborted.
+	MergeReset
 )
 
 // ResetOptions describes how a reset operation should be performed.
 type ResetOptions struct {
 	// Commit, if commit is pressent set the current branch head (HEAD) to it.
 	Commit plumbing.Hash
-	// Mode
+	// Mode, form resets the current branch head to Commit and possibly updates
+	// the index (resetting it to the tree of Commit) and the working tree
+	// depending on Mode. If empty MixedReset is used.
 	Mode ResetMode
 }
 
--- a/plumbing/object/treenoder.go
+++ b/plumbing/object/treenoder.go
@@ -21,10 +21,11 @@ type treeNoder struct {
 	name     string // empty string for the root node
 	mode     filemode.FileMode
 	hash     plumbing.Hash
-	children []noder.Noder // memoized
+	children []noder.Noder // memorized
 }
 
-func newTreeNoder(t *Tree) *treeNoder {
+// NewTreeRootNode returns the root node of a Tree
+func NewTreeRootNode(t *Tree) *treeNoder {
 	if t == nil {
 		return &treeNoder{}
 	}
@@ -74,7 +75,7 @@ func (t *treeNoder) Children() ([]noder.Noder, error) {
 		return noder.NoChildren, nil
 	}
 
-	// children are memoized for efficiency
+	// children are memorized for efficiency
 	if t.children != nil {
 		return t.children, nil
 	}
--- a/plumbing/object/treenoder.go
+++ b/plumbing/object/treenoder.go
@@ -1,13 +1,5 @@
 package object
 
-// A treenoder is a helper type that wraps git trees into merkletrie
-// noders.
-//
-// As a merkletrie noder doesn't understand the concept of modes (e.g.
-// file permissions), the treenoder includes the mode of the git tree in
-// the hash, so changes in the modes will be detected as modifications
-// to the file contents by the merkletrie difftree algorithm.  This is
-// consistent with how the "git diff-tree" command works.
 import (
 	"io"
 
@@ -16,16 +8,24 @@ import (
 	"gopkg.in/src-d/go-git.v4/utils/merkletrie/noder"
 )
 
+// A treenoder is a helper type that wraps git trees into merkletrie
+// noders.
+//
+// As a merkletrie noder doesn't understand the concept of modes (e.g.
+// file permissions), the treenoder includes the mode of the git tree in
+// the hash, so changes in the modes will be detected as modifications
+// to the file contents by the merkletrie difftree algorithm.  This is
+// consistent with how the "git diff-tree" command works.
 type treeNoder struct {
 	parent   *Tree  // the root node is its own parent
 	name     string // empty string for the root node
 	mode     filemode.FileMode
 	hash     plumbing.Hash
-	children []noder.Noder // memorized
+	children []noder.Noder // memoized
 }
 
 // NewTreeRootNode returns the root node of a Tree
-func NewTreeRootNode(t *Tree) *treeNoder {
+func NewTreeRootNode(t *Tree) noder.Noder {
 	if t == nil {
 		return &treeNoder{}
 	}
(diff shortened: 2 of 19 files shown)


alcortesm:
> Two new implementations of merkletrie.Noder has been implemented, those implementations are ?
> inside of utils/merkletrie/<use-case>, and the original one is at plumbing/objects/, worth to put all the > implementations in the same place?

I don't think putting all the implementations in the same place is worth it.

> Due to the lack of error in the noder.Hash interface method, the implementation contains panic, We should add an error to this method?

I don't think so.

@smola and I talked about it a few months ago when I was working on merkletrie.  We agreed that the hash method should not return errors, as valid objects should have valid hashes, while objects, with a  hash that cannot be calculated, should return errors upon construction.

Let us know if you would rather do it any other way.

plumbing/object/treenoder.go
```diff
@@ -21,10 +21,11 @@ type treeNoder struct {
 	name     string // empty string for the root node
 	mode     filemode.FileMode
 	hash     plumbing.Hash
-	children []noder.Noder // memoized
+	children []noder.Noder // memorized
```
smola: memoized was actually the right term ;-)

plumbing/object/treenoder.go
```diff
@@ -21,10 +21,11 @@ type treeNoder struct {
...
 }
 
-func newTreeNoder(t *Tree) *treeNoder {
+// NewTreeRootNode returns the root node of a Tree
+func NewTreeRootNode(t *Tree) *treeNoder {
```
smola: Shouldn't this be just `NewTreeNode`? The `Root` part seems redundant?

plumbing/object/treenoder.go
```diff
@@ -74,7 +75,7 @@ func (t *treeNoder) Children() ([]noder.Noder, error) {
 		return noder.NoChildren, nil
 	}
 
-	// children are memoized for efficiency
+	// children are memorized for efficiency
```
smola: s/memorized/memoized/

utils/merkletrie/filesystem/node.go
```diff
@@ -0,0 +1,128 @@
...
+
+	return &Node{fs: fs, info: info, isDir: true, name: ""}, nil
+}
+
+func (n *Node) String() string {
+	return filepath.Join(n.parent, n.name)
+}
+
+func (n *Node) Hash() []byte {
+	if n.IsDir() {
```
smola: Shouldn't Hash() be cached? I think DiffTree does a lot of repeated calls to Hash. So either implementations cache Hash, or DiffTree caches them internally.

options.go
```diff
@@ -177,3 +177,55 @@ type SubmoduleUpdateOptions struct {
 	// submodules (and so on). Until the SubmoduleRescursivity is reached.
 	RecurseSubmodules SubmoduleRescursivity
 }
+
+// CheckoutOptions describes how a checkout operation should be performed.
+type CheckoutOptions struct {
+	// Branch to be checked out, if empty uses `master`
+	Branch plumbing.ReferenceName
+	Hash   plumbing.Hash
+	// RemoteName is the name of the remote to be pushed to.
```
smola: This is godoc for RemoteName and the variable is Force, copypasta?

status.go
```diff
@@ -0,0 +1,92 @@
...
+	Untracked
+	Modified
+	Added
+	Deleted
+	Renamed
+	Copied
+	UpdatedButUnmerged
+)
+
+func (c StatusCode) String() string {
```
alcortesm: I think it will be simpler if declare StatusCode as a rune or a string instead of an int8?

utils/merkletrie/filesystem/node.go
```diff
@@ -0,0 +1,128 @@
...
+	pathB := b.(noder.Path)
+	if pathA[len(pathA)-1].IsDir() || pathB[len(pathB)-1].IsDir() {
+		return false
+	}
+
+	return bytes.Equal(a.Hash(), b.Hash())
+}
+
+type Node struct {
+	parent string
```
alcortesm: Is this the name of the parent or its path?

utils/merkletrie/filesystem/node.go
```diff
@@ -0,0 +1,128 @@
...
+	fs     billy.Filesystem
+}
+
+func NewRootNode(fs billy.Filesystem) (*Node, error) {
+	info, err := fs.Stat("/")
+	if err != nil && !os.IsNotExist(err) {
+		return nil, err
+	}
+
+	return &Node{fs: fs, info: info, isDir: true, name: ""}, nil
```
alcortesm: if you set name to the empty string, I think you should also set parent, for consistency.

utils/merkletrie/filesystem/node.go
alcortesm: Would it make sense for parent to be a Node?

utils/merkletrie/filesystem/node.go
```diff
@@ -0,0 +1,128 @@
...
+
+	return bytes.Equal(a.Hash(), b.Hash())
+}
+
+type Node struct {
+	parent string
+	name   string
+	isDir  bool
+	info   billy.FileInfo
+	fs     billy.Filesystem
```
alcortesm: we should probably add a hash attribute also, instead of calculating it on demand.

utils/merkletrie/filesystem/node.go
```diff
@@ -0,0 +1,128 @@
...
+	}
+
+	return &Node{fs: fs, info: info, isDir: true, name: ""}, nil
+}
+
+func (n *Node) String() string {
+	return filepath.Join(n.parent, n.name)
+}
+
+func (n *Node) Hash() []byte {
```
alcortesm: You should document how are you calculating the hash, especially why the hash of a dir is nil.

utils/merkletrie/filesystem/node.go
```diff
@@ -0,0 +1,128 @@
...
+	"gopkg.in/src-d/go-git.v4/plumbing"
+	"gopkg.in/src-d/go-git.v4/plumbing/filemode"
+	"gopkg.in/src-d/go-git.v4/utils/merkletrie/noder"
+)
+
+var ignore = map[string]bool{
+	".git": true,
+}
+
+func IsEquals(a, b noder.Hasher) bool {
```
alcortesm: This should be documented extensively, including what it will be the algorithm for comparing hashes in directories.

utils/merkletrie/filesystem/node.go
```diff
@@ -0,0 +1,128 @@
...
+			isDir:  file.IsDir(),
+		})
+	}
+
+	return c, nil
+}
+
+func (n *Node) NumChildren() (int, error) {
+	files, err := n.readDir()
+	return len(files), err
```
alcortesm: Not sure about this, be careful with submodules and empty dirs.

utils/merkletrie/filesystem/node.go
```diff
@@ -0,0 +1,128 @@
+package filesystem
```
alcortesm: this should probably be moved to the billy package, but definately it should not be under merkletrie utilities.

utils/merkletrie/index/node.go
```diff
@@ -0,0 +1,113 @@
+package index
```
alcortesm: this should probably no go under the merkletrie package, but on the index.

Please document extensively how are you planning to map an index to a merkletrie, the challenges, surprises and your solutions. Otherwise, I can only guess your intentions and it is very time consuming and error prone.

utils/merkletrie/filesystem/node.go
```diff
@@ -0,0 +1,128 @@
...
+)
+
+var ignore = map[string]bool{
+	".git": true,
+}
+
+func IsEquals(a, b noder.Hasher) bool {
+	pathA := a.(noder.Path)
+	pathB := b.(noder.Path)
+	if pathA[len(pathA)-1].IsDir() || pathB[len(pathB)-1].IsDir() {
```
alcortesm: I don't understand this, reviewing this is too time consuming due to lack of documentation.

Can you can explain the general guidelines of how are you planning to map a filesystem into a merkletrie at the very begining?, also the main challenges, surprises and your solutions.

alcortesm:
I will continue reviewing this later, it is too long.

utils/merkletrie/filesystem/node.go
mcuadros: doesn't make any sense at billy, is a very specific things to git, maybe outside of the merkletrie package.

utils/merkletrie/filesystem/node.go
mcuadros: is the parent of the fs, doesn't make any sense to me document an internal property

utils/merkletrie/filesystem/node.go
mcuadros: sure

utils/merkletrie/filesystem/node.go
mcuadros: leftover

utils/merkletrie/filesystem/node.go
mcuadros: the hash calculation is universal to all the Noder implementations in go-git, otherwise you can compare it. That's why I believe that all the implementations should be in the same place.

@smola

utils/merkletrie/filesystem/node.go
mcuadros: done, but I could see any performance improvement.

utils/merkletrie/filesystem/node.go
mcuadros: done, since now the children are pre-computed, return just the len of the children slice.

The submodule are not tested, and yes for sure are failing since, the hash of a submodule is not computed in the same way. I am implementing this.

utils/merkletrie/index/node.go
mcuadros: You can't place this in index, you will end with circular dependencies.

mcuadros:
@alcortesm with this goal I created nice commits, to allow you review commit by commit if you wish.

options.go
```diff
@@ -177,3 +177,55 @@ type SubmoduleUpdateOptions struct {
...
+type ResetMode int
+
+const (
+	// HardReset resets the index and working tree. Any changes to tracked files
+	// in the working tree are discarded.
+	HardReset ResetMode = iota
+	// MixedReset Resets the index but not the working tree (i.e., the changed
+	// files are preserved but not marked for commit) and reports what has not
+	// been updated. This is the default action.
+	MixedReset
```
mcuadros: `ResetMode` has 5 posible values

options.go
```diff
@@ -177,3 +177,55 @@ type SubmoduleUpdateOptions struct {
...
+type CheckoutOptions struct {
+	// Branch to be checked out, if empty uses `master`
+	Branch plumbing.ReferenceName
+	Hash   plumbing.Hash
+	// RemoteName is the name of the remote to be pushed to.
+	Force bool
+}
+
+// Validate validates the fields and sets the default values.
+func (o *CheckoutOptions) Validate() error {
```
mcuadros: This was already discussed on previous PRs, and is not related to this PR

status.go
```diff
@@ -0,0 +1,92 @@
+package git
+
+import "fmt"
+
+// Status current status of a Worktree
+type Status map[string]*FileStatus
+
+func (s Status) File(filename string) *FileStatus {
```
mcuadros: You were documented moved code. This doesn't bellow to this PR

utils/merkletrie/filesystem/node.go
```diff
@@ -0,0 +1,144 @@
...
+
+	return &node{
+		fs:    n.fs,
+		path:  path,
+		hash:  hash,
+		isDir: file.IsDir(),
+	}, nil
+}
+
+func (n *node) calculateHash(path string, file billy.FileInfo) ([]byte, error) {
```
alcortesm: On line 71, the method `calculateChildren` sets the children of the node and only returns an error.

Here, the method `calculateHash` doesn't set the hash but return it instead.

I think we should use the same strategy in both methods, for coherency, or just make this a function instead of a method.

utils/merkletrie/filesystem/node.go
```diff
@@ -0,0 +1,144 @@
...
+func NewRootNode(fs billy.Filesystem) noder.Noder {
+	return &node{fs: fs, isDir: true}
+}
+
+// Hash the hash of a filesystem is the result of concatenating the computed
+// plumbing.Hash of the file as a Blob and its plumbing.FileMode; that way the
+// difftree algorithm will detect changes in the contents of files and also in
+// their mode.
+//
+// The hash of a directory is always a 24-bytes slice of zero values
```
alcortesm: Also, the root node has a nil hash, instead of a slice of 24 zeroes.  Maybe we should comment this especial case here.

utils/merkletrie/filesystem/node_test.go
```diff
@@ -0,0 +1,114 @@
...
+	. "gopkg.in/check.v1"
+	"gopkg.in/src-d/go-billy.v2"
+	"gopkg.in/src-d/go-billy.v2/memfs"
+	"gopkg.in/src-d/go-git.v4/utils/merkletrie"
+	"gopkg.in/src-d/go-git.v4/utils/merkletrie/noder"
+)
+
+func Test(t *testing.T) { TestingT(t) }
+
+type NoderSuite struct{}
```
alcortesm: I think it is better to call it `NodeSuite` instead.

utils/merkletrie/filesystem/node_test.go
```diff
@@ -0,0 +1,114 @@
...
+		err = err1
+	}
+	return err
+}
+
+var empty = make([]byte, 24)
+
+func IsEquals(a, b noder.Hasher) bool {
+	if bytes.Equal(a.Hash(), empty) || bytes.Equal(b.Hash(), empty) {
+		return false
```
alcortesm: This is very weird, as it is defeating the whole purpose of using the merkletrie package and implementing Noder.

Maybe it is ok for the tests, but in that case I would mention it in a comment.

utils/merkletrie/filesystem/node_test.go
```diff
@@ -0,0 +1,114 @@
...
+	"gopkg.in/src-d/go-git.v4/utils/merkletrie/noder"
+)
+
+func Test(t *testing.T) { TestingT(t) }
+
+type NoderSuite struct{}
+
+var _ = Suite(&NoderSuite{})
+
+func (s *NoderSuite) TestDiff(c *C) {
```
alcortesm: I think this way of testing `node` will make debugging harder than if we do unit testing of its public methods instead: it would be easier to detect errors with the unit tests than having to trace back difftree's results mismatches in the tests to see what errors in `node` caused them.

I would write unit tests for `node` and maybe left a cople of these as integrations tests.

utils/merkletrie/index/node.go
alcortesm: I would probably place it in an internal package under the index package.

alcortesm:
I am exhausted, I will continue reviewing this the next day.

alcortesm:
I think you merged before my review, you should consider reopening for my comments today.

I will probably a few more in the next days.

PR #346, opened 2017-04-16
transport: ssh, default HostKeyCallback

mcuadros:
Solves https://github.com/src-d/go-git/issues/328

Introduce a helper function `NewKnownHostsCallback` that reads a `knownhosts` files. This method is used by default in any `ssh.AuthMethod` reading the env variable `SSH_KNOWN_HOSTS` if empty a `~/.ssh/known_hosts` or/and `/etc/ssh/ssh_known_hosts`.

It uses `golang.org/x/crypto/ssh/knownhosts` as suggested @peefourtee.

This PR cannot be merged until this https://go-review.googlesource.com/#/c/40911/ get merge or this issue solved https://github.com/golang/go/issues/19994

Related https://github.com/src-d/go-git/pull/329

Thanks to @adampointer @Xfennec

### DIFF
--- a/plumbing/transport/ssh/auth_method.go
+++ b/plumbing/transport/ssh/auth_method.go
@@ -5,9 +5,12 @@ import (
 	"fmt"
 	"net"
 	"os"
+	"os/user"
+	"path/filepath"
 
 	"golang.org/x/crypto/ssh"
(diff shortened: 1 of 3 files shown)


kris-nova:
Is there an example of how to use this floating around anywhere?

Issue #342, opened 2017-04-12
SSH PublicKeys authentication requires importing two `ssh` packages


jediorange:
In order to authenticate to a repo via ssh using public/private keys, one must import both the `golang.org/x/crypto/ssh` and the `gopkg.in/src-d/go-git.v4/plumbing/transport/ssh`, which are both named `ssh` by default.

```go
package main

import (
	"os"

	"golang.org/x/crypto/ssh"
	git "gopkg.in/src-d/go-git.v4"
	gitssh "gopkg.in/src-d/go-git.v4/plumbing/transport/ssh"
)

func main() {
	sshPrivateKey := os.Getenv("SSH_KEY")
	signer, _ := ssh.ParsePrivateKey([]byte(sshPrivateKey))
	git.PlainClone("example", false, &git.CloneOptions{
		URL:      "ssh://github.com/example/example.git",
		Auth:     &gitssh.PublicKeys{User: "git", Signer: signer},
		Progress: os.Stdout,
	})
}
```

This necessitates renaming one of the imports, and is a bit confusing.  A more ideal situation would be to provide a function within the `git` package to create a new auth from a name/private key, such as:

```go
func NewSSHPublicKeyAuth(user string, pemPrivateKey []byte) (*PublicKeys, error)
```

An even better solution would be to include "New" auth functions (HTTP, SSH) in the main `git` package, so that sub-packages (`plumbing`) do not need to be imported for basic authentication functionality.

icco:
This really confused me. I was really surprised 

```
	git.PlainClone("example", false, &git.CloneOptions{
		URL:      "ssh://github.com/example/example.git",
	})
```

didn't just work.

jdoklovic:
Why not:
`Auth: ssh.NewSSHAgentAuth("git")`

jediorange:
@jdoklovic I can't rely on a good key being installed on my server. 

That may work while running on somebody's personal machine, or on a persistent server, but when deployed to an ephemeral container, or following 12factor best practices, I can't rely on the ssh key being installed on the server.

amendgit:
what I found in https://github.com/golang/crypto/blob/master/ssh/example_test.go
I'm not quite sure, does this need to consider the decryption.
```
        // A public key may be used to authenticate against the remote
	// server by using an unencrypted PEM-encoded private key file.
	//
	// If you have an encrypted private key, the crypto/x509 package
	// can be used to decrypt it.
[...]

PR #347, opened 2017-04-18
transport: ssh, NewPublicKeys helper

mcuadros:
Fixes https://github.com/src-d/go-git/issues/342

### DIFF
--- a/plumbing/transport/ssh/auth_method.go
+++ b/plumbing/transport/ssh/auth_method.go
@@ -3,6 +3,7 @@ package ssh
 import (
 	"errors"
 	"fmt"
+	"io/ioutil"
 	"net"
 	"os"
 	"os/user"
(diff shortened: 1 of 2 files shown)


plumbing/transport/ssh/auth_method.go
```diff
@@ -13,6 +14,8 @@ import (
 	"golang.org/x/crypto/ssh/agent"
 )
 
+const DefaultSSHUsername = "git"
```
smola: Maybe just `DefaultUsername`? It's already `ssh` package.

jdoklovic:
I think this is a good start but it should be even simpler... it should use sane defaults just like any other ssh/git tool. e.g. there should be a simple way for discovering ~/.ssh/id_rsa automatically.  Also, there should be a "wrapper" of sorts that can figure out that ssh-agent either is or isn't running and if not fall back to using the default ~/.ssh dir.  There should always be a way to force the key though as this PR already contains.

PR #353, opened 2017-04-20
added credentials provider for basic auth

### LINKED ISSUE #350: AuthMethod in http/common.go needs to export setAuth
right now the setAuth is not exported. It should be SetAuth. Otherwise, there's no way for a consumer to apply basic auth (e.g. ask for user/pass) when it's needed and instead they need to do it when the Auth is set on the Config which means you may ask for a user/pass before you know if an endpoint actually reuires auth.

Making the SetAuth exported is a quick fix so consumers can just implement the interface instead of using the built-in BasicAuth.

A better route would probably be to add a callback for grabbing the user/pass.

jdoklovic:
Fixes #350

### DIFF
--- a/plumbing/transport/http/common.go
+++ b/plumbing/transport/http/common.go
@@ -116,8 +116,11 @@ func (a *BasicAuth) setAuth(r *http.Request) {
 		return
 	}
 
-	u, p := a.CredentialsProvider()
-	r.SetBasicAuth(u, p)
+	if len(a.username) < 1 || len(a.password) < 1 {
+		a.username, a.password = a.CredentialsProvider()
+	}
+
+	r.SetBasicAuth(a.username, a.password)
 }
 
 // Name is name of the auth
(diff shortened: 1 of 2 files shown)


bzz:
Number of CI failures on Win might be related to #369 

Ping @smola @alcortesm @mcuadros for a review

jdoklovic:
I'm not sure how those failures got introduced unless windows tests were failing on master when i branched.
My changes are obviously very small and don't have anything to do with the failures listed

jdoklovic:
@smola @mcuadros any suggestions on how to fix this up and get it merged? I need this change for a CLI tool i'm building...

smola:
@bzz Please, just ignore Windows failures, except for PR fixing stuff on Windows.

smola:
Closing. See: https://github.com/src-d/go-git/issues/350#issuecomment-300096008

PR #355, opened 2017-04-21
transport: ssh, NewPublicKeys support for encrypted PEM files

mcuadros:
In order to avoid code repetitions from our user dealing with encrypted PEM files, and since is not trivial, this PR adds the support to NewPublicKeys to deal with encrypted PEM files

### DIFF
--- a/plumbing/transport/ssh/auth_method_test.go
+++ b/plumbing/transport/ssh/auth_method_test.go
@@ -109,7 +109,14 @@ func (s *SuiteCommon) TestNewSSHAgentAuth(c *C) {
 }
 
 func (*SuiteCommon) TestNewPublicKeys(c *C) {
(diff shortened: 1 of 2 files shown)


plumbing/transport/ssh/auth_method_test.go
```diff
@@ -109,7 +109,14 @@ func (s *SuiteCommon) TestNewSSHAgentAuth(c *C) {
 }
 
 func (*SuiteCommon) TestNewPublicKeys(c *C) {
-	auth, err := NewPublicKeys("foo", testdata.PEMBytes["rsa"])
+	auth, err := NewPublicKeys("foo", testdata.PEMBytes["rsa"], "")
```
ajnavarro: I miss some test with password non empty.

alcortesm:
LGTM, nice comments by the way.

PR #361, opened 2017-04-26
worktree: add method

mcuadros:
This is the bare minimum implementation of `git add`

### DIFF
--- a/worktree_status.go
+++ b/worktree_status.go
@@ -2,16 +2,20 @@ package git
 
 import (
 	"bytes"
+	"io"
 
 	"gopkg.in/src-d/go-git.v4/plumbing"
+	"gopkg.in/src-d/go-git.v4/plumbing/filemode"
+	"gopkg.in/src-d/go-git.v4/plumbing/format/index"
 	"gopkg.in/src-d/go-git.v4/plumbing/object"
+	"gopkg.in/src-d/go-git.v4/utils/ioutil"
 	"gopkg.in/src-d/go-git.v4/utils/merkletrie"
 	"gopkg.in/src-d/go-git.v4/utils/merkletrie/filesystem"
-	"gopkg.in/src-d/go-git.v4/utils/merkletrie/index"
+	mindex "gopkg.in/src-d/go-git.v4/utils/merkletrie/index"
 	"gopkg.in/src-d/go-git.v4/utils/merkletrie/noder"
 )
 
-// Status returns the working tree status
+// Status returns the working tree status.
 func (w *Worktree) Status() (Status, error) {
 	ref, err := w.r.Head()
 	if err == plumbing.ErrReferenceNotFound {
@@ -80,7 +84,7 @@ func (w *Worktree) diffStagingWithWorktree() (merkletrie.Changes, error) {
 		return nil, err
 	}
 
-	from := index.NewRootNode(idx)
+	from := mindex.NewRootNode(idx)
 	submodules, err := w.getSubmodulesStatus()
 	if err != nil {
 		return nil, err
@@ -131,7 +135,7 @@ func (w *Worktree) diffCommitWithStaging(commit plumbing.Hash, reverse bool) (me
 		return nil, err
 	}
 
-	to := index.NewRootNode(idx)
+	to := mindex.NewRootNode(idx)
 	from := object.NewTreeRootNode(t)
 
 	if reverse {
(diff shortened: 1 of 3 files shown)


worktree_status.go
```diff
@@ -159,3 +163,103 @@ func diffTreeIsEquals(a, b noder.Hasher) bool {
...
 }
+
+// Add adds the file contents of a file in the worktree to the index.
+func (w *Worktree) Add(path string) (plumbing.Hash, error) {
+	s, err := w.Status()
+	if err != nil {
+		return plumbing.ZeroHash, err
+	}
+
+	if s.File(path).Worktree == Unmodified {
```
smola: I wonder if returning a zero hash when the operation is successful but unmodified is the right behavior. If it is, please, explain it in godoc, it is not that intuitive.

worktree_status.go
```diff
@@ -159,3 +163,103 @@ func diffTreeIsEquals(a, b noder.Hasher) bool {
...
+	}
+
+	defer ioutil.CheckClose(f, &err)
+
+	h := plumbing.NewHasher(plumbing.BlobObject, fi.Size())
+	if _, err := io.Copy(h, f); err != nil {
+		return plumbing.ZeroHash, err
+	}
+
+	return h.Sum(), nil
```
smola: `return h.Sum(), err` to get close error?

worktree_test.go
```diff
@@ -5,6 +5,7 @@ import (
 	"os"
 	"path/filepath"
 
+	billy "gopkg.in/src-d/go-billy.v2"
```
smola: mixed external/internal imports

worktree_status.go
alcortesm: Better yet, declare error in the signature and defer a closure on it instead.

worktree_status.go
```diff
@@ -159,3 +163,103 @@ func diffTreeIsEquals(a, b noder.Hasher) bool {
...
+
+	err = w.doUpdateFileToIndex(idx, filename, h)
+	if err != nil {
+		return err
+	}
+
+	return w.r.Storer.SetIndex(idx)
+}
+
+func (w *Worktree) doAddFileToIndex(idx *index.Index, filename string) error {
```
alcortesm: change the name of this method to `addFileToIndex` and make it a function, instead of a method.

worktree_status.go
```diff
@@ -159,3 +163,103 @@ func diffTreeIsEquals(a, b noder.Hasher) bool {
...
+	idx.Entries = append(idx.Entries, index.Entry{
+		Name: filename,
+	})
+
+	return nil
+}
+
+func (w *Worktree) doUpdateFileToIndex(idx *index.Index, filename string, h plumbing.Hash) error {
```
alcortesm: Change the name of the method to `updateFileToIndex`.

worktree_status.go
```diff
@@ -159,3 +163,103 @@ func diffTreeIsEquals(a, b noder.Hasher) bool {
...
+	info, err := w.fs.Stat(filename)
+	if err != nil {
+		return err
+	}
+
+	for i, e := range idx.Entries {
```
alcortesm: Can we use the Entry method here instead of looking for it again? (like we did in line 217)

mcuadros: nope, I need to know the position for overwritten it.

worktree_status.go
mcuadros: it need context, and since is a very deep method I prefer having the `do`  prefix

worktree_status.go
mcuadros: could be a function, but since update cannot be a function I rather keep it as a method

worktree_status.go
alcortesm: What does `do` mean for you?

worktree_status.go
alcortesm: That is part of the problem with this PR, it is too C style, everything changing other things that are sometimes several levels of indirection away.

I think you should refactor and modify the index and the entries, adding the necessary methods there util the functionality here is trivial and it is performed by the entities responsible for it instead of doing everything here.

mcuadros: I don't if worth do this change, and change the other places where the index is read. When he is the only place that an entry is updated.

worktree_status.go
mcuadros: that is performing the action with a function is a name close to this one without the do

mcuadros:
merging this making other PR change the Index implementation

PR #363, opened 2017-04-26
add support for .git as file, fixes #348

### LINKED ISSUE #348: Broken on submodules since .git is not a directory
Happens with https://github.com/mvdan/git-picked, which does `git.PlainOpen(".")` and reads some info about the branches. This software works fine when the clone isn't a submodule.

But when it is...

```
 $ git-picked
open .git/objects/pack: not a directory
```

More info:

```
 $ ls -la .git
-rw-r--r-- 1 mvdan mvdan 34 Mar 27 16:58 .git
 $ cat .git
gitdir: ../.git/modules/webclient
```

smola:


smola:
Given current `billy` assumptions, this only supports relative paths in the `.git` file. Are absolute paths actually used in practice? Can we add an error when the path is not relative? Otherwise, we probably need significative changes.

repository.go
```diff
@@ -196,15 +199,25 @@ func PlainOpen(path string) (*Repository, error) {
...
+	if err != nil {
 		if !os.IsNotExist(err) {
 			return nil, err
 		}
 
 		dot = fs
 	} else {
 		wt = fs
-		dot = fs.Dir(".git")
+		if fi.IsDir() {
```
mcuadros: a very nested logic, consider move this to a function

repository.go
```diff
@@ -196,15 +199,25 @@ func PlainOpen(path string) (*Repository, error) {
...
+			dot = fs.Dir(".git")
+		} else {
+			dot, err = dotGitFileToFilesystem(fs)
+			if err != nil {
+				return nil, err
+			}
+
+			fmt.Println("BASE", dot.Base())
```
alcortesm: leftover?

repository.go
```diff
@@ -196,15 +199,25 @@ func PlainOpen(path string) (*Repository, error) {
 	var wt, dot billy.Filesystem
 
 	fs := osfs.New(path)
-	if _, err := fs.Stat(".git"); err != nil {
+	fi, err := fs.Stat(".git")
+	if err != nil {
```
alcortesm: I think the logic here is correct, but we can improve its readability by hiding the low-level details: add a switch with few functions, like `isWorkingDirectory`, `isSubModule`, `isBare`... that reveal our intentions to the reader, hiding the details, that are sometimes unkown to the reader and much more difficult to get right).

This will also fix @mcuadros comment as a bonus.

alcortesm:
> Are absolute paths actually used in practice? Can we add an error when the path is not relative?

The `git submodule add` command allows absolute paths, but the documentation does not mention if absolute paths are allowed in the `.git` file.

To learn if absolute paths are common I propose to run a query in our database over all our cloned repos.

repository.go
smola: done

repository.go
smola: removed

smola:
@alcortesm Finally, it does support both absolute and relative paths, I did not realized that we're not working with arbitrary billy filesystems here, only OS.

repository.go
smola: Separating it into such functions would require doing redundant stats operations, or passing quite different arguments to each function. Could you review the new version and check if it's more readable now?

PR #364, opened 2017-04-26
plumbing: index, Entries converted in a slice of pointers

mcuadros:
In order to provide a nicer API, index.Entries was converted to `[]*Entry`


alcortesm:
Now you can simplify 219-231 a lot, and also make it much more efficient.

smola:
Maybe it makes sense to do what @alcortesm proposed in this same PR?

PR #365, opened 2017-04-27
transport: make Endpoint an interface, fixes #362

### LINKED ISSUE #362: Repository paths are always considered absolute
Our company git repository is hosted as a plain collection of git directories in the home dir of the hosting user `git`. That is when I clone `git@gitserver:a/b` I am actually cloning from `gitserver:/home/git/a/b`. The standard git client supports this setup without any further client side configuration.

Trying to clone the same repository with `go-git` results in a `repository not found` error. Adding the server-side base path to the repository on the client side, that is cloning `git@gitserver:home/git/a/b` solves the issue, but this behaviour is different from the standard client!

The error does not occur when a daemon is run with a correct value for `base-path` (such as in public services like github), but the general problem is that the path passed to the server is absolute, that is, it starts with a slash. Dropping the slash prefix in place where the URL is parsed keeps the support for github and Co, but also enables it for this most basic setup.

Here is what happens step by step: 

The `transport.NewEndpoint` function first converts `git@gitserver:a/b` into `ssh://git@gitserver/a/b`, then it calls the `url.Parse` method, which, in turn, returns the `URL` structure with the `Path` already set to `/a/b`. Then the client is created and it sends this absolute path to the server. Dropping the leading slash between parsing the original URL and returning the endpoint solves the issue, but seems ugly.

smola:
* add internal *url.URL implementation for regular URLs.
* add internal implementation for SCP-like URLs.

### DIFF
--- a/plumbing/transport/common.go
+++ b/plumbing/transport/common.go
@@ -18,6 +18,7 @@ import (
 	"io"
 	"net/url"
 	"regexp"
+	"strconv"
 
 	"gopkg.in/src-d/go-git.v4/plumbing"
 	"gopkg.in/src-d/go-git.v4/plumbing/protocol/packp"
@@ -28,7 +29,7 @@ var (
 	ErrRepositoryNotFound     = errors.New("repository not found")
 	ErrEmptyRemoteRepository  = errors.New("remote repository is empty")
 	ErrAuthenticationRequired = errors.New("authentication required")
-	ErrAuthorizationFailed	  = errors.New("authorization failed")
+	ErrAuthorizationFailed    = errors.New("authorization failed")
 	ErrEmptyUploadPackRequest = errors.New("empty git-upload-pack given")
 	ErrInvalidAuthMethod      = errors.New("invalid auth method")
 	ErrAlreadyConnected       = errors.New("session already established")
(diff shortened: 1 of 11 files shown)


plumbing/transport/common.go
```diff
@@ -88,42 +89,135 @@ type ReceivePackSession interface {
...
+// Endpoint represents a Git URL in any supported protocol.
+type Endpoint interface {
+	// Protocol returns the protocol (e.g. git, https, file). It should never
+	// return the empty string.
+	Protocol() string
+	// User returns the user or an empty string if none is given.
+	User() string
+	// Password returns the password or an empty string if none is given.
+	Password() string
+	// User returns the user or an empty string if none is given.
```
alcortesm: copypasta typo.

plumbing/transport/common.go
```diff
@@ -88,42 +89,135 @@ type ReceivePackSession interface {
...
+
+	return fmt.Sprintf("%s%s:%s", user, e.host, e.path)
+}
+
+var (
+	isSchemeRegExp   = regexp.MustCompile("^[^:]+://")
+	scpLikeUrlRegExp = regexp.MustCompile("^(?:(?P<user>[^@]+)@)?(?P<host>[^:]+):/?(?P<path>.+)$")
+)
+
+func parseSCPLike(endpoint string) Endpoint {
```
alcortesm: Returning `(Endpoint, bool)` would be more idiomatic, so it can be used like:
```
if e, ok := parseSCPLike(s) ; ok {
    return e, nil
}
// try other formats...
```

smola: :+1: done

plumbing/transport/common.go
smola: :+1: done

PR #368, opened 2017-04-28
do not convert local paths to URL

smola:
* Do not convert local paths to URLs, just keep them as they
  are.
* This way we add support for Windows without taking care of
  Windows path-to-URL conversion.

### DIFF
--- a/_examples/common_test.go
+++ b/_examples/common_test.go
@@ -96,7 +96,7 @@ func createBareRepository(dir string) string {
 
 func setEmptyRemote(dir string) string {
 	remote := createBareRepository(tempFolder())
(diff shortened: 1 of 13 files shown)


smola:
It seems it works :)

We used to get a lot of these errors in Windows tests:

```
FAIL: references_test.go:291: ReferencesSuite.TestObjectNotFoundError
references_test.go:300:
    c.Assert(err, IsNil)
... value *plumbing.PermanentError = &plumbing.PermanentError{Err:(*url.Error)(0xc043723950)} ("permanent client error: parse file://C:\\Users\\appveyor\\AppData\\Local\\Temp\\1\\tmp-tgz-678061493: invalid character \"\\\\\" in host name")
```

not anymore

PR #369, opened 2017-04-28
Partial windows support

smola:


mcuadros:
@smola still is WIP?

CI's are failing.

smola:
@mcuadros This is still WIP, but some of the commits are correct already. Let me split this so that we can start merging some parts.

mcuadros:
I am merging since is already an improve.

PR #374, opened 2017-05-03
format/packfile: fix bug when the delta depth is equals to 50

ajnavarro:


plumbing/format/packfile/delta_selector_test.go
```diff
@@ -197,3 +197,8 @@ func (s *DeltaSelectorSuite) TestObjectsToPack(c *C) {
 	c.Assert(otp[2].IsDelta(), Equals, true)
 	c.Assert(otp[2].Depth, Equals, 2)
 }
+
+func (s *DeltaSelectorSuite) TestMaxDepth(c *C) {
+	dsl := s.ds.deltaSizeLimit(0, 0, 50, true)
```
alcortesm: Use `maxDepth` instead of `50`.

alcortesm: Can we test this on a public function instead? private functions can change (or be removed) at any time, which may remove this test in the future.

ajnavarro: The logic to be able to create a delta with a depth of 50 is really complicated using public methods in `deltaSelector` (the only one is `ObjectsToPack()`). That's why I choose to test int using a private method.

mcuadros:
@ajnavarro ping

alcortesm:
Changes are still pending,  do you want me to ignore them and approve the PR instead?

plumbing/format/packfile/delta_selector_test.go
alcortesm: :+1:

ajnavarro:
Rebased and using `maxDepth` instead of 50.

mcuadros:
@ajnavarro CI is failing

alcortesm:
> plumbing/format/packfile/delta_selector_test.go:202: cannot use maxDepth (type int64) as type int in argument to s.ds.deltaSizeLimit

yes, a go vet issue, it should be easy to fix.

PR #375, opened 2017-05-04
worktree: Commit method implementation

mcuadros:
This is the implementation of a basic `git commit` command. 

With the objective of create `Tree` objects the `tree` and `parents` were make public, the names are not the best ones, but I will create another PR improving this. (I didn't wanted to become this PR in a monster)

Fixes https://github.com/src-d/go-git/issues/256

### DIFF
--- a/options.go
+++ b/options.go
@@ -43,7 +43,7 @@ type CloneOptions struct {
 	// Limit fetching to the specified number of commits.
 	Depth int
 	// RecurseSubmodules after the clone is created, initialize all submodules
-	// within, using their defaut settings. This option is ignored if the
+	// within, using their default settings. This option is ignored if the
 	// cloned repository does not have a worktree.
 	RecurseSubmodules SubmoduleRescursivity
 	// Progress is where the human readable information sent by the server is
@@ -73,7 +73,7 @@ func (o *CloneOptions) Validate() error {
 type PullOptions struct {
 	// Name of the remote to be pulled. If empty, uses the default.
 	RemoteName string
-	// Remote branch to clone.  If empty, uses HEAD.
+	// Remote branch to clone. If empty, uses HEAD.
 	ReferenceName plumbing.ReferenceName
 	// Fetch only ReferenceName if true.
 	SingleBranch bool
@@ -254,8 +254,7 @@ type LogOptions struct {
 }
 
 var (
-	ErrMissingAuthor    = errors.New("author field is required")
-	ErrMissingCommitter = errors.New("committer field is required")
+	ErrMissingAuthor = errors.New("author field is required")
 )
 
 // CommitOptions describes how a commit operation should be performed.
@@ -266,10 +265,10 @@ type CommitOptions struct {
 	// Author is the author's signature of the commit.
 	Author *object.Signature
 	// Committer is the committer's signature of the commit. If Committer is
-	// equal to nil the Author signature is used.
+	// nil the Author signature is used.
 	Committer *object.Signature
-	// Parents parents commits for the new commit, by default is the hash of
-	// HEAD reference.
+	// Parents are the parents commits for the new commit, by default when
+	// len(Parents) is zero, the hash of HEAD reference is used.
 	Parents []plumbing.Hash
 }
 
--- a/options.go
+++ b/options.go
@@ -5,6 +5,7 @@ import (
 
 	"gopkg.in/src-d/go-git.v4/config"
 	"gopkg.in/src-d/go-git.v4/plumbing"
+	"gopkg.in/src-d/go-git.v4/plumbing/object"
 	"gopkg.in/src-d/go-git.v4/plumbing/protocol/packp/sideband"
 	"gopkg.in/src-d/go-git.v4/plumbing/transport"
 )
@@ -42,7 +43,7 @@ type CloneOptions struct {
 	// Limit fetching to the specified number of commits.
 	Depth int
 	// RecurseSubmodules after the clone is created, initialize all submodules
-	// within, using their default settings. This option is ignored if the
+	// within, using their defaut settings. This option is ignored if the
 	// cloned repository does not have a worktree.
 	RecurseSubmodules SubmoduleRescursivity
 	// Progress is where the human readable information sent by the server is
@@ -250,4 +251,48 @@ type LogOptions struct {
 	// reachable from it. If this option is not set, HEAD will be used as
 	// the default From.
 	From plumbing.Hash
+}
+
+var (
+	ErrMissingAuthor    = errors.New("author field is required")
+	ErrMissingCommitter = errors.New("committer field is required")
+)
+
+// CommitOptions describes how a commit operation should be performed.
+type CommitOptions struct {
+	// All automatically stage files that have been modified and deleted, but
+	// new files you have not told Git about are not affected.
+	All bool
+	// Author is the author's signature of the commit.
+	Author *object.Signature
+	// Committer is the committer's signature of the commit. If Committer is
+	// equal to nil the Author signature is used.
+	Committer *object.Signature
+	// Parents parents commits for the new commit, by default is the hash of
+	// HEAD reference.
+	Parents []plumbing.Hash
+}
+
+// Validate validates the fields and sets the default values.
+func (o *CommitOptions) Validate(r *Repository) error {
+	if o.Author == nil {
+		return ErrMissingAuthor
+	}
+
+	if o.Committer == nil {
+		o.Committer = o.Author
+	}
+
+	if len(o.Parents) == 0 {
+		head, err := r.Head()
+		if err != nil && err != plumbing.ErrReferenceNotFound {
+			return err
+		}
+
+		if head != nil {
+			o.Parents = []plumbing.Hash{head.Hash()}
+		}
+	}
+
+	return nil
 }
--- a/plumbing/object/tag_test.go
+++ b/plumbing/object/tag_test.go
@@ -234,33 +234,32 @@ func (s *TagSuite) TestString(c *C) {
 	)
 }
 
-func (s *TagSuite) TestTagToTagString(c *C) {
+func (s *TagSuite) TestStringNonCommit(c *C) {
 	store := memory.NewStorage()
 
-	tagOneHash := plumbing.NewHash("TAGONE")
-	tagTwoHash := plumbing.NewHash("TAGTWO")
-
-	tagOne := &Tag{
-		Target:     tagTwoHash,
-		Hash:       tagOneHash,
+	target := &Tag{
+		Target:     plumbing.NewHash("TAGONE"),
 		Name:       "TAG ONE",
+		Message:    "tag one",
 		TargetType: plumbing.TagObject,
 	}
-	tagTwo := &Tag{
-		Target:     tagOneHash,
-		Hash:       tagTwoHash,
+
+	targetObj := &plumbing.MemoryObject{}
+	target.Encode(targetObj)
+	store.SetEncodedObject(targetObj)
+
+	tag := &Tag{
+		Target:     targetObj.Hash(),
 		Name:       "TAG TWO",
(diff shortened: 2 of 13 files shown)


options.go
```diff
@@ -251,3 +252,47 @@ type LogOptions struct {
...
 	From plumbing.Hash
 }
+
+var (
+	ErrMissingAuthor    = errors.New("author field is required")
+	ErrMissingCommitter = errors.New("committer field is required")
+)
+
+// CommitOptions describes how a commit operation should be performed.
+type CommitOptions struct {
```
ajnavarro: I'm missing the commit message.

alcortesm:
Please, fix CI.

options.go
```diff
@@ -251,3 +252,47 @@ type LogOptions struct {
...
+	}
+
+	if len(o.Parents) == 0 {
+		head, err := r.Head()
+		if err != nil && err != plumbing.ErrReferenceNotFound {
+			return err
+		}
+
+		if head != nil {
+			o.Parents = []plumbing.Hash{head.Hash()}
```
ajnavarro: If head is nil what happens?

plumbing/object/tag_test.go
```diff
@@ -234,41 +234,40 @@ func (s *TagSuite) TestString(c *C) {
 	)
 }
 
-func (s *TagSuite) TestTagToTagString(c *C) {
+func (s *TagSuite) TestStringNonCommit(c *C) {
```
ajnavarro: I don't understand why this is related to this PR, and the name changing.

mcuadros:
Solve, the problem was on master.

plumbing/object/tree.go
```diff
@@ -240,29 +240,21 @@ func (t *Tree) Encode(o plumbing.EncodedObject) error {
 		return err
 	}
 
-	var size int
```
ajnavarro: This change is related to this commit, or is just a separated code improvement?

worktree_commit.go
```diff
@@ -0,0 +1,243 @@
...
+)
+
+// Commit stores the current contents of the index in a new commit along with
+// a log message from the user describing the changes.
+func (w *Worktree) Commit(msg string, opts *CommitOptions) (plumbing.Hash, error) {
+	if err := opts.Validate(w.r); err != nil {
+		return plumbing.ZeroHash, err
+	}
+
+	if opts.All == true {
```
ajnavarro: better do:
```
if opts.All {
...
```

worktree_commit.go
```diff
@@ -0,0 +1,243 @@
...
+	"gopkg.in/src-d/go-git.v4/plumbing/object"
+	"gopkg.in/src-d/go-git.v4/storage"
+	"gopkg.in/src-d/go-git.v4/utils/ioutil"
+
+	"gopkg.in/src-d/go-billy.v2"
+)
+
+// Commit stores the current contents of the index in a new commit along with
+// a log message from the user describing the changes.
+func (w *Worktree) Commit(msg string, opts *CommitOptions) (plumbing.Hash, error) {
```
ajnavarro: Why the message is not part of the CommitOptions?

options.go
mcuadros: the commit message is at the method, the message is not a option

options.go
```diff
@@ -251,3 +252,47 @@ type LogOptions struct {
...
+	Committer *object.Signature
+	// Parents parents commits for the new commit, by default is the hash of
+	// HEAD reference.
+	Parents []plumbing.Hash
+}
+
+// Validate validates the fields and sets the default values.
+func (o *CommitOptions) Validate(r *Repository) error {
+	if o.Author == nil {
+		return ErrMissingAuthor
```
ajnavarro: You can test it easily.

options.go
mcuadros: that is the initial commit, and doesn't have any parents?

worktree_commit.go
```diff
@@ -0,0 +1,243 @@
...
+	return nil
+}
+
+func (h *commitIndexHelper) buildTree(e *index.Entry, parent, path string) bool {
+	if _, ok := h.trees[path]; ok {
+		return false
+	}
+
+	if _, ok := h.entries[path]; ok {
+		return false
```
ajnavarro: You can test it easily.

plumbing/object/tag_test.go
mcuadros: I was fixing the test because after changing the Write to auto-count the size, I figure out that the test was wrong.

1) The test functions are called "Test<method>[variant]" 
2) The test was wrong
3) Didn't worked after autocalculate the size, because you can't use a custom hash in the objects. But since size was 0, the hash was not being calculated.

plumbing/object/tree.go
mcuadros: https://github.com/src-d/go-git/pull/375/commits/ff18ce3751ad80cfd0297845872ba1d796c36ca5

worktree_commit.go
```diff
@@ -0,0 +1,243 @@
...
+
+func (w *Worktree) autoAddModifiedAndDeleted() error {
+	s, err := w.Status()
+	if err != nil {
+		return err
+	}
+
+	for path, fs := range s {
+		if fs.Worktree != Modified && fs.Worktree != Deleted {
+			continue
```
ajnavarro: You can test it easily.

worktree_commit.go
mcuadros: I followed the pattern that we have in some other options. I can imaging someone reusing `opts`, but never the message.

worktree_commit.go
mcuadros: lol

worktree_commit.go
```diff
@@ -0,0 +1,243 @@
...
+	"gopkg.in/src-d/go-git.v4/utils/ioutil"
+
+	"gopkg.in/src-d/go-billy.v2"
+)
+
+// Commit stores the current contents of the index in a new commit along with
+// a log message from the user describing the changes.
+func (w *Worktree) Commit(msg string, opts *CommitOptions) (plumbing.Hash, error) {
+	if err := opts.Validate(w.r); err != nil {
+		return plumbing.ZeroHash, err
```
ajnavarro: You can test it easily.

options.go
ajnavarro: I think should be an option (https://git-scm.com/docs/git-commit#git-commit--mltmsggt), but it's depending of how we want to manage the public API.

worktree_commit.go
alcortesm: I find the design of the commit operation very surprising:

- I would expect this method to receive what you want to commit (an index or whatever), along with the msg and the options.

- I would expect this to be a method of Repository or Storage or any other entity, except for worktree.

Maybe you should document the overall design of this somewhere.

worktree_commit.go
```diff
@@ -0,0 +1,243 @@
...
+	if err != nil {
+		return plumbing.ZeroHash, err
+	}
+
+	commit, err := w.buildCommitObject(msg, opts, tree)
+	if err != nil {
+		return plumbing.ZeroHash, err
+	}
+
+	return commit, w.updateHEAD(commit)
```
alcortesm: Please, explain that this method updates the head in the method signature, it is very important information about what the method does.

worktree_commit.go
```diff
@@ -0,0 +1,243 @@
...
+
+	obj := w.r.Storer.NewEncodedObject()
+	if err := commit.Encode(obj); err != nil {
+		return plumbing.ZeroHash, err
+	}
+	return w.r.Storer.SetEncodedObject(obj)
+}
+
+// commitIndexHelper converts a given index.Index file into multiple git objects
+// reading the blogs from the given filesystem and creating the trees from the
```
alcortesm: s/blogs/blobs/

worktree_commit.go
mcuadros: What you commit is what is in staging, is a very basic concept of git. If you want to do dirty things like storing an espontaneous commit just push is to the Storage. 

Since you are committing the staging and this is a concept of the worktree, I don't see why this should be at Repository.

I think that this is very natural way of doing a commit:
```go
r, err := Init(storage, fs)
w, err := r.Worktree()
billy.WriteFile(fs, "foo", []byte("foo"), 0644)
 w.Add("foo")
 w.Commit("foo\n", &CommitOptions{Author: defaultSignature()})
```

worktree_commit.go
mcuadros: When you are going a commit? I don't know if we should explain the basic of git, here.
I can add an extra documentation. But could be an endless job.

alcortesm: You are right, explaining all the things that a commit do is quite a lot.

Maybe the piece of information that is missing is that this method is supposed to emulate the porcelain git-commit command.  It was not clear to me if this was a low-level operation or a high-level one.

mcuadros: all the operations in git package are porceleain ones.

worktree_commit.go
alcortesm: You can have that if you want, with wrapper functions.  But you don't actually need to write the core logic like that to get that usage pattern.  The whole purpose of the index is to isolate the worktree from the repository:

 - there is a set of conceptual operations between the worktree and the index (like git add).

 - there is a different set of conceptual operations between the index and the repository (like git commit).

(except with a few concessions here and there due to pure usability that are always implemented at a very high level)

See, passing the index to the commit function avoids relaying on internal state and make it much easier to test and to mock.  Then moving the commit function to a class that is not the worktree will be help to distribute responsibilities where they originally belong in the real git.

worktree_commit.go
alcortesm: And yet we don't explain that important fact anywhere in the package documentation.

Why should we not document that important feature of the package? why is it better to hide that information from the user?, why not clearly mention that on the functions that are meant to emulate git commands (or the opposite on the ones that are not, if they are minority)?

worktree_commit.go
mcuadros: Sorry but I am complete missing. Why I will want to make to the user pass the index, the index is a plumbing concept. Not exposed on the `git` package. 

Mock and test is very easy because you can replace the storage where all is contained.

This is a small example of the jgit doing a commit with the porcelain API:
```java
Git git = new Git(db);
CommitCommand commit = git.commit();
commit.setMessage("initial commit").call();
```

worktree_commit.go
mcuadros: Explain what? We are not emulating the git command we are offering a easy way to manipulate a repository in the easy way. I invite you to document this if you want.

Is the main package and expose an API that is enough to manipulate a repository, the goal is not replicate a command, we are not doing this. 

Doing a commit implies add to a repository, explaining that this change the HEAD is a detail of implementation.

worktree_commit.go
alcortesm: Never mind, it is just a different points of view about the design.

These examples are very nice, maybe we can reuse them as actual real examples.  They will help with the review of the PR and with the general documentation of the library.

worktree_commit.go
alcortesm: :+1:

worktree_commit.go
mcuadros: What I pasted are not examples are the tests. As I said once that is merge I will do some examples

alcortesm: It would love to have these examples before the actual code, as a PR, so I can review them and understand the purposes of a change before reading the code.

worktree_commit.go
smola: Just a clarification, IMHO, we're not replicating git commands because there translating commands to Go functions as they are would lead to a weird API. However, we do get a lot of inspiration from Git commands, both plumbing and porcelain, so that we work with the same concepts (even if it's with a different API) and they look familiar to advance Git users.

This is reflected on the fact that in some functions we documented what command they are equivalent or similar to.

dreampuf:
When to release this?

PR #383, opened 2017-05-09
on demand http auth example

alcortesm:
The goal of the pseudo-code at #350 (comment) can be improved with this example.

### DIFF
--- a/_examples/http_auth/main.go
+++ b/_examples/http_auth/main.go
@@ -1,0 +1,42 @@
+package main
+
+import (
+	"os"
+
+	git "gopkg.in/src-d/go-git.v4"
+	. "gopkg.in/src-d/go-git.v4/_examples"
+	"gopkg.in/src-d/go-git.v4/plumbing/transport"
+	"gopkg.in/src-d/go-git.v4/plumbing/transport/http"
+)
+
+func main() {
+	CheckArgs("<url>", "<directory>")
+	url := os.Args[1]
+	directory := os.Args[2]
+
+	// Clone the given repository to the given directory
+	Info("git clone %s %s --recursive", url, directory)
+
+	opts := git.CloneOptions{
+		URL:               url,
+		RecurseSubmodules: git.DefaultSubmoduleRecursionDepth,
+		Auth:              nil,
+	}
+
+	_, err := git.PlainClone(directory, false, &opts)
+	if err != nil {
+		if err == transport.ErrAuthorizationRequired {
+			Warning("Authentication required")
+			Warning("Removing previously created directory")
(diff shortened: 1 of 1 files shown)


alcortesm:
umm, tests fails, any idea on how to write this example so it doesn't fail?

smola:
@alcortesm You'd need to get credentials from an environment variable, skip the tests if the variable is not defined and add credentials to Travis using the secrets API.

jdoklovic:
This gets me closer in terms of http auth, however, I think there should be a high-level "helper" to handle different url schemes.  e.g.:
```go
func AuthMethodForURL(url string) transport.AuthMethod {
    if IsSSH(url) {
        // 1:scheme, 2:user, 3:host, 4:path
        m := scpLikeUrlRegExp.FindStringSubmatch(url)
        a, _ := ssh.NewSSHAgentAuth(m[2])
        return a
    }
    ba := http.NewBasicAuth("", "")
    return ba
}
opts := git.CloneOptions{
        URL:      gitURL,
        Auth:    AuthMethodForURL(gitURL),
        Progress: os.Stdout,
}
err := git.Clone(directory, opts)

if err != nil {
    if err == transport.ErrAuthorizationRequired {
        Warning("Authentication required")
        Warning("Removing previously created directory")
        os.RemoveAll(directory)
        Warning("Trying again with some fake auth information")

        // ask for authentication credentials and try again...
        opts.Auth = http.NewBasicAuth("foo", "bar")
        _, err = git.PlainClone(directory, false, &opts)
        CheckIfError(err)
    } else {
        CheckIfError(err)
    }
}
```

Just saying, seems like this is going to be a common thing to do since git can handle multiple url schemes.
I might even consider putting a "fallback" on AuthMethod that gets called when there's an AuthRequired error so you could chain together different auth methods easily rather than having every consumer write this error handling code.

mcuadros:
@jdoklovic If what you want is extract an BasicAuth from the endpoint, this is already implemented. 
https://github.com/src-d/go-git/blob/master/plumbing/transport/http/upload_pack.go#L23-L27

```
PlainClone("http://user:pass@github.com/user/repository.git", ...) 
```

smola:
Actually, having a way to pass a default auth for any kind of URL could make sense. A helper to get the protocol out of a URL could indeed make sense, since this functionality already exists in the `transport` package but it's not expose to the public.

We could consider exposing certain functionality that makes implementing this easier.

smola:
I created a PR for a `clone` cli command that could subsume this example, but including all auth protocols: https://github.com/src-d/go-git/pull/399

alcortesm:
Awesome!

PR #394, opened 2017-05-16
transport/server: use Endpoint string representation as a map key.

ajnavarro:
Two endpoints are not equals between them, even if they were generated using the same url or path.


mcuadros:
Because is a pointer

PR #398, opened 2017-05-21
storage: filesystem, initialize the default folder scaffolding

mcuadros:
It creates the default scaffolding for a `.git` folder exactly as cgit does 

/cc @osklyar


osklyar:
@mcuadros When merged I will be able to simply the init logic in my client, nice!

PR #399, opened 2017-05-22
cli: add clone command

smola:
The purpose of this command is to provide a full example of how to replicate different git behaviors on clone, given the high amount of issues that we receive regarding to this: different options, auth handling, retry policies, etc.

### DIFF
--- a/cli/go-git/clone.go
+++ b/cli/go-git/clone.go
@@ -1,0 +1,146 @@
+package main
+
+import (
+	"fmt"
+	"os"
+	"path"
+	"strings"
+
+	"gopkg.in/src-d/go-git.v4"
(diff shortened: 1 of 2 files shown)


smola:
Discussed this with @mcuadros and we decided not to merge this. However, we still have to:
- Add more examples related to this.
- Create a getting started tutorial
- Maybe creating an extenal project with a `git` clone implemented with `go-git`.

The rationale is that we don't want to create false expectations on users about the support, compatibility or stability of this CLI. So we better create a separate project with separate release process and huge experimental disclaimers.

bzz:
Having separate something like a `go-git-examples` and moving this impl there sounds as great idea.

Issue #390, opened 2017-05-11
Pushing Tags Takes Many Minutes


ekyoung:
Pushing tags to a remote takes about 15 minutes in one repository I'm working with. Using `git push --tags` on the same repo only takes a second or two.

The code I'm using is similar to the below.

```go
rs := config.RefSpec("/refs/tags/v1.0.0:/refs/tags/v1.0.0")
err := rmt.Push(&git.PushOptions{
	RefSpecs: []config.RefSpec{rs},
})
```

The time seems to be taken in lot of loops through `func (dw *deltaSelector) tryToDeltify(base, target *ObjectToPack) error` going over blobs.

smola:
@ekyoung Thanks for the report. There are some performance problems when generating packfiles. We're looking into th is. For this case, we have to check if we're trying to send more objects than needed to the server, which would be aggravating the problem.

We'll look into this.

ekyoung:
Now takes about 1 minute on the repo I'm working with instead of 15. A big improvement. Thank you.

smola:
@ekyoung I'm glad it helped. There still some upcoming improvements that will make this faster.

PR #400, opened 2017-05-23
format/packfile: improve binary delta algorithm

ajnavarro:
Implement Xdelta algorithm from this paper (http://pop.xmailserver.net/xdfs.pdf) using adler32 checksum for fingerprinting and 16 bytes chunks, like in git. 

Fixes https://github.com/src-d/go-git/issues/390


ajnavarro:
Using the same source and target to create the deltas:
- old version: `379373078 ns/op`
- new version: `613210 ns/op`

alcortesm:
Speed!

ajnavarro:
@alcortesm I added the name of the author and the paper title, because the link can disappear.

Issue #371, opened 2017-04-28
Fetch remote tags to local


manuwell:
Hey guys, 

I'm trying to fetch remote tags but it seems that tags are ignored when fetching or pulling from remote. Trying to perform the following operations:

```go
         r.Fetch(&git.FetchOptions{
                 RemoteName: "origin",
         })

         refs, err := r.References()
         CheckIfError(err)

         err = refs.ForEach(func(ref *plumbing.Reference) error {
                 fmt.Println(ref)
                 return nil
         })
```

outputs
```
3dbebd24ff81327fd7171ccfa4d4ed81a6070069 refs/remotes/origin/master
3dbebd24ff81327fd7171ccfa4d4ed81a6070069 refs/heads/master
ref: refs/remotes/origin/master refs/remotes/origin/HEAD
ref: refs/heads/master HEAD
```

But no tags were fetched. What I'm trying to perform is a `git pull --tags` git operation. Is it already supported on `go-git`? 

Thanks and congrats for this project.

smola:
I think we're doing the equivalent of `git fetch --no-tags` and you can fetch tags explicitely by adding the refspec `+refs/tags/*:refs/tags/*` to your fetch. We could look into being able to replicate the default git behavior though.

manuwell:
I tried passing the suggested `refspec` but seems that object already exists. 

Debugging here I found this [piece of code](https://github.com/src-d/go-git/blob/727bf94da8e3cebd3ff467d30425b12d671fbca7/remote.go#L381) returning `true` to a fetched `tag` because, I think, it is looking for the commit hash instead the tag name to check the presence. 

The `storer` used on my local tests is `filesystem`. I couldn't go further in this problem.

smola:
@manuwell Ok, this does look like a bug, we'll look into it. Thanks!

PR #407, opened 2017-05-30
Fix fetch tags

zacharya19:
Quick fix for #371.
Not sure it's the best way, but this bug is very disturbing to me, so I fixed it quickly.

In tags we do not need to check hash, we need to check if it's name exists in the local storer, so I saved all local tags in slice and then if the reference is a tag, I say it's exists if it's in the slice.

Better idea?

### DIFF
--- a/remote.go
+++ b/remote.go
@@ -329,6 +329,21 @@ func getWants(
 		}
 	}
 
+	tags := []plumbing.Hash{}
+	localIter, err := localStorer.IterReferences()
+	if err != nil {
+		return nil, err
+	}
+
+	localIter.ForEach(func(ref *plumbing.Reference) error {
+		if !ref.IsTag() {
+			return nil
+		}
+
+		tags = append(tags, ref.Hash())
+		return nil
+	})
+
 	iter, err := remoteRefs.IterReferences()
 	if err != nil {
 		return nil, err
@@ -354,10 +369,15 @@ func getWants(
 		}
 
 		hash := ref.Hash()
+		var exists bool
 
-		exists, err := objectExists(localStorer, hash)
-		if err != nil {
-			return err
+		if !ref.IsTag() {
+			exists, err = objectExists(localStorer, hash)
+			if err != nil {
+				return err
+			}
+		} else {
+			exists = tagInList(hash, tags)
 		}
 
 		if !exists {
@@ -376,6 +396,16 @@ func getWants(
 	}
 
 	return result, nil
+}
+
+func tagInList(h plumbing.Hash, list []plumbing.Hash) bool {
+	for _, value := range list {
+		if value == h {
+			return true
+		}
+	}
+
+	return false
 }
 
 func objectExists(s storer.EncodedObjectStorer, h plumbing.Hash) (bool, error) {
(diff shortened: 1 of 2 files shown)


remote.go
```diff
@@ -329,6 +329,21 @@ func getWants(
 		}
 	}
 
+	tags := []plumbing.Hash{}
```
smola: Prefer `var tags []plumbing.Hash` initialization.

remote.go
```diff
@@ -329,6 +329,21 @@ func getWants(
...
+	localIter, err := localStorer.IterReferences()
+	if err != nil {
+		return nil, err
+	}
+
+	localIter.ForEach(func(ref *plumbing.Reference) error {
```
smola: Check return error.

smola:
@zacharya19 Thanks for the contribution!

I wonder how this works. If I understood correctly, this implies that the repository contains a tag returned by `IterReferences()` that points to a hash which does not exist in the `localStorer`.

That should not happen at all. Could you give more details about a repository exhibiting this error? Did you clone it with go-git from the beginning? Or did you clone with git and then pulled with go-git? I suspect that if this fix works, it's only because there is a related bug that has lead the repository into an inconsistent state.

zacharya19:
@smola I use PlainOpen to open a repo I cloned with git cli.
But there is something strange here, with my fix I get new tags with the same info but other name (I have tag named "release-1.23.1" and after I run the code, my local git has "release-1.23.1" and "1.23.1" with the same hash and message.
I'm not really sure why.

zacharya19:
BTW - the tag could not by exists locally if the hash does not exists locally, am I wrong?

smola:
@zacharya19 This code implies that tags have hashes of objects that are not in the object storage. A tag is either a tag object (in the object storage) or a reference to a non-tag object (in the object storage too). So if there are cases where `IterReferences()` return tags pointing to hashes that are not in the object storage (`objectExists(localStorer, hash) == false`) it might indicate a bug previos in the execution.

```
if !ref.IsTag() {
	exists, err = objectExists(localStorer, hash)
	if err != nil {
		return err
	}
} else {
	exists = hashInList(hash, tags)
}
```

zacharya19:
Not sure I got it, since the problem here is the opposite, on the remote we see tags that point to an hash that exists locally, but the tag it self does not exist.

smola:
@zacharya19 Ok. This can happen with lightweight tags. Although, this still wouldn't fix the case where a tag `T1`already exist locally pointing to hash `H1` and the server contains a new hash `T2` pointing to `H1` too.

zacharya19:
@smola I made some changes, let me know what you think.

mcuadros:
@smola

smola:
We should add a test for this though.

smola:
I'm working on a proper test before moving on.

zacharya19:
@smola I added basic test, let me know what you think.

mcuadros:
@zacharya19 thanks for your efforts and your time, but I created another PR https://github.com/src-d/go-git/pull/485 with a canonical solution.

PR #409, opened 2017-05-31
storage/filesystem: call initialization explicitly, fixes #408

### LINKED ISSUE #408: PlainOpen on non-git directory creates git directory structure
Calling `PlainOpen` on a directory path that is not a git repository has a side effect of creating the git directory structure within the directory:
```
├── objects
│   ├── info
│   └── pack
└── refs
    ├── heads
    └── tags
```

I would expect the following test to pass, but it fails as follows:
```go
func (s *RepositorySuite) TestPlainOpenEmptyDir(c *C) {
	dir, err := ioutil.TempDir("", "plain-open-empty-dir")
	c.Assert(err, IsNil)
	defer os.RemoveAll(dir)

	r, err := PlainOpen(dir)
	c.Assert(err, Equals, ErrRepositoryNotExists)
	c.Assert(r, IsNil)

	entries, err := ioutil.ReadDir(dir)
	c.Assert(err, IsNil)
	c.Assert(len(entries), Equals, 0)
}
```

However, it currently fails:
```
FAIL: repository_test.go:368: RepositorySuite.TestPlainOpenEmptyDir

repository_test.go:379:
    c.Assert(len(entries), Equals, 0)
... obtained int = 2
... expected int = 0

OOPS: 139 passed, 1 FAILED
--- FAIL: Test (33.04s)
```

smola:
filesystem.Storage was initializing the gitdir (creating objects
and refs) on NewStorage. But this should be done only on init and
clone operations, not on open.

Now there is a new interface storer.Initializer that storers can
implement if they need any initialization step before init or clone.
filesystem.Storage is one of such implementations.

git.Init and git.Clone now call to the storer Init() method if it
does implement it. Otherwise, it just ignores initialization.

### DIFF
--- a/storage/filesystem/storage_test.go
+++ b/storage/filesystem/storage_test.go
@@ -1,6 +1,7 @@
 package filesystem
 
 import (
+	"io/ioutil"
 	"testing"
 
 	"gopkg.in/src-d/go-git.v4/storage/test"
@@ -14,25 +15,17 @@ func Test(t *testing.T) { TestingT(t) }
 
 type StorageSuite struct {
 	test.BaseStorageSuite
+	dir string
 }
 
 var _ = Suite(&StorageSuite{})
 
 func (s *StorageSuite) SetUpTest(c *C) {
-	storage, err := NewStorage(osfs.New(c.MkDir()))
+	s.dir = c.MkDir()
+	storage, err := NewStorage(osfs.New(s.dir))
 	c.Assert(err, IsNil)
 
 	s.BaseStorageSuite = test.NewBaseStorageSuite(storage)
-}
-
-func (s *StorageSuite) TestNewStorage(c *C) {
-	fs := memfs.New()
-	storage, err := NewStorage(fs)
-	c.Assert(err, IsNil)
-	c.Assert(storage, NotNil)
-
-	_, err = fs.Stat("refs/tags")
-	c.Assert(err, IsNil)
 }
 
 func (s *StorageSuite) TestFilesystem(c *C) {
@@ -41,4 +34,10 @@ func (s *StorageSuite) TestFilesystem(c *C) {
 	c.Assert(err, IsNil)
 
 	c.Assert(storage.Filesystem(), Equals, fs)
+}
+
+func (s *StorageSuite) TestNewStorageShouldNotAddAnyContentsToDir(c *C) {
+	fis, err := ioutil.ReadDir(s.dir)
+	c.Assert(err, IsNil)
+	c.Assert(fis, HasLen, 0)
 }
--- a/plumbing/storer/storer.go
+++ b/plumbing/storer/storer.go
@@ -5,3 +5,11 @@ type Storer interface {
 	EncodedObjectStorer
 	ReferenceStorer
 }
+
+// Initializer should be implemented by storers that require to perform any
+// operation when creating a new repository (i.e. git init).
+type Initializer interface{
+	// Init performs initialization of the storer and returns the error, if
+	// any.
+	Init() error
+}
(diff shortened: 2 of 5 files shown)


smola:
Fixes https://github.com/src-d/go-git/issues/408

storage/filesystem/storage_test.go
```diff
@@ -8,18 +8,21 @@ import (
...
 	"gopkg.in/src-d/go-billy.v2/memfs"
 	"gopkg.in/src-d/go-billy.v2/osfs"
+	"io/ioutil"
 )
 
 func Test(t *testing.T) { TestingT(t) }
 
 type StorageSuite struct {
 	test.BaseStorageSuite
+	Dir string
```
alcortesm: Let's make this attribute private rather than public.

storage/filesystem/storage_test.go
```diff
@@ -42,3 +45,9 @@ func (s *StorageSuite) TestFilesystem(c *C) {
 
 	c.Assert(storage.Filesystem(), Equals, fs)
 }
+
+func (s *StorageSuite) TestStructureShouldBeCreatedLazily(c *C) {
```
alcortesm: The test name is misleading, what Structure?, and we are also not testing that contents are created lazily, just that the dir is created empty.

I suggest a new name for the test: `TestNewStorageShouldNotAddAnyContentsToDir`.

alcortesm:
Of course storage_test.go:31: StorageSuite.TestNewStorage is failing now, that was the whole point of the patch :).  By the way, the name of that test was not very informative :).

ajnavarro:
@alcortesm What I meant is, if the behavior has changed, this test should not test if the directories are created, so this test should pass too (removing the Stats line, the entire test because now does not make sense, or whatever)

alcortesm:
@ajnavarro yes, I understood you and I think you are right.  My comment was just to point out that the name of the test and the things it tested were not aligned.

I also agree with you that that whole test makes no sense.

storage/filesystem/storage_test.go
smola: :+1:

storage/filesystem/storage_test.go
smola: Renamed.

smola:
@alcortesm @ajnavarro Right. I removed `TestNewStorage`, we want exactly the oposite behavior.

alcortesm:
@mcuadros 

What do you mean by "the structure" and why it should be there?

Can you write a test to illustrate exactly what **has** to be there and what is not really needed until we use it for something?

smola:
@mcuadros and me talked about this offline and agreed that this should be done explicitly. That is: a storer needs a way to get called when initialization is being done.

The end result is essentially the same: as you see, the test has not changed.
The difference, however, is that now `git.Init` (and `git.Clone` indirectly) explicitly calls `Init()` on the storer if it implements it.

alcortesm:
I see, thanks!

storage/filesystem/storage_test.go
```diff
@@ -14,31 +15,29 @@ func Test(t *testing.T) { TestingT(t) }
...
 }
 
-func (s *StorageSuite) TestNewStorage(c *C) {
+func (s *StorageSuite) TestFilesystem(c *C) {
 	fs := memfs.New()
 	storage, err := NewStorage(fs)
 	c.Assert(err, IsNil)
-	c.Assert(storage, NotNil)
 
-	_, err = fs.Stat("refs/tags")
```
mcuadros: why we delete this test?

smola: Its only purpose was checking that refs/tags was created before any action. This is no longer the case, so it's enough if we check the other cases.

Issue #310, opened 2017-03-16
Can't push to private repo using SSH key


mdelillo:
Hi, I'm having some trouble pushing a commit to a private repo. I see the commit on my remote but my code hangs on the `remote.Push()` call.

Here's some sample code to reproduce the issue I'm running into:

```golang
package main

import (
	"fmt"
	"golang.org/x/crypto/ssh"
	"io/ioutil"
	"os"
	"os/exec"
	"path/filepath"
	"gopkg.in/src-d/go-git.v4"
	gitssh "gopkg.in/src-d/go-git.v4/plumbing/transport/ssh"
	"strconv"
	"time"
)

func main() {
	repoUrl := os.Getenv("REPO_URL")
	deployKey := os.Getenv("DEPLOY_KEY")

	gitDir, err := ioutil.TempDir("", "go-git-test")
	checkError(err)
	defer os.RemoveAll(gitDir)

	signer, err := ssh.ParsePrivateKey([]byte(deployKey))
	checkError(err)
	auth := &gitssh.PublicKeys{User: "git", Signer: signer}
	repo, err := git.PlainClone(gitDir, false, &git.CloneOptions{URL: repoUrl, Auth: auth})
	checkError(err)

	commitChange(gitDir)

	err = repo.Push(&git.PushOptions{Auth: auth})
	checkError(err)

	fmt.Println("Never gets here")
}

func commitChange(gitDir string) {
	timestamp := strconv.FormatInt(time.Now().UnixNano(), 10)
	ioutil.WriteFile(filepath.Join(gitDir, timestamp), nil, 0644)
	err := run(gitDir, "add", "-A")
	checkError(err)
	err = run(gitDir, "-c", "user.name=TEST", "-c", "user.email=<>", "commit", "-m", "Commit at "+timestamp)
	checkError(err)
}

func checkError(err error) {
	if err != nil {
		panic(err)
	}
}

func run(dir string, args ...string) error {
	cmd := exec.Command("git", args...)
	cmd.Dir = dir
	return cmd.Run()
}
```

I've tracked it down to these calls:
[...]

PR #418, opened 2017-06-12
fix push on git and ssh

smola:
This PR should fix this issue: https://github.com/src-d/go-git/issues/310
Push was getting blocked with `git` and `ssh` transport because connection was not closed properly.
This is fixed in the first commit of the PR. All the others are fixing other issues
that arised while adding code to test the original issue.

### DIFF
--- a/plumbing/transport/git/receive_pack_test.go
+++ b/plumbing/transport/git/receive_pack_test.go
@@ -1,0 +1,140 @@
+package git
+
+import (
+	"fmt"
+	"io"
+	"io/ioutil"
+	"net"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+	"time"
+
+	"github.com/src-d/go-git-fixtures"
+	"gopkg.in/src-d/go-git.v4/plumbing/transport"
+	"gopkg.in/src-d/go-git.v4/plumbing/transport/test"
+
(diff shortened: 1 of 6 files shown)


plumbing/transport/git/receive_pack_test.go
```diff
@@ -0,0 +1,134 @@
...
+
+	ep, err = transport.NewEndpoint(fmt.Sprintf("git://localhost:%d/empty.git", port))
+	c.Assert(err, IsNil)
+	s.ReceivePackSuite.EmptyEndpoint = ep
+
+	ep, err = transport.NewEndpoint(fmt.Sprintf("git://localhost:%d/non-existent.git", port))
+	c.Assert(err, IsNil)
+	s.ReceivePackSuite.NonExistentEndpoint = ep
+
+	fmt.Println("BASE", base)
```
ajnavarro: leftover?

plumbing/transport/internal/common/common.go
```diff
@@ -329,10 +333,12 @@ func (s *session) checkNotFoundError() error {
 }
 
 var (
```
ajnavarro: just a question, why these strings are variables instead of constants?

smola: they could actually be constants ;)

PR #419, opened 2017-06-12
plumbing/transport/http: add callback based basic auth

roboll:
Hey - great work. 

I'm using go-git in a Github app which uses dynamic credentials, so having http basic auth backed by a callback function is really useful. Let me know what you think of this.

### DIFF
--- a/plumbing/transport/http/common.go
+++ b/plumbing/transport/http/common.go
@@ -113,6 +113,34 @@ func (a *BasicAuth) String() string {
 	return fmt.Sprintf("%s - %s:%s", a.Name(), a.username, masked)
 }
 
+// BasicAuthCallback represent a HTTP basic auth with a password callback
+type BasicAuthCallback struct {
+	username string
+	callback func() (password string)
+}
+
+// NewBasicAuthCallback returns a basicAuthCallback based on the callback function
+func NewBasicAuthCallback(username string, callback func() (password string)) *BasicAuthCallback {
+	return &BasicAuthCallback{username, callback}
+}
+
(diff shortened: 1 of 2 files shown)


plumbing/transport/http/common.go
```diff
@@ -113,6 +113,34 @@ func (a *BasicAuth) String() string {
...
+	username string
+	callback func() (password string)
+}
+
+// NewBasicAuthCallback returns a basicAuthCallback based on the callback function
+func NewBasicAuthCallback(username string, callback func() (password string)) *BasicAuthCallback {
+	return &BasicAuthCallback{username, callback}
+}
+
+func (a *BasicAuthCallback) setAuth(r *http.Request) {
```
mcuadros: this function is completely untested, can you cover it?

smola:
So far the preferred way to implement this kind of logic is to set the basic auth from the beginning, or using `ErrAuthRequired` to ask for the password only when required. See example: https://github.com/src-d/go-git/issues/350#issuecomment-300096008

smola:
There is also a pending PR adding an example about intended usage: https://github.com/src-d/go-git/pull/383

roboll:
hey @smola - I did see the other PR and example usage. I _think_ this code solves an additional use case, which is constantly changing credentials. For example, I'm using this library with credentials that are rotated every hour as part of a [Github App](https://developer.github.com/apps/).

I could conceivably handle it externally, but this felt like a common enough use case to try to include here. If you'd rather not, I understand.

@mcuadros If you all think you may accept this, then yes, I'll definitely follow up with some test coverage for that function :+1:.

Issue #340, opened 2017-04-11
support for .gitignore in the worktree


mcuadros:
Currently the `.gitignore` is not implemented

mcuadros:
duplicate https://github.com/src-d/go-git/issues/307

Issue #307, opened 2017-03-13
Walking through tree of non `gitignore`d files


cirocosta:
Hey,
first i'd like to say that `go-git` is amazing; nice work! Discovered this week and still playing around with it :)

Now to the question, is it possible to walk through the tree of objects and gather only those that are not ignored by a `.gitignore` file? What i'd like to do is clone a repository, apply some changes (which might produce some files that would be ignored) and then walk through this tree.

Thx!

smola:
@cirocosta go-git has no `.gitignore` support yet, so you'd have to read .gitignore first, and then filter the tree yourself.

We are not planning to work on this in the short term, but PRs are welcome.

osklyar:
PR enclosed

cirocosta:
Great! Thx @osklyar  :D

osklyar:
Closed the original PR, but am working on a new fix for this issue.

osklyar:
New PR #429 wit a complete implementation

PR #420, opened 2017-06-12
Adds support for .gitignore at repository root

osklyar:
This PR adds basic support for `.gitignore` files. The code uses `github.com/monochromegane/go-gitignore`, which, however, could be easily ported into a local package to avoid an additional external dependency.

The PR covers 99% of the normal `.gitignore` usage, however, the following limitations as compared to the standard git client exist:
* only `.gitignore` files at the root of the repository are considered, files located deeper in the tree structure are not taken into account
* files matching the exclusion pattern in the `.gitignore` will be considered not present on the file system, so if forcefully committed, they may be shown as a change (has not been investigated).

The effect of the change is easily seen in the following example using the `go-git` based client from [github.com/silvertern/geat](https://github.com/silvertern/geat) (WIP):

Older state, without the PR:
```
➜  geat git:(01172b3) geat status
.: git@github.com:silvertern/geat.git [origin] HEAD Changed
?? .idea/workspace.xml
?? .idea/geat.iml
?? .idea/modules.xml
?? .idea/vcs.xml
?? .idea/libraries/GOPATH__geat_.xml
?? .idea/misc.xml
?? vendor/github.com/caarlos0/spin/.gitignore
?? vendor/github.com/caarlos0/spin/.travis.yml
?? vendor/github.com/caarlos0/spin/README.md
?? vendor/github.com/caarlos0/spin/spin_test.go
?? vendor/github.com/caarlos0/spin/Makefile
?? vendor/github.com/caarlos0/spin/spin.go
?? Gopkg.lock
```

With the PR:
```
➜  geat git:(master) dep ensure
➜  geat git:(master) go install
➜  geat git:(master) geat status
.: git@github.com:silvertern/geat.git [origin] master Clean

➜  geat git:(master) echo "change" >> Gopkg.toml 
➜  geat git:(master) ✗ geat status                
.: git@github.com:silvertern/geat.git [origin] master Changed
 M Gopkg.toml
```

Addresses #307, #340

### DIFF
--- a/utils/merkletrie/filesystem/node.go
+++ b/utils/merkletrie/filesystem/node.go
@@ -5,6 +5,8 @@ import (
 	"os"
 	"path/filepath"
 
+	"github.com/monochromegane/go-gitignore"
+
 	"gopkg.in/src-d/go-billy.v2"
 	"gopkg.in/src-d/go-git.v4/plumbing"
 	"gopkg.in/src-d/go-git.v4/plumbing/filemode"
@@ -25,6 +27,7 @@ type node struct {
 	submodules map[string]plumbing.Hash
 
 	path     string
+	matcher  gitignore.IgnoreMatcher
 	hash     []byte
 	children []noder.Noder
 	isDir    bool
@@ -39,7 +42,11 @@ func NewRootNode(
 	fs billy.Filesystem,
 	submodules map[string]plumbing.Hash,
 ) noder.Noder {
-	return &node{fs: fs, submodules: submodules, isDir: true}
+	n := &node{fs: fs, submodules: submodules, isDir: true}
+	if f, err := n.fs.Open(filepath.Join(n.path, ".gitignore")); err == nil {
+		n.matcher = gitignore.NewGitIgnoreFromReader(n.path, f)
+	}
+	return n
 }
 
 // Hash the hash of a filesystem is the result of concatenating the computed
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -1 +1,4 @@
 coverage.out
+vendor/
+.idea/
```
alcortesm: Please avoid adding content related to your choice of tools and editors to the project.

I suggest using a global $HOME/.gitignore file as an alternative.
```
git config --global core.excludesfile $HOME/.gitignore
```

osklyar:
@alcortesm done

osklyar:
@mcuadros The limitation has been listed in the original PR message, so it is a known limitation. I would generally agree with you for the change to be complete, but it would be much more involved and I would rather suggest doing it in a second iteration. I would not be able to offer a PR now for a change as suggested for the reasons of time constraints.

mcuadros:
@osklyar what I am asking is for move this to Worktree, not about implement git add force. Otherwise we will need to redo this PR as soon as we need the force, or any other similar feature.

osklyar:
@mcuadros Ok, fair enough. If it is more or less a move of the code I will investigate and try to provide the corresponding change.

osklyar:
Ok, I now know how to introduce this change in the Worktree directly, however, I have found out that the third-party gitignore pattern matcher that I've been using the change is incorrect on some patterns. Therefore, I will close this PR, work on a change to have a clean pattern matcher in the Worktree and then submit a new one.

PR #423, opened 2017-06-13
transport/ssh: allow passing SSH options

smola:
Adds the possibility of passing options to SSH transport.
Options have the form of functions modifying ssh.ClientConfig.

### DIFF
--- a/plumbing/transport/ssh/common.go
+++ b/plumbing/transport/ssh/common.go
@@ -3,6 +3,7 @@ package ssh
 
 import (
 	"fmt"
+	"reflect"
 
 	"gopkg.in/src-d/go-git.v4/plumbing/transport"
 	"gopkg.in/src-d/go-git.v4/plumbing/transport/internal/common"
@@ -11,16 +12,12 @@ import (
 )
 
 // DefaultClient is the default SSH client.
-var DefaultClient = NewClient()
+var DefaultClient = NewClient(nil)
 
-// NewClient creates a new SSH client with the given options.
-func NewClient(opts ...ClientOption) transport.Transport {
-	return common.NewClient(&runner{options: opts})
+// NewClient creates a new SSH client with an optional *ssh.ClientConfig.
+func NewClient(config *ssh.ClientConfig) transport.Transport {
+	return common.NewClient(&runner{config: config})
 }
-
-// ClientOption is a function that gets a standard ssh.ClientConfig and modifies
-// it.
-type ClientOption func(config *ssh.ClientConfig)
 
 // DefaultAuthBuilder is the function used to create a default AuthMethod, when
 // the user doesn't provide any.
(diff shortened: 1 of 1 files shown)


plumbing/transport/ssh/common.go
```diff
@@ -11,7 +11,16 @@ import (
 )
 
 // DefaultClient is the default SSH client.
-var DefaultClient = common.NewClient(&runner{})
+var DefaultClient = NewClient()
+
+// NewClient creates a new SSH client with the given options.
+func NewClient(opts ...ClientOption) transport.Transport {
```
mcuadros: why not just accepting a `ssh.ClientConfig`?

smola: To play nice with other options that modify `ssh.ClientConfig`:

```
 	config := c.auth.clientConfig()
 	config.HostKeyCallback, err = c.auth.hostKeyCallback()
 	if err != nil {
 		return err
 	}
```

`ssh.ClientConfig` is created and initialized by client on each session. Some fields are set based on `AuthMethod`, which is not global, but per-session setting. Then setting the new options for `ssh.ClientConfig` are set globally for the transport, but applied per session after it has been created.

mcuadros: You can archive the same just make a copy of the `ssh.ClientConfig`. And the API for the go-git user will be more simple and easy to use.

smola: That's right. The only difference is that global options could not override session values (e.g. adding a global override on every `HostKeyCallback`, such as an interceptor). Is that ok?

mcuadros: you can, checking if is zero value before assign the `HostKeyCallback`

smola: But you cannot use it to have a _global wrapper_ (instead of a _global override_). As in:

```
if c.HostKeyCallback != nil {
    c.HostKeyCallback = globalWrapper(c.HostKeyCallback)
}
```

Note that we discussed about this offline and agreed to make the API simpler with just a default `ssh.ClientConfig` that is copied. I will update the PR accordingly.

smola:
@mcuadros Is this what you had in mind? Note that it's still not tested.

plumbing/transport/ssh/common.go
```diff
@@ -129,3 +140,25 @@ func (c *command) setAuthFromEndpoint() error {
 func endpointToCommand(cmd string, ep transport.Endpoint) string {
 	return fmt.Sprintf("%s '%s'", cmd, ep.Path())
 }
+
+func overrideConfig(overrides *ssh.ClientConfig, c *ssh.ClientConfig) {
+	if overrides == nil {
+		return
+	}
+
+	vo := reflect.ValueOf(*overrides)
```
mcuadros: For do the copy should be enough something like:

```
var config *ssh.ClientConfig
*config = *overrides
```

smola: But we don't want to override values set for the session, such as `HostKeyCallback`.

Issue #425, opened 2017-06-14
Refs not updated when NoErrAlreadyUpToDate


taralx:
When fetching, if the remote ref is already available, the local ref is not updated per the refspec.

smola:
@taralx Could you elaborate more about the issue? If `NoErrAlreadyUpToDate` is returned it's because there is no reference to update.
Could you add which version of go-git are you using?

taralx:
Version: v4.0.0rc10

Basically, do a Fetch("+refs/heads/*:refs/heads/*") and then change one of the refs to its parent. If you do Fetch again, it will not update the ref because getWants returns nothing because the object is already present.

PR #434, opened 2017-06-16
Update local remote references during fetch even if no pack needs to be received

orirawlings:
I recently ran into some behavior that was not consistent with the standard git tooling.

It seems that doing a `(*Remote).Fetch()` will not update the local remote refs (i.e. `refs/remotes/origin/*`) if we already have all the new commit objects locally. The standard git tooling will continue to update the local remote refs even if it doesn't need to fetch a pack file with new commit objects.

You can see an example below where we are in a repository that already has both `7a58be` and `2d8e82`.

    $ git show-ref
    7a58be863cf6c431619df28b709b3bc87f2a2e29 refs/heads/master
    2d8e82007fa1df7f9e203d56543f95388cac9449 refs/remotes/origin/HEAD
    2d8e82007fa1df7f9e203d56543f95388cac9449 refs/remotes/origin/master

But the current remote references are out of date:

    $ git ls-remote origin
    7a58be863cf6c431619df28b709b3bc87f2a2e29  HEAD
    7a58be863cf6c431619df28b709b3bc87f2a2e29  refs/heads/master

The fetch will not pull down any new commit objects, but will proceed with updating the remote refs in the local repository.

    $ git fetch origin
    From github.com:orirawlings/test
       2d8e820..7a58be8  master     -> origin/master
    $ git show-ref
    7a58be863cf6c431619df28b709b3bc87f2a2e29 refs/heads/master
    7a58be863cf6c431619df28b709b3bc87f2a2e29 refs/remotes/origin/HEAD
    7a58be863cf6c431619df28b709b3bc87f2a2e29 refs/remotes/origin/master

I'm not really sure from an API design side if we want to continue to return the `NoErrAlreadyUpToDate` error, but the current PR changes preserve this in the response.

### DIFF
--- a/remote.go
+++ b/remote.go
@@ -47,7 +47,7 @@ func (r *Remote) String() string {
 
 // Fetch fetches references from the remote to the local repository.
 // Returns nil if the operation is successful, NoErrAlreadyUpToDate if there are
-// no changes to be fetched, or an error.
+// no changes to be fetched and no local references to update, or an error.
 func (r *Remote) Fetch(o *FetchOptions) error {
 	_, err := r.fetch(o)
 	return err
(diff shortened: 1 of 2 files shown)


orirawlings:
I've updated this PR so that `NoErrAlreadyUpToDate` is only returned if no pack is fetched and none of the local references need to be created/updated.

I'm not really sure why the code coverage report dropped for some of the `plumbing/transport/ssh` files. I wouldn't expect that to be related to this change. False positive?

Also, not really sure what is going on with the AppVeyor build either. Looks like some latent issues with windows support.

taralx:
Heh, funny. I reported this bug as #425.

orirawlings:
Hmm. I must have overlooked the issue when I opened the PR. Thanks for pointing that out @taralx!

mcuadros:
Yep, the SSH tests only runs under the master branch since the repository contains an encrypted key. 
AppVeyor is there to remember us that Windows support should be fixed.

Issue #358, opened 2017-04-24
Symlinks are checked out as regular files.


jmalloc:
When attempting to clone a repository containing symlinks, the links are created as regular files directly from the object blobs - i.e, the file contents is the symlink target.

I started an attempt to implement the correct behaviour in `Worktree.checkoutEntry()` but it looks like the filesystem abstraction does not support symlink creation.

The following snippet clones a repo that demonstrates the issue.

```go
package main

import (
	"fmt"
	"os"
	"path"

	git "gopkg.in/src-d/go-git.v4"
)

func main() {
	dir := path.Join(os.TempDir(), "symlink-test")
	fmt.Println(dir)

	_, err := git.PlainClone(
		dir,
		false, // bare
		&git.CloneOptions{
			URL:      "git@github.com:recoilphp/dev.git",
			Progress: os.Stdout,
		},
	)

	if err != nil {
		panic(err)
	}
}
```

smola:
@jmalloc Thanks for reporting this. We first have to figure out how do we want symlinks to be handled by billy, then fixing this should be easy.

jmalloc:
:+1: I'm happy to contribute towards this with some guidance. I expect the `os` based implementation will be straight forward, but I wasn't sure how to tackle the memory implementation.

smola:
@jmalloc Are you working on this? Feel free to create a partial PR to get further feedback, we can also help completing it.

PR #436, opened 2017-06-18
worktree: symlink support

mcuadros:
Fixes: https://github.com/src-d/go-git/issues/358

### DIFF
--- a/worktree.go
+++ b/worktree.go
@@ -4,7 +4,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
-	"io/ioutil"
+	stdioutil "io/ioutil"
 	"os"
 	"path/filepath"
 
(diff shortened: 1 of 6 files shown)


worktree.go
```diff
@@ -352,6 +356,21 @@ func (w *Worktree) checkoutFile(f *object.File) error {
 	return err
 }
 
+func (w *Worktree) checkoutFileSymlink(f *object.File) error {
+	from, err := f.Reader()
+	if err != nil {
+		return err
+	}
+	defer from.Close()
```
ajnavarro: we are missing the close error.

worktree.go
```diff
@@ -328,16 +328,20 @@ func (w *Worktree) checkoutChangeRegularFile(name string,
...
-	mode, err := f.Mode.ToOSFileMode()
+	if mode&os.ModeSymlink != 0 {
+		return w.checkoutFileSymlink(f)
+	}
+
+	from, err := f.Reader()
 	if err != nil {
 		return err
 	}
+	defer from.Close()
```
ajnavarro: we are missing the close error.

PR #437, opened 2017-06-18
plumbing/transport: export Transport impls

### LINKED ISSUE #424: Export urlEndpoint
Can urlEndpoint be exported? Right now I have to String() and re-parse the endpoint in order to get the URL object out for my custom transport. :-(

taralx:
Fixes #424

### DIFF
--- a/plumbing/transport/common.go
+++ b/plumbing/transport/common.go
@@ -129,17 +129,17 @@ func NewEndpoint(endpoint string) (Endpoint, error) {
 		))
 	}
 
-	return urlEndpoint{u}, nil
+	return URLEndpoint{u}, nil
 }
 
-type urlEndpoint struct {
+type URLEndpoint struct {
 	*url.URL
 }
 
-func (e urlEndpoint) Protocol() string { return e.URL.Scheme }
-func (e urlEndpoint) Host() string     { return e.URL.Hostname() }
+func (e URLEndpoint) Protocol() string { return e.URL.Scheme }
+func (e URLEndpoint) Host() string     { return e.URL.Hostname() }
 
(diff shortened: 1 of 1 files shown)


plumbing/transport/common.go
```diff
@@ -228,10 +229,10 @@ func parseSCPLike(endpoint string) (Endpoint, bool) {
...
 
 	m := scpLikeUrlRegExp.FindStringSubmatch(endpoint)
-	return &scpEndpoint{
-		user: m[1],
-		host: m[2],
-		path: m[3],
+	return &SCPEndpoint{
+		Username: m[1],
+		Hostname: m[2],
+		Pathname: m[3],
```
mcuadros: can you keep `Pathname` as `Path`?

taralx: I cannot, for the same reason as I needed to rename the other fields -- they conflict with the method names.

taralx: Happy to rename it if you can think of something useful to name it. Alternatively I could convert it into a type alias of a string instead of a struct.

taralx:
Ping?

mcuadros:
@taralx sorry about the delay, I am ok exporting the types, but you don't need to export the fields, since they can be accessed by the functions. This works for you?

taralx:
@mcuadros In that case there's no advantage in exporting them, since they cannot be constructed.

mcuadros:
Can you explain your use case? Maybe you can create your own endpoint

taralx:
It's #424 as specified in the commit message. If you don't want to export the fields of the others, that's fine by me -- I just don't know that it's worth exporting their types either if they're not constructible.

PR #429, opened 2017-06-15
Adds .gitignore support

osklyar:
This is a follow up from #420. As I am not sure if the PR approach is the one that you expect, I request feedback on this WIP change before I introduce all the tests, which are the only missing bits.

Improvements on top of the original PR #420:

* As requested, the functionality is added into `Worktree` this time. 
* Because the initially used library for matching gitignore patterns had a lot of gaps concerning even some simple cases I have written a [new one](https://github.com/silvertern/gitignore) from scratch, fully tested to the gitignore pattern specification. I will round up the documentation for that one in due course, but it is both feature and test complete.
* Further, this PR introduces support for nested .gitignore files according to the specification, not just the top level one as in #420. It would be easily extensible to higher prio patterns given on a command line or lower prio patterns from the environment.

PR #444, opened 2017-06-20
Fixes checkout not possible with (untracked) files under gitignore

osklyar:
This is a follow up for #429. The original implementation was based on the assumption that `Worktree.Status` is used to check differences between the index and the worktree. However, the `Worktree.Checkout` uses `Worktree.diffStagingWithWorktree` (via `Worktree.cointainsUnstagedChanges`), which is also used by the `Status`. This has prevented a checkout of a branch due to "unstaged changes" even when the status shows clean. 

In this PR the application of ignores is moved from `Status` directly into `diffStagingWithWorktree`, so the `Status` logic remains unchanged, but now the same logic applies to `Checkout`. With this change `Checkout` proceeds as there are no unstaged changes any longer (listed in ignores).

### DIFF
--- a/worktree_status.go
+++ b/worktree_status.go
@@ -68,8 +68,6 @@ func (w *Worktree) status(commit plumbing.Hash) (Status, error) {
 		return nil, err
 	}
 
-	right = w.excludeIgnoredChanges(right)
-
 	for _, ch := range right {
 		a, err := ch.Action()
 		if err != nil {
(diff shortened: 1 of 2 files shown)


mcuadros:
can you provide a test?

osklyar:
The Status test is still covered by the original one, will do one for Checkout

mcuadros:
You should provide a test where the checkout has ignored files.

osklyar:
This is exactly what the test does: (1) checkout on an empty repo -- no errors, (2) file `file` added and checkout produces an error, (3) `.gitignore` added and checkout still produces an error (`file` is no more a problem, but `.gitignore` itself is, (4) `.gitignore` staged and now checkout produces no error because `file` is excluded from unstaged by `.gitignore`

PR #446, opened 2017-06-20
WIP: Improves gitignore performance & adds (future) option for force-commit of gitignored files

osklyar:
This PR contains #444 and builds upon it. 

Changed to WIP: testing on a collection of large repositories with nested .gitignore some inconsistencies have been identified

*The problem*: having gitignore matcher acting in the worktree on the diffs results in traversing over potentially massive trees of file names of directories which are fully excluded (think e.g. of a `vendor/` gitignore record for Go modules). This is noticeably slow. 

*Solution*: Unless forcing an ignored file into the commit, one could exclude these files at the file system level in the noder (as originally suggested), so the noder would not dive into the directory at all and save uncovering and then removing thousands of files. This PR combines the two approaches: if force disabled, then files are ignored at the file system level, if force is enabled, file system delivers the full tree and files are ignored in the diff, however those marked M or D in the worktree are forced through, along with all in staging. The point here was not to implement the force, but to speed up the normal case, and force comes as an added value, albeit _not_ integrated into Status or Commit. Currently it is just a var flag set to false and only reset to true in the test to demo that force would work this way if added to Status and Commit.

### DIFF
--- a/utils/merkletrie/filesystem/node.go
+++ b/utils/merkletrie/filesystem/node.go
@@ -4,11 +4,13 @@ import (
 	"io"
 	"os"
 	"path/filepath"
+	"strings"
 
 	"gopkg.in/src-d/go-billy.v3"
 	"gopkg.in/src-d/go-git.v4/plumbing"
 	"gopkg.in/src-d/go-git.v4/plumbing/filemode"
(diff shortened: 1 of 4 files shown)


osklyar:
Won't fix now, maybe later

PR #472, opened 2017-07-10
improve delete support on push

smola:
* server: implement delete-refs and announce it.
* remote: check if server announced delete-refs before trying
   to delete and fail fast if it does not.
    
Note that the client does not need no send 'delete-refs' back
to the server to be able to delete references:

 ```
delete-refs
-----------
    
If the server sends back the 'delete-refs' capability, it means that
it is capable of accepting a zero-id value as the target
value of a reference update.  It is not sent back by the client, it
simply informs the client that it can be sent zero-id values
to delete references.
```
    
So our server implementation does not check if the client sent
delete-refs back, it just accepts deletes if it receives them.

### DIFF
--- a/remote.go
+++ b/remote.go
@@ -21,7 +21,10 @@ import (
 	"gopkg.in/src-d/go-git.v4/utils/ioutil"
 )
 
-var NoErrAlreadyUpToDate = errors.New("already up-to-date")
+var (
+	NoErrAlreadyUpToDate     = errors.New("already up-to-date")
+	ErrDeleteRefNotSupported = errors.New("server does not support delete-refs")
+)
 
 // Remote represents a connection to a remote repository.
 type Remote struct {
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: remote.go [hunk 1]
```diff
@@ -21,7 +21,10 @@ import (
 	"gopkg.in/src-d/go-git.v4/utils/ioutil"
 )
 
-var NoErrAlreadyUpToDate = errors.New("already up-to-date")
+var (
+	NoErrAlreadyUpToDate          = errors.New("already up-to-date")
+	ErrServerDoesNotSupportDelete = errors.New("server does not support delete")
```
mcuadros: ErrDeleteRefNotSupported = errors.New("server does not support delete-refs")

smola: :+1:

smola:
Updated the PR, since my initial approach was not fully working. @ajnavarro is now verifying this in `src-d/borges`.

PR #474, opened 2017-07-11
git: unify ErrObjectNotFound with plumbing.ErrObjectNotFound

taralx:
This minimizes surprises and also avoids having to swap one for the
other.

### DIFF
--- a/repository.go
+++ b/repository.go
@@ -22,7 +22,7 @@ import (
 )
 
 var (
-	ErrObjectNotFound          = errors.New("object not found")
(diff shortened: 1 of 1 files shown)


mcuadros:
I prefer replace all the git.ErrNotFound with the plumbing.ErrNotFound. Can you take care of this @taralx

taralx:
I can, but I'd rather do that in a separate PR?

mcuadros:
@taralx as you want

taralx:
On second thought, I'll just take care of it directly.

PR #485, opened 2017-07-17
remote: fetch, correct behavior on tags

mcuadros:
This fix https://github.com/src-d/go-git/issues/371 and https://github.com/src-d/go-git/pull/407

This PR remove the all behavior or fetching all the tags from a remote when a wildcard is being use, to the correct one. 

It implements two different options `TagFollowing`, this means that fetch the tags from objects being fetch (or fetched) making usage of the `include-tag` capability, or the option `AllTags` downloading all the available tags.

### DIFF
--- a/options.go
+++ b/options.go
@@ -103,6 +103,18 @@ func (o *PullOptions) Validate() error {
 	return nil
 }
 
+type TagFetchMode int
+
+var (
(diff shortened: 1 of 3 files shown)


mcuadros:
Yep, is kind of useless, and git is not using it when `--tags` is used.

Issue #481, opened 2017-07-14
How to create remote branch?


svetomir:
Hi,

Can you please write an example on how to create branch from an existing branch in a remote repository?

`git checkout -b release develop`
`git push origin release`

Thanks a lot

orirawlings:
The push step is well covered by `_examples/push/main.go`. The harder part here is creating the new branch.

Currently the API is a little awkward for creating new references in a `Repository`. This affects creating branches and lightweight tags. #430 has an example for creating lightweight tags.

We would probably benefit from some CreateBranch and/or CreateTag methods in `Repository`.

Issue #430, opened 2017-06-15
How to work with tags ?


se0wtf:
Hello,
I have a question about how to handle tags... i know they are supported but i don't know how to : 
- get the tag list
- create a new tag and push it

Everything else is working fine (clone/add/commit/push) but i did'nt find an example about the tags.

Can you explain it ?
Thanks guys !

avarabyeu:
+1

orirawlings:
Here is some example code to demonstrate how to list tags:

```
package main

import (
	"fmt"
	"os"

	"gopkg.in/src-d/go-git.v4"
	. "gopkg.in/src-d/go-git.v4/_examples"
	"gopkg.in/src-d/go-git.v4/plumbing"
	"gopkg.in/src-d/go-git.v4/plumbing/object"
)

// Basic example of how to list tags.
func main() {
	CheckArgs("<path>")
	path := os.Args[1]

	// We instance a new repository targeting the given path (the .git folder)
	r, err := git.PlainOpen(path)
	CheckIfError(err)

	// List all tag references, both lightweight tags and annotated tags
	Info("git show-ref --tag")

	tagrefs, err := r.Tags()
	CheckIfError(err)
	err = tagrefs.ForEach(func(t *plumbing.Reference) error {
		fmt.Println(t)
		return nil
	})
	CheckIfError(err)

	// Print each annotated tag object (lightweight tags are not included)
	Info("git tag | xargs -n 1 git cat-file -p")

	tags, err := r.TagObjects()
	CheckIfError(err)
	err = tags.ForEach(func(t *object.Tag) error {
		fmt.Println(t)
		return nil
	})
	CheckIfError(err)
}
```

orirawlings:
Here is an example of how to create lightweight tags.

```
package main

import (
	"fmt"
	"os"

	"gopkg.in/src-d/go-billy.v3/osfs"
	. "gopkg.in/src-d/go-git.v4/_examples"
	"gopkg.in/src-d/go-git.v4/plumbing"
	"gopkg.in/src-d/go-git.v4/storage/filesystem"
)

// Basic example of how create a lightweight tag
func main() {
	CheckArgs("<path> <tag> <commit>")
	path := os.Args[1]
	tag := os.Args[2]
	hash := plumbing.NewHash(os.Args[3])

	// We instance a new Storage for the path (path must be a non-bare repository)
	dot, err := osfs.New(path).Chroot(".git")
	CheckIfError(err)
[...]

PR #486, opened 2017-07-17
Repository API for creating references

orirawlings:
A few issues have been reported recently (#481 and #430) that relate to some API confusion around creating new references for a repository.

I'm curious if there is some benefit to providing some high level API in Repository to enable the creation of branches and/or tags.

I threw together this patch to gather some quick feedback on whether this is something beneficial for the project, and to see if the API makes sense. I started with `CreateBranch` for now. Tests, further documentation, and demo examples are pending.

I figured that sticking to the options pattern for the arguments to CreateBranch would give us the ability to extend this easily in the future to support other features available in the standard `git branch` tool. Such features include configuring an upstream branches, tracking branches, editing description, etc.

### DIFF
--- a/options.go
+++ b/options.go
@@ -2,6 +2,7 @@ package git
 
 import (
 	"errors"
+	"strings"
 
 	"gopkg.in/src-d/go-git.v4/config"
 	"gopkg.in/src-d/go-git.v4/plumbing"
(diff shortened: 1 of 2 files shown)


mcuadros:
This is related to https://github.com/src-d/go-git/pull/487

orirawlings:
Closing for now

PR #487, opened 2017-07-17
worktree: checkout, create branch

mcuadros:
This PR adds the options of create a branch when a Checkout is perform.
Fixes https://github.com/src-d/go-git/issues/481

### DIFF
--- a/options.go
+++ b/options.go
@@ -194,13 +194,20 @@ type SubmoduleUpdateOptions struct {
 	RecurseSubmodules SubmoduleRescursivity
 }
 
+var (
+	ErrBranchHashExclusive  = errors.New("Branch and Hash are mutually exclusive")
+	ErrCreateRequiresBranch = errors.New("Branch is mandatory when Create is used")
+)
+
 // CheckoutOptions describes how a checkout 31operation should be performed.
 type CheckoutOptions struct {
 	// Hash to be checked out, if used HEAD will in detached mode. Branch and
-	// Hash are mutual exclusive.
+	// Hash are mutually exclusive, if Create is not used.
 	Hash plumbing.Hash
 	// Branch to be checked out, if Branch and Hash are empty is set to `master`.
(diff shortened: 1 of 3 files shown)


options.go
```diff
@@ -194,20 +194,35 @@ type SubmoduleUpdateOptions struct {
 	RecurseSubmodules SubmoduleRescursivity
 }
 
+var (
+	ErrBranchHashExclusive  = errors.New("Branch and Hash are mutual exclusive")
```
orirawlings: mutual exclusive -> mutually exclusive

options.go
```diff
@@ -194,20 +194,35 @@ type SubmoduleUpdateOptions struct {
...
+	ErrCreateRequiresBranch = errors.New("Branch is mandatory when Create is used")
+)
+
 // CheckoutOptions describes how a checkout 31operation should be performed.
 type CheckoutOptions struct {
 	// Hash to be checked out, if used HEAD will in detached mode. Branch and
-	// Hash are mutual exclusive.
+	// Hash are mutual exclusive, if Create is not used.
```
orirawlings: mutual exclusive -> mutually exclusive

worktree.go
```diff
@@ -71,6 +77,29 @@ func (w *Worktree) Checkout(opts *CheckoutOptions) error {
 
 	return w.Reset(ro)
 }
+func (w *Worktree) createBranch(opts *CheckoutOptions) error {
+	_, err := w.r.Storer.Reference(opts.Branch)
```
orirawlings: Since the checking existence of and creation of the reference all delegates through `w.r`, does it make sense to push the API for creating and updating branches into `Repository` instead? Especially considering that the standard `git branch` tool allows the creation and management of branches within bare repositories (and bare `go-git` repositories have `nil` Worktrees)?

mcuadros: Yes, maybe make sense having `Repository.CreateBranch` let me thing about it.

Issue #455, opened 2017-06-29
git push new branch with depth=1


joshbetz:
Pushing a new branch after cloning with depth=1 doesn't seem to work. I get `object not found`. The problem seems to be in `revlist.Object()` -- at least that's where the error happens. As best I can tell, I think it's trying to verify too much of the tree. This same thing works with git:

```
git clone --depth 1 <repo>
git checkout -b new-branch
touch file
git add file
git commit -m "add test file"
git push
```

jucardi:
How are you doing the `checkout -b` ?

smola:
@joshbetz Thank you for reporting this. We'll look into it.

joshbetz:
@smola Thank you! Let me know if you have any questions about reproducing the issue I'm having. I tried debugging it for a while, but couldn't quite follow the logic in `remote.Push()`.

dirkaholic:
Hi there, 

experiencing the same issue. When going into the respective directory and doing the git push manually using the normal Linux git client afterwards it works. It's the same with another depth used during the clone (tried 2 or 3). So I removed Depth option from the clone command but now the clone command needs a lot more time (which is expected) but also the push after just changing a single file is running veeery long.

mcuadros:
@joshbetz how are you doing the `git checkout -b`?

joshbetz:
```
	// local is a *git.Repository
	// Create tmp branch
	ref, err := local.Head()
	new := plumbing.NewReferenceFromStrings("refs/heads/branch-name", ref.Hash().String())
	err = local.Storer.SetReference(new)
	if err != nil {
		return err
	}

	err = w.Checkout(&git.CheckoutOptions{Branch: plumbing.ReferenceName("refs/heads/branch-name")})
	if err != nil {
		return err
	}
```

dirkaholic:
Thanks for fixing the issue. What do you think when it can be pushed to the v4 branch ?

mvpmvh:
was this actually fixed? i'm getting the same issue when doing a shallow fetch, followed by a push.
The push fails with the same `unpack error: error Missing unknown <commit_sha>` 

go-git version `gopkg.in/src-d/go-git.v4 v4.13.1`

PR #489, opened 2017-07-18
repository: allow push from shallow repositories

mcuadros:
This PR allows to push from a shallow repository.
Fixes: https://github.com/src-d/go-git/issues/455


remote.go
```diff
@@ -121,6 +122,15 @@ func (r *Remote) Push(o *PushOptions) (err error) {
 		return err
 	}
 
+	stop, err := r.s.Shallow()
+	if err != nil {
+		return err
+	}
+
+	// is we have shallow we should include this as part of the objects that
```
ajnavarro: is -> if

PR #493, opened 2017-07-19
*: several windows support fixes

mcuadros:


storage/filesystem/internal/dotgit/dotgit.go
```diff
@@ -388,7 +383,15 @@ func (d *DotGit) rewritePackedRefsWithoutRef(name plumbing.ReferenceName) (err e
 		return nil
 	}
 
-	return d.fs.Rename(tmpPath, packedRefsPath)
+	if err := f.Close(); err != nil {
```
smola: Use CheckClose here for both cases? Otherwise, if closing `f` fails, `tmp` is not closed.

PR #494, opened 2017-07-19
WIP: cache: Change cache implementation from FIFO to LRU

ajnavarro:


mcuadros:
See https://github.com/src-d/go-git/pull/514

Issue #490, opened 2017-07-18
Missing fetch options: `--force`, `--no-tags` and `--prune`


lupine:
Encouraged by the success of go-git in our elasticsearch indexer (https://gitlab.com/gitlab-org/gitlab-elasticsearch-indexer), I was hoping to use it in our repository mirroring code, as I'm adding support for SSH public key auth (https://gitlab.com/gitlab-org/gitlab-shell/issues/95 https://gitlab.com/gitlab-org/gitlab-shell/merge_requests/147). This is a simple wrapper around `git fetch`, but I don't seem able to replicate several aspects of existing behaviour with go-git:

* `git fetch --force` seems to be the default behaviour? I can't see any mention of the check `--force` disables in `updateReferenceStorerIfNeeded` or its callers, at any rate.

* `git fetch --tags` is `FetchOptions.Tags = git.AllTags`, but what about `git fetch --no-tags`? There doesn't seem to be a `NoTags` constant

* `git fetch --prune` seems unsupported. In fact, all aspects of `git gc` seem unsupported - although this could be worked around.

I'd be happy to submit PRs to implement the first two options if they'd be considered for merge.

mcuadros:
Great news!

About the missing features: 
> git fetch --force seems to be the default behaviour? I can't see any mention of the check --force disables in updateReferenceStorerIfNeeded or its callers, at any rate.

Yep is the default behavior will be great having it as git does.

> git fetch --tags is FetchOptions.Tags = git.AllTags, but what about git fetch --no-tags? There doesn't seem to be a NoTags constant

Great to have it in a PR

> git fetch --prune seems unsupported. In fact, all aspects of git gc seem unsupported - although this could be worked around.

Yep we don't have gc, but at least can be implemented for the references.

lupine:
@mcuadros for `force`, are you willing to change the default behaviour in v4 so that the default becomes *not* to force? Or would you see that as a breaking change?

It affects whether `FetchOptions` (and `PullOptions`) get a `Force` or `NoForce` member.

lupine:
`git fetch` without `--force` will update all the refs it can, rather than failing early. Worth preserving?

strib:
@lupine: I'm very interested in having a non-force mode as well.  I'm curious what you're plan is regarding locking, since presumably you'll need to lock the reference between when you check it and when you set it.  Will this be a new ability of the `Storer` interface?  Ideally it would be something that can be abstracted out, as different storage layers might need to do different things there.

[...]

PR #496, opened 2017-07-22
Implement a NoTags mode for fetch that mimics git fetch --no-tags

lupine:
See https://github.com/src-d/go-git/issues/490


lupine:
@mcuadros here's part one. The test actually passed without me needing to change the logic at all, but I think it's right (the relevant line being https://github.com/src-d/go-git/blob/master/remote.go#L413)

lupine:
Unsure about the test failures. Are they related?

mcuadros:
Great, about the CI don't worry, since a while, some test are failing randomly in Travis, looks like the machines are now slower than before and takes more time in start a service.

PR #497, opened 2017-07-22
Add a Force flag to Fetch and Clone, mimicking the command-line client

lupine:
See #490

### DIFF
--- a/remote.go
+++ b/remote.go
@@ -24,6 +24,7 @@ import (
 var (
 	NoErrAlreadyUpToDate     = errors.New("already up-to-date")
 	ErrDeleteRefNotSupported = errors.New("server does not support delete-refs")
+	ErrForceNeeded           = errors.New("some refs were not updated")
 )
 
 // Remote represents a connection to a remote repository.
@@ -223,7 +224,7 @@ func (r *Remote) fetch(o *FetchOptions) (storer.ReferenceStorer, error) {
 		}
 	}
 
-	updated, err := r.updateLocalReferenceStorage(o.RefSpecs, refs, remoteRefs)
+	updated, err := r.updateLocalReferenceStorage(o.RefSpecs, refs, remoteRefs, o.Force)
 	if err != nil {
 		return nil, err
 	}
@@ -589,8 +590,11 @@ func buildSidebandIfSupported(l *capability.List, reader io.Reader, p sideband.P
 func (r *Remote) updateLocalReferenceStorage(
 	specs []config.RefSpec,
 	fetchedRefs, remoteRefs memory.ReferenceStorage,
+	force bool,
 ) (updated bool, err error) {
 	isWildcard := true
+	forceNeeded := false
+
 	for _, spec := range specs {
 		if !spec.IsWildcard() {
 			isWildcard = false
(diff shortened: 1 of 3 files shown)


lupine:
@mcuadros this is currently lacking tests but you get the basic idea. WDYT?

remote.go
```diff
@@ -580,6 +584,18 @@ func (r *Remote) updateLocalReferenceStorage(
 
 			new := plumbing.NewHashReference(spec.Dst(ref.Name()), ref.Hash())
 
+			if ref.IsBranch() && !force {
+				ff, err := isFastForward(r.s, ref.Hash(), ref.Hash())
```
lupine: Need to find the hash of the local branch

mcuadros:
Looks good!

lupine:
I'm having a little trouble with the tests, actually. Consider:

```go
func (s *RemoteSuite) TestFetchNoForce(c *C) {
       r := newRemote(memory.NewStorage(), &config.RemoteConfig{
               URL: s.GetBasicLocalRepositoryURL(),
       })

       // Can be fast-forwarded
       r.s.SetReference(plumbing.NewReferenceFromStrings(
               "refs/heads/master", "918c48b83bd081e863dbe1b80f8998f058cd8294",
       ))

       s.testFetch(c, r, &FetchOptions{
               RefSpecs: []config.RefSpec{
                       config.RefSpec("+refs/heads/master:refs/remotes/origin/master"),
               },
       }, []*plumbing.Reference{
               plumbing.NewReferenceFromStrings("refs/heads/master", "f7b877701fbf855b44c0a9e86f3fdce2c298b07f"),
               plumbing.NewReferenceFromStrings("refs/remotes/origin/master", "f7b877701fbf855b44c0a9e86f3fdce2c298b07f"),
       })
}
```

I get the error:

```
remote_test.go:193:
    c.Assert(exp.String(), Equals, r.String())
... obtained string = "f7b877701fbf855b44c0a9e86f3fdce2c298b07f refs/heads/master"
... expected string = "918c48b83bd081e863dbe1b80f8998f058cd8294 refs/heads/master"
```

I want to set up the local ref "master" such that the fetch has to update it. I can then set the ref to commits that are (or are not) ancestors and exercise the behaviour that way, but I've not managed to get it working just yet.

I'll look into it again on Thursday, but in the meantime, any tips appreciated!

mcuadros:
You were lucky? Can you elaborate a little more, I don't understand the goal.

mcuadros:
@lupine any update? If not I will try to take a look to it, since we want to release the v4, and this should be included.

lupine:
Sorry @mcuadros, the tests and the safe locking behaviour both prevented me from finishing this, and in the end we went for a "wrap the SSH client" approach over at GitLab.

There's also the question of respecting "+" in refspecs.

If you can finish this off, I'd be super-grateful!

mcuadros:
Over the go-git SSH client?

lupine:
That's right :/ we're doing https://gitlab.com/gitlab-org/gitlab-shell/blob/master/lib/gitlab_projects.rb#L432 instead.

I'd love to get it cleaned up to be more like https://gitlab.com/gitlab-org/gitlab-shell/merge_requests/147 but couldn't quite manage it.

mcuadros:
https://github.com/src-d/go-git/issues/490#issuecomment-327136452

PR #499, opened 2017-07-22
move Repository.Pull to Worktree.Pull

mcuadros:
This PR moves the method `Pull` from the `Repository` to `Worktree`, this method should be there because is only useful when the repository is not bare. This is the behavior at `cgit`:

```
> git clone git clone git@github.com:git-fixtures/basic.git --bare
....     
> git pull                            
fatal: This operation must be run in a work tree
````

### DIFF
--- a/_examples/remotes/main.go
+++ b/_examples/remotes/main.go
@@ -42,9 +42,9 @@ func main() {
 		fmt.Println(r)
 	}
 
-	// Pull using the create repository
-	Info("git pull example")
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: _examples/remotes/main.go [hunk 1]
```diff
@@ -43,8 +43,8 @@ func main() {
 	}
 
 	// Pull using the create repository
```
smola: s/Pull/Fetch

Issue #500, opened 2017-07-24
Push should work with multiple URLs


smola:
`git push` can push to multiple URLs for the same remote if there is more than one:
http://marc.info/?l=git&m=116231242118202&w=2

PR #501, opened 2017-07-24
config: multiple values in RemoteConfig (URLs and Fetch)

smola:
* Change `URL string` to `URL []string` in `RemoteConfig`, since
  git allows multiple URLs per remote. See:
  http://marc.info/?l=git&m=116231242118202&w=2

* Fix marshalling of multiple fetch refspecs.

### DIFF
--- a/remote.go
+++ b/remote.go
@@ -43,8 +43,11 @@ func (r *Remote) Config() *config.RemoteConfig {
 }
 
 func (r *Remote) String() string {
-	fetch := r.c.URL
-	push := r.c.URL
+	var fetch, push string
+	if len(r.c.URLs) > 0 {
+		fetch = r.c.URLs[0]
+		push = r.c.URLs[0]
+	}
 
 	return fmt.Sprintf("%s\t%s (fetch)\n%[1]s\t%[3]s (push)", r.c.Name, fetch, push)
 }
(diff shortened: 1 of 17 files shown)


smola:
This will be useful in `src-d/borges` to store all endpoints for each repository. It will be required to implement push to multiple URLs (https://github.com/src-d/go-git/issues/500) at some point too.

bzz:
Looks like rebase might be in order.

remote.go
```diff
@@ -42,8 +42,8 @@ func (r *Remote) Config() *config.RemoteConfig {
 }
 
 func (r *Remote) String() string {
-	fetch := r.c.URL
-	push := r.c.URL
+	fetch := r.c.URLs[0]
```
ajnavarro: this can throw panics

remote_test.go
```diff
@@ -27,33 +27,33 @@ type RemoteSuite struct {
 var _ = Suite(&RemoteSuite{})
 
 func (s *RemoteSuite) TestFetchInvalidEndpoint(c *C) {
-	r := newRemote(nil, &config.RemoteConfig{Name: "foo", URL: "http://\\"})
+	r := newRemote(nil, &config.RemoteConfig{Name: "foo", URLs: []string{"http://\\"}})
```
ajnavarro: Add a test with a RemoteConfig with a nil slice of URLs. It will panic

PR #503, opened 2017-07-24
Add TokenAuth as an implementation of AuthMethod

strin:


plumbing/transport/http/common.go
```diff
@@ -145,6 +145,25 @@ type AuthMethod interface {
...
+func NewTokenAuth(token string) *TokenAuth {
+	return &TokenAuth{token}
+}
+
+func (a *TokenAuth) setAuth(r *http.Request) {
+	if a == nil {
+		return
+	}
+
+	r.Header.Set("Authorization", a.token)
```
mcuadros: To be a proper Bearer authentication you must to change the to `fmt.Sprintf("Bearer %s", a.token)`

plumbing/transport/http/common.go
```diff
@@ -145,6 +145,25 @@ type AuthMethod interface {
 	setAuth(r *http.Request)
 }
 
+// TokenAuthMethod is concrete implementation of common.AuthMethod for HTTP services
+// Allow Bearer Token used in git authentication.
+type TokenAuth struct {
```
mcuadros: Call it `BearerAuth` instead of a generic name

plumbing/transport/http/common.go
```diff
@@ -145,6 +145,25 @@ type AuthMethod interface {
...
+	token string
+}
+
+// NewTokenAuth returns a tokenAuth on the given authrorization token.
+func NewTokenAuth(token string) *TokenAuth {
```
mcuadros: Can you add some tests?

mcuadros:
Please ro-open once the are done.

Issue #505, opened 2017-07-25
Very slow for large repos


strin:
```
options := git.PushOptions{
		RemoteName: remoteName,
		Auth:       auth,
		RefSpecs: []config.RefSpec{
			config.RefSpec(fmt.Sprintf("refs/heads/%s:refs/heads/%s", hash, hash)),
		},
	}

err = repo.GitRepo.Push(&options)
```

The time for second push should be significantly less than the first push. But it seems `go-git` is pushing everything again.

strin:
To test, use this repo: https://github.com/dummy-ai/models

smola:
@strin Thank you for reporting this. Indeed, a lot of objects were being pushed when they shouldn't. This should be fixed by PR https://github.com/src-d/go-git/pull/516

PR #506, opened 2017-07-25
Dedup hashes that have already been tracked by remote refs

strin:
An attempt to solve issue #505 

Based on my educated guess about the way git works. Need someone more experienced to double check this.


ajnavarro:
Can you check the code again? It appears to do not compile.

Anyways, we are working in better ways to improve this, like add a cache at ObjectStorer level to improve speed retrieving single objects, and change the way that revlist is obtaining objects to do at the same way that git does.

PR #514, opened 2017-07-27
cache: reuse object cache for delta resolution, use LRU policy

smola:


plumbing/cache/object_lru.go
```diff
@@ -0,0 +1,77 @@
...
+			return
+		}
+	}
+
+	ee := c.ll.PushFront(obj)
+	c.cache[key] = ee
+	c.actualSize += objSize
+}
+
+// Get returns an object by its hash. If the object is not found in the cache.
```
ajnavarro: If the object is not found in the cache, what happens?

plumbing/cache/object_lru.go
```diff
@@ -0,0 +1,77 @@
...
+	ll         *list.List
+	cache      map[interface{}]*list.Element
+}
+
+func NewObjectLRU(maxSize FileSize) *ObjectLRU {
+	return &ObjectLRU{MaxSize: maxSize}
+}
+
+// Put puts an object into the cache.
+func (c *ObjectLRU) Put(obj plumbing.EncodedObject) {
```
ajnavarro: Can you explain more what will happen if the object is already into the cache?

storage/filesystem/object.go
```diff
@@ -17,13 +18,17 @@ import (
...
+	DeltaBaseCache cache.Object
+
 	dir   *dotgit.DotGit
 	index map[plumbing.Hash]*packfile.Index
 }
 
 func newObjectStorage(dir *dotgit.DotGit) (ObjectStorage, error) {
 	s := ObjectStorage{
-		dir: dir,
+		DeltaBaseCache: cache.NewObjectLRU(92 * cache.MiByte),
```
ajnavarro: can we put the `92 * cache.MiByte` as a constant?

PR #516, opened 2017-07-27
revlist: ignore all objects reachable from ignored objects

smola:
Usually we call revlist.Objects ignoring a set of commits.
This is not enough to ignore everything reachable from such
set, so we first get all objects reachable from the ignored
set and then walk the tree again ignoring that new set.

Fixes https://github.com/src-d/go-git/issues/505

### DIFF
--- a/plumbing/revlist/revlist.go
+++ b/plumbing/revlist/revlist.go
@@ -16,7 +16,26 @@ import (
 // the reachable objects from the given objects. Ignore param are object hashes
 // that we want to ignore on the result. All that objects must be accessible
 // from the object storer.
-func Objects(s storer.EncodedObjectStorer, objects, ignore []plumbing.Hash) ([]plumbing.Hash, error) {
+func Objects(
+	s storer.EncodedObjectStorer,
+	objs,
+	ignore []plumbing.Hash,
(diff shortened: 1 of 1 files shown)


plumbing/revlist/revlist.go
```diff
@@ -16,7 +16,26 @@ import (
 // the reachable objects from the given objects. Ignore param are object hashes
 // that we want to ignore on the result. All that objects must be accessible
 // from the object storer.
-func Objects(s storer.EncodedObjectStorer, objects, ignore []plumbing.Hash) ([]plumbing.Hash, error) {
+func Objects(
+	s storer.EncodedObjectStorer,
+	objs []plumbing.Hash,
+	ignore []plumbing.Hash) ([]plumbing.Hash, error) {
```
mcuadros: why are you duplicating the types?

smola: They are not duplicated. The public signature is still the same. However, I moved most of the logic of `Objects` to a private implementation `objects` which can be called with further arguments. That one is used from `Objects` and it's called twice.

smola: Got it ;)

PR #534, opened 2017-08-03
plumbing: object, commit.Parent() method

josharian:
First parents are somewhat special in git.
There's even a --first-parent flag to 'git log'.

Add a helper method to look them up.
This avoids boilerplate and spares the client from
having to arrange for a handle to the Storer,
which is stored in the unexported field Commit.s.

### DIFF
--- a/plumbing/object/commit.go
+++ b/plumbing/object/commit.go
@@ -3,6 +3,7 @@ package object
 import (
 	"bufio"
 	"bytes"
+	"errors"
 	"fmt"
 	"io"
 	"strings"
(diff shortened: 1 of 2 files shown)


plumbing/object/commit.go
```diff
@@ -91,6 +92,16 @@ func (c *Commit) NumParents() int {
 	return len(c.ParentHashes)
 }
 
+var ErrNoParents = errors.New("commit has no parents")
+
+// FirstParent returns the first parent of c.
+func (c *Commit) FirstParent() (*Commit, error) {
```
mcuadros: can you create create test or this?

mcuadros:
@ajnavarro @orirawlings what do you thing about this?

orirawlings:
I agree it needs a test, but I think it can be useful.

Perhaps it would be more useful to generalize further:

```
// Parent returns the ith parent of c. 
func (c *Commit) Parent(i int) (*Commit, error) {
```

ajnavarro:
`FirstParent()` appears good to me, it has some use cases. Into the method documentation I would add what happens if the commit does not have parents.

mcuadros:
@orirawlings this is the signature from "git2go" and I must to say that for me wasn't clear that the first commit has any special meaning. But it's because doesn't have it, the order of the parents is something relevant when you are flattering the tree (like in `git log`).

So yes, I prefer the signature 
```
func (c *Commit) Parent(i int) (*Commit, error) {
```
@joshbetz can you update the PR with the request? Thanks!

joshbetz:
I think you mean @josharian :)

mcuadros:
Yup, sorry.

josharian:
Thanks for the review. Will add a test and update docs a bit later; distracted by other matters at the moment.

PR #538, opened 2017-08-07
Avoid using user.Current()

### LINKED ISSUE #537: user.Current() causes error when cross compiled
I build go-git on my MacBook Pro and use the binary on Linux machine with `GOOS=linux GOARCH=amd64 CGO_ENABLED=0` flags. This makes go-git to make the following error when used on Linux machine.

`user: Current not implemented on linux/amd64`

I've seen the same issues in other repositories e.g. https://github.com/andreaskoch/allmark/commit/09b235aa80cef3b696eabd9afe90acbcff909154

kimh:
`user.Current()` causes `Current not implemented` error when
cross compiled. See https://github.com/golang/go/issues/6376

Fixes #537

### DIFF
--- a/plumbing/transport/ssh/auth_method.go
+++ b/plumbing/transport/ssh/auth_method.go
@@ -3,6 +3,7 @@ package ssh
 import (
 	"crypto/x509"
 	"encoding/pem"
+	"errors"
 	"fmt"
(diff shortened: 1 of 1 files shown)


mcuadros:
Thanks!

PR #544, opened 2017-08-11
fix race condition on ObjectLRU

erizocosmico:
With concurrent usage, it could lead to a `concurrent map access` panic on runtime.

Sadly, the test only adds value with `-race` enabled. If you have any idea how to get that panic on the test, it would improve its robustness.

### DIFF
--- a/plumbing/cache/object_lru.go
+++ b/plumbing/cache/object_lru.go
@@ -2,6 +2,7 @@ package cache
 
 import (
 	"container/list"
+	"sync"
 
 	"gopkg.in/src-d/go-git.v4/plumbing"
 )
@@ -14,6 +15,7 @@ type ObjectLRU struct {
 	actualSize FileSize
 	ll         *list.List
 	cache      map[interface{}]*list.Element
+	mut        sync.Mutex
 }
 
 // NewObjectLRU creates a new ObjectLRU with the given maximum size. The maximum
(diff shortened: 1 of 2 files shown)


plumbing/cache/object_test.go
```diff
@@ -67,6 +69,20 @@ func (s *ObjectSuite) TestClear(c *C) {
 	c.Assert(obj, IsNil)
 }
 
+func (s *ObjectSuite) TestConcurrentAccess(c *C) {
+	var wg sync.WaitGroup
+
+	for i := 0; i < 1000; i++ {
+		wg.Add(1)
+		go func(i int) {
+			s.c.Put(newObject(fmt.Sprint(i), FileSize(i)))
```
orirawlings: Do we also want to exercise the other operations that require synchronization? Like Get and Clear?

erizocosmico:
@orirawlings amended with usage of `Clear` and `Get` as well

plumbing/cache/object_lru.go
```diff
@@ -67,6 +72,9 @@ func (c *ObjectLRU) Put(obj plumbing.EncodedObject) {
 // Get returns an object by its hash. It marks the object as used. If the object
 // is not in the cache, (nil, false) will be returned.
 func (c *ObjectLRU) Get(k plumbing.Hash) (plumbing.EncodedObject, bool) {
+	c.mut.Lock()
```
mcuadros: Maybe we can use a RWLock? as suggested at: https://blog.golang.org/go-maps-in-action

erizocosmico: `Get` mutates the state in `c.ll.MoveToFront(ee)`, that's why I used a `Mutex` instead of a `RWMutex`

PR #548, opened 2017-08-16
Added support for http user basic authentication

moisespsena:
Added support for http user basic authentication


_examples/custom_http_auth/main.go
```diff
@@ -0,0 +1,59 @@
...
+		Timeout: 15 * time.Second,
+
+		// don't follow redirect
+		CheckRedirect: func(req *http.Request, via []*http.Request) error {
+			return http.ErrUseLastResponse
+		},
+	}
+
+	// Override http(s) default protocol to use our custom client
+	client.InstallProtocol("https", githttp.NewClient(customClient))
```
orirawlings: Apologies in advance if this is merely my own ignorance, but is there a specific reason we need to include this in the example code? It seems to me like encouraging people to disable SSL verification and redirects isn't good general practice. I think this is more or less copy/pasted from the existing `custom_http` example but maybe doesn't make sense here.

For this example code I think it makes more sense to just focus on demonstrating the basic http auth usage in isolation.

_examples/custom_http_auth/main.go
```diff
@@ -0,0 +1,59 @@
+package main
```
orirawlings: I'm thinking we might want to rename this example from `custom_http_auth` to `http_basic_auth`

orirawlings: We should add a link to the example in the `_examples/README.md`

_examples/custom_http_auth/main.go
```diff
@@ -0,0 +1,59 @@
...
+	"gopkg.in/src-d/go-git.v4"
+	. "gopkg.in/src-d/go-git.v4/_examples"
+	"gopkg.in/src-d/go-git.v4/plumbing/transport/client"
+	githttp "gopkg.in/src-d/go-git.v4/plumbing/transport/http"
+	"gopkg.in/src-d/go-git.v4/storage/memory"
+)
+
+// Here is an example to configure http client according to our own needs.
+func main() {
+	CheckArgs("<url>", "<username>", "<password>")
```
orirawlings: To get the tests for the example code to pass, you'll need to set some arguments in the test file: https://github.com/src-d/go-git/blob/master/_examples/common_test.go#L17

```
=== RUN   TestExamples/custom_http_auth
Usage: /tmp/go-build169800819/command-line-arguments/_obj/exe/main <url> <username> <password>
exit status 1
```

_examples/custom_http_auth/main.go
moisespsena: It's a good idea.

moisespsena: Excelent!

_examples/custom_http_auth/main.go
moisespsena: Ok.

_examples/custom_http_auth/main.go
moisespsena: I agree.

plumbing/transport/http/common.go
```diff
@@ -145,6 +145,27 @@ type AuthMethod interface {
...
+func (a *BasicAuthMethod) String() string {
+	return a.auth.String()
+}
+
+func (a *BasicAuthMethod) setAuth(r *http.Request)  {
+	a.auth.setAuth(r)
+}
+
+func NewBasicAuthMethod(username, password string) *BasicAuthMethod {
+	return &BasicAuthMethod{auth:NewBasicAuth( "moisespsena", "ghis5509")}
```
mcuadros: err, this is a user and a password...

mcuadros:
Can you explain the goals of this PR, why is not enough [`http.BasicAuth`](https://godoc.org/gopkg.in/src-d/go-git.v4/plumbing/transport/http#BasicAuth)

plumbing/transport/http/common.go
moisespsena: Thanks.

mcuadros:
close due to inactivity

Issue #152, opened 2016-11-30
packp: AdvRefs is not encoded correctly when missing Head pointer


smola:
From https://github.com/git/git/blob/master/Documentation/technical/pack-protocol.txt
```
If HEAD is a valid ref, HEAD MUST appear as the first advertised
ref.  If HEAD is not a valid ref, HEAD MUST NOT appear in the
advertisement list at all, but other refs may still appear.
```

So HEAD, if present, should be the first reference. Otherwise, the first reference is encoded as usual.

This is different from our current implementation, which encodes a zero hash instead:

```
func (s *AdvRefsEncodeSuite) TestRefs(c *C) {
	references := map[string]plumbing.Hash{
		"refs/heads/master":      plumbing.NewHash("a6930aaee06755d1bdcfd943fbf614e4d92bb0c7"),
		"refs/tags/v2.6.12-tree": plumbing.NewHash("1111111111111111111111111111111111111111"),
		"refs/tags/v2.7.13-tree": plumbing.NewHash("3333333333333333333333333333333333333333"),
		"refs/tags/v2.6.13-tree": plumbing.NewHash("2222222222222222222222222222222222222222"),
		"refs/tags/v2.6.11-tree": plumbing.NewHash("5dc01c595e6c6ec9ccda4f6f69c131c0dd945f8c"),
	}
	ar := &AdvRefs{
		References: references,
	}

	expected := pktlines(c,
		"0000000000000000000000000000000000000000 capabilities^{}\x00\n",
		"a6930aaee06755d1bdcfd943fbf614e4d92bb0c7 refs/heads/master\n",
		"5dc01c595e6c6ec9ccda4f6f69c131c0dd945f8c refs/tags/v2.6.11-tree\n",
		"1111111111111111111111111111111111111111 refs/tags/v2.6.12-tree\n",
		"2222222222222222222222222222222222222222 refs/tags/v2.6.13-tree\n",
		"3333333333333333333333333333333333333333 refs/tags/v2.7.13-tree\n",
		pktline.FlushString,
	)

	testEncode(c, ar, expected)
}
```

mcuadros:
@smola still happening?

mcarmonaa:
PR https://github.com/src-d/go-git/pull/552 solves it

orirawlings:
Can this be closed?

PR #552, opened 2017-08-18
packp: fixed encoding when HEAD is not a valid ref

mcarmonaa:
It solves issue https://github.com/src-d/go-git/issues/152


mcuadros:
Awesome!

ajnavarro:
good @mcarmonaa !!

PR #554, opened 2017-08-22
plumbing: use LookPath instead of Stat to fix Windows executables

strib:
When git-core isn't in the user's PATH, we need to use `LookPath` to
verify the existence of the executable, rather than `os.Stat`, so that
on Windows it will search for files with executable suffixes.


strib:
@mcuadros @erizocosmico this is ready for review/merge too, if you get the chance.  The file transport protocol is currently broken on Windows without this.

PR #558, opened 2017-08-24
repository: Resolve commit when cloning annotated tag, fixes #557

### LINKED ISSUE #557: Can't clone annotated tags with `ReferenceName` in Clone
Hi! Great library. 

The problem is that given an annotated tag ref, I get an "object not found" error:

```go
repo, err := git.PlainClone(dir, false, &git.CloneOptions{
	URL: "...",
	ReferenceName: plumbing.ReferenceName("refs/tags/v1.0.0"),
})
```

Where `v1.0.0` is an **annotated tag** (not a lightweight one). 

I'm not sure if this is working as intended or not, but I'd expect the library to recognize the commit the annotated tag is pointing to and load that.

Reproducing this is extremely easy:

  1. Create a new git repo
  2. Commit a single file so its not empty
  3. `git tag -a "v1.0.0" -m "1.0.0"`
  4. Push
  5. Run the code above using that repo's address. It will fail.

Thanks!

orirawlings:
Fixes #557

### DIFF
--- a/repository.go
+++ b/repository.go
@@ -30,6 +30,7 @@ var (
 	ErrRemoteExists            = errors.New("remote already exists	")
 	ErrWorktreeNotProvided     = errors.New("worktree should be provided")
 	ErrIsBareRepository        = errors.New("worktree not available in a bare repository")
+	ErrUnableToResolveCommit   = errors.New("unable to resolve commit")
 )
 
 // Repository represents a git repository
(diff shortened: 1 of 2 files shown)


orirawlings:
Tests are pending still, but wanted to make sure my general approach made sense. I still have to dig into how the current tests for the repository Clone* methods are implemented.

repository.go
```diff
@@ -400,6 +400,25 @@ func (r *Repository) DeleteRemote(name string) error {
...
+	case plumbing.TagObject:
+		t, err := object.DecodeTag(r.Storer, obj)
+		if err != nil {
+			return plumbing.ZeroHash, err
+		}
+		return r.resolveToCommitHash(t.Target)
+	case plumbing.CommitObject:
+		return h, nil
+	default:
+		return plumbing.ZeroHash, errors.New("Unable to resolve to commit")
```
orirawlings: Whoops. Forgot to extract an error constant here.

orirawlings:
I've added a failing test. My current strategy prevents an error from occurring, but it leaves the HEAD ref pointed to the tag object rather the underlying commit object. I'll continue working on a more comprehensive fix.

orirawlings:
I fixed the PR to ensure that the HEAD ref is left pointing to the commit object.

PR #564, opened 2017-08-25
plumbing: use `seen` map in tree walker

strib:
This helps avoid iterating down the same trees for every commit.  For a big-ish repo with 35K objects (17K commits), this reduced the time for calling `revlist.Objects` during a push (with 0 hashes to ignore) from more than ten minutes to less than a minute.

### DIFF
--- a/plumbing/object/tree.go
+++ b/plumbing/object/tree.go
@@ -297,9 +297,10 @@ func (iter *treeEntryIter) Next() (TreeEntry, error) {
 
 // TreeWalker provides a means of walking through all of the entries in a Tree.
 type TreeWalker struct {
-	stack     []treeEntryIter
+	stack     []*treeEntryIter
 	base      string
 	recursive bool
+	seen      map[plumbing.Hash]bool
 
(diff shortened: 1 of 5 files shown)


plumbing/object/tree.go
```diff
@@ -297,9 +297,10 @@ func (iter *treeEntryIter) Next() (TreeEntry, error) {
 
 // TreeWalker provides a means of walking through all of the entries in a Tree.
 type TreeWalker struct {
-	stack     []treeEntryIter
+	stack     []*treeEntryIter
```
strib: This seems safer to me; I was worried things that accessed the stack might end up calling `Next` on a copy, rather than the real object.  But the way the stack is used now, I don't think it's a real bug.  If you want me to revert it, let me know.

erizocosmico: I personally don't mind the change, the struct is just a pointer and an int so copies are not expensive and it might be faster due to locality but I don't think it matters much.

strib:
Thanks for looking @erizocosmico.  Now there's a simple test.

PR #565, opened 2017-08-25
plumbing: use sliding window in delta calculations, like git CL

strib:
This sets a default sliding window of 10 for the delta calculation, just like git command line:
https://git-scm.com/docs/git-pack-objects#git-pack-objects---windowltngt

For a big-ish repo with 35K objects (17K commits), this reduced the time for calling `deltaSelection.walk` during a push from more than 14 minutes to about a minute.

### DIFF
--- a/plumbing/format/packfile/delta_selector.go
+++ b/plumbing/format/packfile/delta_selector.go
@@ -8,6 +8,9 @@ import (
 )
 
 const (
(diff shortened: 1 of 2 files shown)


mcuadros:
Any chance to cover this with a test?

strib:
Thanks @mcuadros and @erizocosmico, added a simple test.  Ready for merge, I think.

PR #567, opened 2017-08-27
remote: avoid expensive revlist operation when only deleting refs

strib:


mcuadros:
Any chance to cover this with a test?

strib:
@mcuadros: any idea what kind of test I could write for this?  My best idea would be to use mocks somehow to make sure that the revlist method is never called, but I don't see precedent for mocks in the test right now, and don't have time to build that infrastructure out myself.

I guess I could just test for correctness when all refspecs are deletes, rather than testing for performance, but that test would pass both with and without this change.

strib:
I have no clue why this is reporting such a large test coverage difference.  I'll try to throw together the simple correctness test and see if that helps.

Issue #415, opened 2017-06-08
worktree Checkout and Reset call to fetch operations on submodules


smola:
`checkoutChangeSubmodule` calls to submodule's update operation, which ends up calling a fetch.
This is called from `Worktree.Reset`, which is called by `Worktree.Checkout`.
Both Checkout and Reset are local operations that do not do any networking.

This was uncovered by this: https://travis-ci.org/src-d/go-git/jobs/240748439

PR #572, opened 2017-08-31
Worktree.Reset refactor and Soft, Merge, Hard and Mixed modes

### LINKED ISSUE #435: Worktree.Checkout:Force bug?
1.I clone a test Repository to /Users/snail/codes/gotest (current branch is master)
2.I delete a file "/Users/snail/codes/gotest/index.php"
3.use git command,git checkout master -f 
4. nothing error showed and index.php returned back,i hope so.
but if is use go-git do step 3 ,i got an error :remove /Users/snail/codes/gotest/index.php: no such file or directory
code snippet:
err = w.Checkout(&git.CheckoutOptions{
		Branch: plumbing.ReferenceName("refs/heads/master"),
		Force:  true,
	})
fmt.Println(err)
is it a bug? I confused .

mcuadros:
Fixes #435  #415

### DIFF
--- a/submodule.go
+++ b/submodule.go
@@ -62,14 +62,17 @@ func (s *Submodule) Status() (*SubmoduleStatus, error) {
 }
 
 func (s *Submodule) status(idx *index.Index) (*SubmoduleStatus, error) {
+	status := &SubmoduleStatus{
+		Path: s.c.Path,
+	}
+
 	e, err := idx.Entry(s.c.Path)
-	if err != nil {
+	if err != nil && err != index.ErrEntryNotFound {
 		return nil, err
 	}
 
-	status := &SubmoduleStatus{
-		Path:     s.c.Path,
-		Expected: e.Hash,
+	if e != nil {
+		status.Expected = e.Hash
 	}
 
 	if !s.initialized {
--- a/worktree.go
+++ b/worktree.go
@@ -107,7 +107,10 @@ func (w *Worktree) PullContext(ctx context.Context, o *PullOptions) error {
 		return err
 	}
 
-	if err := w.Reset(&ResetOptions{Commit: ref.Hash()}); err != nil {
+	if err := w.Reset(&ResetOptions{
+		Mode:   MergeReset,
+		Commit: ref.Hash(),
+	}); err != nil {
 		return err
 	}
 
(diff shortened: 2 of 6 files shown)


submodule.go
```diff
@@ -90,7 +93,7 @@ func (s *Submodule) status(idx *index.Index) (*SubmoduleStatus, error) {
 		err = nil
 	}
 
-	return status, err
+	return status, nil
```
erizocosmico: Are we intentionally discarding the error here? If so, the previous if can be removed.

erizocosmico: Are we intentionally discarding the error here? If so, the previous if can be removed.

worktree.go
```diff
@@ -270,65 +273,99 @@ func (w *Worktree) Reset(opts *ResetOptions) error {
...
+			name = ch.To.String()
+			e, err = t.FindEntry(name)
+			if err != nil {
+				return err
+			}
+		case merkletrie.Delete:
+			name = ch.From.String()
+		}
+
+		_, _ = idx.Remove(name)
```
erizocosmico: can this operation fail? If so, should we leave it unhandled?

worktree.go
```diff
@@ -424,9 +485,7 @@ func (w *Worktree) checkoutChangeRegularFile(name string,
 ) error {
 	switch a {
 	case merkletrie.Modify:
-		if err := w.rmIndexFromFile(name, idx); err != nil {
-			return err
-		}
+		_, _ = idx.Remove(name)
```
erizocosmico: same, should we ignore this error?

worktree.go
```diff
@@ -503,6 +556,7 @@ func (w *Worktree) checkoutFileSymlink(f *object.File) (err error) {
 }
 
 func (w *Worktree) addIndexFromTreeEntry(name string, f *object.TreeEntry, idx *index.Index) error {
+	_, _ = idx.Remove(name)
```
erizocosmico: same

worktree.go
```diff
@@ -513,6 +567,7 @@ func (w *Worktree) addIndexFromTreeEntry(name string, f *object.TreeEntry, idx *
 }
 
 func (w *Worktree) addIndexFromFile(name string, h plumbing.Hash, idx *index.Index) error {
+	_, _ = idx.Remove(name)
```
erizocosmico: same

worktree.go
mcuadros: the only error returned by Remove, is `ErrEntryNotFound` and we don't care about it.

PR #577, opened 2017-09-04
Worktree.Add: Support Add deleted files, fixes #571

### LINKED ISSUE #571: git commit -a should not refer to the storage but to the filesystem
Hi,

[x] feature request

**How would the same operation be performed using git from the command line terminal?**
\# bar.txt is tracked by git
rm -f bar.txt
git commit -am "Delete that file"

**Current behavior:**
Currently, removing a file in a repository without using go-git will break the git commit All functionalities.  error = Lstat error no such file or directory

**Expected behavior:**
worktree_commit.go, Commit.
```go
if opts.All {
		if err := w.autoAddModifiedAndDeleted(); err != nil {
			return plumbing.ZeroHash, err
		}
}
```

```go
func (w *Worktree) autoAddModifiedAndDeleted() error {
	s, err := w.Status()
	if err != nil {
		return err
	}

	for path, fs := range s {
		if fs.Worktree != Modified && fs.Worktree != Deleted {
			continue
		}

		if _, err := w.Add(path); err != nil {
			return err
		}

	}

	return nil
}
```

If the file does not exist anymore Add will return an Lstat error, which is expected because a deletion can be considered a modification. Hence it should be covered by a Commit.  

So my guess is, if not found the StatusCode of the FileStatus should be modified to  Deleted.

**go-git version:**
~4.0.0-rc9

I'll be happy to submit a PR if you want. 

If you have a better idea,I will be happy to read it !

grunenwflorian:
In order to fix the bug mentioned in the issue, I modified the Add method of Worktree to handle the case where a file is removed but is still in the repository.

My fix just ensure that the repo and the filesystem are always in sync.

I also found what I think to be a false positive in the TestFilenameNormalization test. 
The wrong file were added.

### DIFF
--- a/worktree_status.go
+++ b/worktree_status.go
@@ -236,7 +236,7 @@ func (w *Worktree) Add(path string) (plumbing.Hash, error) {
 
 	h, err := w.copyFileToStorage(path)
 	if err != nil {
-		return h, err
+		return w.removeIfDeleted(path, h, err)
 	}
 
 	if s.File(path).Worktree == Unmodified {
@@ -248,6 +248,23 @@ func (w *Worktree) Add(path string) (plumbing.Hash, error) {
 	}
 
 	return h, err
+}
+
+func (w *Worktree) removeIfDeleted(path string, hash plumbing.Hash, err error) (plumbing.Hash, error) {
+	pathError, ok := err.(*os.PathError)
+	if ok && pathError.Op == "lstat" || w.isErrNotExist(err) {
+		h, errDelete := w.deleteFromIndex(path)
+		if errDelete != nil {
+			return h, errDelete
+		}
+		return hash, nil
+	}
+
+	return hash, err
+}
+
+func (w *Worktree) isErrNotExist(err error) bool {
+	return err.Error() == os.ErrNotExist.Error()
 }
 
 func (w *Worktree) copyFileToStorage(path string) (hash plumbing.Hash, err error) {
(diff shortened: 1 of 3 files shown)


grunenwflorian:
I cannot reproduce that error. Any idea why ?

mcuadros:
Its a test randomly failing, don't worry

worktree_status.go
```diff
@@ -266,6 +266,23 @@ func (w *Worktree) Add(path string) (plumbing.Hash, error) {
 	return h, err
 }
 
+func (w *Worktree) removeIfDeleted(path string, hash plumbing.Hash, err error) (plumbing.Hash, error) {
+	pathError, ok := err.(*os.PathError)
+	if ok && pathError.Op == "lstat" || w.isErrNotExist(err) {
```
mcuadros: Can you use the built-in function`os.IsNotExist`?

worktree_status.go
```diff
@@ -266,6 +266,19 @@ func (w *Worktree) Add(path string) (plumbing.Hash, error) {
 	return h, err
 }
 
+func (w *Worktree) removeIfDeleted(path string, hash plumbing.Hash, err error) (plumbing.Hash, error) {
+	pathError, ok := err.(*os.PathError)
```
mcuadros: you don't need to do the cast or check if is a lsstat, the function is smart enough for detect any non-existan file

mcuadros: Also I rather having the if condition where the `err` is generated instead of having another function. Thanks

grunenwflorian: Done.

PR #582, opened 2017-09-06
packfile: improve performance of delta generation

erizocosmico:
*Caveats*

I kept the old API for deltas untouched and added some private methods that reuse indexes, so the methods can be used outside and the index is automatically created.

I ported a modified version of JGit's DeltaIndex and DeltaIndexScanner that is faster (not much more, though) and, somehow, make the encode process faster. A process that consistently took ~2m30s before now takes ~1m with this. This might be caused by a bug in the index, of course, but everything seems to be ok according to the tests and some manual check I did with the aforementioned process.


erizocosmico:
This is ready for review now

PR #587, opened 2017-09-09
config: support a configurable, and turn-off-able, pack.window

strib:
One use of go-git is to transfer git data from a non-standard git repo (not stored in a file system, for example) to a "remote" backed by a standard, local .git repo.

In this scenario, delta compression is not needed to reduce transfer time over the "network", because there is no network.  The underlying storage layer has already taken care of the data tranfer, and sending the objects to local .git storage doesn't require compression.  So this PR gives the user the option to turn off compression when it isn't needed.

Of course, this results in a larger, uncompressed local .git repo, but the user can then run `git gc` or `git repack` on that repo if they care about the storage costs.

Turning the pack window to 0 reduces total push time of a 36K repo by 50 seconds (out of a pre-PR total of 3m26s).

### DIFF
--- a/plumbing/format/packfile/delta_selector.go
+++ b/plumbing/format/packfile/delta_selector.go
@@ -9,9 +9,6 @@ import (
 )
 
 const (
-	// How far back in the sorted list to search for deltas.  10 is
-	// the default in command line git.
-	deltaWindowSize = 10
 	// deltas based on deltas, how many steps we can do.
 	// 50 is the default value used in JGit
 	maxDepth = int64(50)
@@ -31,12 +28,22 @@ func newDeltaSelector(s storer.EncodedObjectStorer) *deltaSelector {
 	return &deltaSelector{s}
 }
 
-// ObjectsToPack creates a list of ObjectToPack from the hashes provided,
-// creating deltas if it's suitable, using an specific internal logic
-func (dw *deltaSelector) ObjectsToPack(hashes []plumbing.Hash) ([]*ObjectToPack, error) {
-	otp, err := dw.objectsToPack(hashes)
+// ObjectsToPack creates a list of ObjectToPack from the hashes
+// provided, creating deltas if it's suitable, using an specific
+// internal logic.  `packWindow` specifies the size of the sliding
+// window used to compare objects for delta compression; 0 turns off
+// delta compression entirely.
+func (dw *deltaSelector) ObjectsToPack(
+	hashes []plumbing.Hash,
+	packWindow uint,
+) ([]*ObjectToPack, error) {
+	otp, err := dw.objectsToPack(hashes, packWindow)
 	if err != nil {
 		return nil, err
+	}
+
+	if packWindow == 0 {
+		return otp, nil
 	}
 
 	dw.sort(otp)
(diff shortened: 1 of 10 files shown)


orirawlings:
Is this a feature in the standard git tooling?

strib:
> Is this a feature in the standard git tooling?

Yes: https://stackoverflow.com/questions/7102053/git-pull-without-remotely-compressing-objects

mcuadros:
Thanks for your contribution, make a lot of sense for people using memory based storage or others.

I cannot see a config flag exactly for this feature, the [core.compression](https://git-scm.com/docs/git-config#git-config-corecompression) flag in stackoverflow, has a different behavior. 

As you can see in the document this kind of configuration it's done at config level. So instead of having it as a flag of the `PullOptions`, should be implemented as property in the config file. 

If this property doesn't exists on the standard git, I am open to merge this PR with a flag call `gogit.deltaCompression` or something similar.

orirawlings:
It looks like this is controlled via a gitattributes option on specific file glob patterns. Ideally we would control this behavior via the same gitattributes, but I don't believe that currently the go-git library supports gitattributes. The are a bunch of other behaviors that can be controlled via gitattributes as well.

mcuadros:
This is handle by the `config` package current we only support a few config settings, but we already have the `Core` property.

https://github.com/src-d/go-git/blob/bb3217ce5d5ed682a5c830c40ea031d3c92a8a7e/config/config.go#L35-L42

strib:
Thanks for taking a look guys.  I don't think this is really equivalent to a core config param.   It's more equivalent to setting `--window=0` when calling `git-pack-objects`: https://git-scm.com/docs/git-pack-objects#git-pack-objects---windowltngt .  That can also be controlled in the config file with `pack.window`: https://git-scm.com/docs/git-config#git-config-packwindow

I'm not sure why `git` doesn't make it a per-remote option, that seems more reasonable to me.  But if you guys want to keep it closer to the real git, I can turn this into support for `pack.window`, where `window=0` means delta compression is turned off.  What do you think?

mcuadros:
Yes, I think this is the best solution, since implementing this as `pack.window` is giving support a feature from git. Also will be nice change the the packfile encoder to accept the argument of the window.

strib:
Thanks @mcuadros, done.  Please take another look.

plumbing/format/packfile/delta_selector.go
```diff
@@ -33,12 +30,19 @@ func newDeltaSelector(s storer.EncodedObjectStorer) *deltaSelector {
 
 // ObjectsToPack creates a list of ObjectToPack from the hashes provided,
 // creating deltas if it's suitable, using an specific internal logic
-func (dw *deltaSelector) ObjectsToPack(hashes []plumbing.Hash) ([]*ObjectToPack, error) {
-	otp, err := dw.objectsToPack(hashes)
+func (dw *deltaSelector) ObjectsToPack(
+	hashes []plumbing.Hash,
+	packWindow uint,
```
mcuadros: Please update the comment about this new argument.

plumbing/format/packfile/encoder.go
```diff
@@ -47,8 +47,11 @@ func NewEncoder(w io.Writer, s storer.EncodedObjectStorer, useRefDeltas bool) *E
 
 // Encode creates a packfile containing all the objects referenced in hashes
 // and writes it to the writer in the Encoder.
-func (e *Encoder) Encode(hashes []plumbing.Hash) (plumbing.Hash, error) {
-	objects, err := e.selector.ObjectsToPack(hashes)
+func (e *Encoder) Encode(
+	hashes []plumbing.Hash,
+	packWindow uint,
```
mcuadros: same here.

erizocosmico:
lgtm once the requests by @mcuadros are added

bzz:
Looks great.

Probably a nitpick, but could you also please double-check if 0d74736 is intended to be part of these changes?

strib:
Added comments and rebased it correctly.  Thanks all!

PR #588, opened 2017-09-12
revlist: do not revisit ancestors as long as all branches are visited

erizocosmico:
This change is the fixed version of the previous performance improvement that was reverted due to some bogus logic.
Now it's fixed and only stops the iteration if and only if all of the branches we've come across have been visited, being a branch a parent commit of a commit we've visited.


mcuadros:
@erizocosmico rebase

erizocosmico:
Done @mcuadros

PR #608, opened 2017-09-28
Add port to SCP Endpoints

balkian:
The port for SCP-like URLs was hardcoded to 22.
This commit modifies the regex to find a port (optional), and adds a new test case that covers this scenario.

All tests pass, and I've tried cloning a repo with a custom port through the command line example and it works.

Some notes:

* 22 is kept as the default port. The ssh transport also has a default port  defined . It may be best to use 0 and let the ssh transport handle it.
* I use `strconv.Atoi` in the getter method to mimic the behavior of the other types. It might be more efficient to just parse it once after the regex.

### DIFF
--- a/plumbing/transport/common.go
+++ b/plumbing/transport/common.go
@@ -187,6 +187,7 @@ func (e urlEndpoint) Path() string {
 type scpEndpoint struct {
 	user string
 	host string
+	port string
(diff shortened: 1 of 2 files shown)


mcuadros:
Awesome tanks!

PR #609, opened 2017-10-02
remote: add support for ls-remote

### LINKED ISSUE #601: get all remote branches
support feature to get all remote branches.

darkowlzz:
Adds support for ls-remote in `Remote`.

Found this in [#601-comment](https://github.com/src-d/go-git/issues/601#issuecomment-330873442), so maybe fixes #601 .

Need some guidance in the test with relevant existing examples. Not sure if this test is enough.

### DIFF
--- a/remote.go
+++ b/remote.go
@@ -728,6 +728,39 @@ func (r *Remote) buildFetchedTags(refs memory.ReferenceStorage) (updated bool, e
 	return
 }
 
+// LSRemote performs ls-remote on the remote.
+func (r *Remote) LSRemote(auth transport.AuthMethod) ([]*plumbing.Reference, error) {
+	s, err := newUploadPackSession(r.c.URLs[0], auth)
+	if err != nil {
+		return nil, err
+	}
+
+	defer ioutil.CheckClose(s, &err)
+
+	ar, err := s.AdvertisedReferences()
+	if err != nil {
+		return nil, err
+	}
+
+	allRefs, err := ar.AllReferences()
+	if err != nil {
+		return nil, err
+	}
+
+	refs, err := allRefs.IterReferences()
+	if err != nil {
+		return nil, err
+	}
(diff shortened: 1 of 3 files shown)


remote.go
```diff
@@ -728,6 +728,39 @@ func (r *Remote) buildFetchedTags(refs memory.ReferenceStorage) (updated bool, e
 	return
 }
 
+// LSRemote performs ls-remote on the remote.
+func (r *Remote) LSRemote(auth transport.AuthMethod) ([]*plumbing.Reference, error) {
```
orirawlings: Maybe we should call this method `List`? 
```
// List the references on the remote repository
```
Thoughts @mcuadros?

orirawlings: Also, does it make sense to generalize the parameters so that it accepts an Options-style struct instead of just `transport.AuthMethod` to help ease future changes without breaking API compatibility?

mcuadros: Yep, a Options will be the best

darkowlzz: Yep, makes sense. I created an Option at first, but then dropped it thinking it would be simpler to just pass the `AuthMethod` directly 😅. Adding it back...

remote_test.go
```diff
@@ -642,3 +642,40 @@ func (s *RemoteSuite) TestGetHaves(c *C) {
...
+	sto, err := filesystem.NewStorage(srcFs)
+	c.Assert(err, IsNil)
+
+	remote := newRemote(sto, &config.RemoteConfig{
+		Name: DefaultRemoteName,
+		URLs: []string{url},
+	})
+
+	rs := config.RefSpec("refs/heads/*:refs/heads/*")
+	err = remote.Push(&PushOptions{
```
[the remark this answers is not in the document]
darkowlzz: Thanks. Makes sense. 🙂

PR #610, opened 2017-10-03
remote: add the last 100 commits for each ref in haves list

strib:
I found that a `Remote.Fetch` call, where the repo's branch is ahead of the remote, causes the remote to send the entire history of the branch being fetched, because no common ancestor is negotiated.  This represents a quick and dirty fix until multi-ack is implemented.  Understandable if you don't want to merge half-measures, thought I'd submit it just in case!

===========
If the local ref is not an ancestor of the remote ref being fetched, then when we send an UploadPack request with that local ref as one of the Haves, the remote will not recognize it, and will think we are asking for the entire history of the repo, even if there's a common ancestor.

To do this right, we need to support the multi-ack protocol so we can negotiate a common commit.  That's hard though; this is a quick fix just to include the previous 100 commits for each local ref in the Haves list, and hope that one of them is the common commit.

### DIFF
--- a/remote.go
+++ b/remote.go
@@ -27,6 +27,14 @@ var (
 	ErrDeleteRefNotSupported = errors.New("server does not support delete-refs")
 )
 
+const (
+	// This describes the maximum number of commits to walk when
+	// computing the haves to send to a server, for each ref in the
+	// repo containing this remote, when not using the multi-ack
+	// protocol.  Setting this to 0 means there is no limit.
+	maxHavesToVisitPerRef = 100
+)
+
 // Remote represents a connection to a remote repository.
 type Remote struct {
 	c *config.RemoteConfig
@@ -284,7 +292,7 @@ func (r *Remote) fetch(ctx context.Context, o *FetchOptions) (storer.ReferenceSt
 
 	req.Wants, err = getWants(r.s, refs)
 	if len(req.Wants) > 0 {
-		req.Haves, err = getHaves(localRefs)
+		req.Haves, err = getHaves(localRefs, remoteRefs, r.s)
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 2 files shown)


strib:
By the way, this is what all my recent PRs having been working towards: https://keybase.io/blog/encrypted-git-for-everyone (shout-out to this project in that post).  Thanks everyone here for the library and your hard work!

remote.go
```diff
@@ -490,8 +490,29 @@ func (r *Remote) references() ([]*plumbing.Reference, error) {
...
+	localRefs []*plumbing.Reference,
+	remoteRefs storer.ReferenceStorer,
+	s storage.Storer,
+) ([]plumbing.Hash, error) {
 	haves := map[plumbing.Hash]bool{}
+
+	// Build a map of all the remote references, to avoid loading too
+	// many parent commits for references we know don't need to be
+	// transferred.
+	stopEarlyRefs := map[plumbing.Hash]bool{}
```
orirawlings: Maybe we can rename this to `remoteRefs` and rename the function parameter to `remoteRefStorer`?

remote.go
```diff
@@ -490,8 +490,29 @@ func (r *Remote) references() ([]*plumbing.Reference, error) {
...
+	iter, err := remoteRefs.IterReferences()
+	if err != nil {
+		return nil, err
+	}
+	err = iter.ForEach(func(ref *plumbing.Reference) error {
+		stopEarlyRefs[ref.Hash()] = true
```
orirawlings: Do we need to do something special here to confirm the `ref` isn't symbolic?

remote.go
```diff
@@ -501,7 +522,36 @@ func getHaves(localRefs []*plumbing.Reference) ([]plumbing.Hash, error) {
...
+		if err != nil {
+			haves[ref.Hash()] = true
+			continue
+		}
+
+		// Until go-git supports proper commit negotiation during an
+		// upload pack request, include up to 100 commits from the
+		// history of each ref.
+		walker := object.NewCommitPreorderIter(commit, haves, nil)
+		const maxToVisit = 100
```
erizocosmico: since it's a constant, what do you think about moving this to the top level?

strib:
Thanks everyone for taking a look.  I made all the suggested changes, and factored out most of the logic into two new helper functions that can be dropped if desired when multi-ack is ready.  I made a top-level constant that can be set to 0, if we decide we want it to send ALL haves that the remote might not know about.  Please take another look, thanks!

PR #613, opened 2017-10-06
Add Stats() to Commit

### LINKED ISSUE #482: retrieving the stats on a commit
Hey guys, thanks for a great library. I was curious if there is a way of get the stats from a commit. Something equivalent to ```git show --stat <hash>``` If not could I be pointed in roughly the right direction to add it?

darkowlzz:
Stats() is similar to `git show --stat <hash>`.

This introduces a new type `FileStat` which stores the stats of a file patch. It can be printed to have
```
common.go | 4 ++++
main.go | 26 ++++++++++++++++++++------
```
git cli style printing of addition and deletion.

Need help with:
1. Figuring out when a file is renamed. I couldn't find any existing code that does it. I came across `merkletrie.Action` which supports `"Insert"`, `"Delete"` and `"Modify"` only. And it made me assume file rename is not supported.
2. How to get patch for the first commit without any parent? To get a patch, I see 2 commits are used. Any pointers on what to do for patch of the first commit?

fixes #482

### DIFF
--- a/plumbing/object/commit.go
+++ b/plumbing/object/commit.go
@@ -265,6 +265,72 @@ func (b *Commit) Encode(o plumbing.EncodedObject) error {
 	return err
 }
 
+// FileStat stores the status of changes in content of a file.
+type FileStat struct {
+	Name     string
+	Addition int
+	Deletion int
+}
+
+func (fs FileStat) String() string {
+	totalChanges := fs.Addition + fs.Deletion
+	adds := strings.Repeat("+", fs.Addition)
+	dels := strings.Repeat("-", fs.Deletion)
+	return fmt.Sprintf(" %s | %d %s%s", fs.Name, totalChanges, adds, dels)
+}
+
+// FileStats is a collection of FileStat.
+type FileStats []FileStat
+
+// Stats shows the status
+func (c *Commit) Stats() (FileStats, error) {
+	var fileStats FileStats
+
+	// Get the previous commit.
+	ci := c.Parents()
+	parentCommit, err := ci.Next()
+	if err != nil {
+		return nil, err
+	}
+
+	patch, err := parentCommit.Patch(c)
+	if err != nil {
+		return nil, err
+	}
+
+	filePatches := patch.FilePatches()
+	for _, fp := range filePatches {
+		cs := FileStat{}
+		from, to := fp.Files()
+		if from == nil {
+			// New File is created.
+			cs.Name = to.Path()
+		} else if to == nil {
+			// File is deleted.
+			cs.Name = from.Path()
+		} else if from.Path() != to.Path() {
+			// File is renamed.
+			cs.Name = fmt.Sprintf("%s => %s", from.Path(), to.Path())
+		} else {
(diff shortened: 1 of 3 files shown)


mcuadros:
@ajnavarro you was working with diff, maybe you can give a hand here.

mcuadros:
@darkowlzz about comparing the parent tree against void, just use a empty `Noder`

plumbing/object/commit.go
```diff
@@ -222,6 +222,72 @@ func (b *Commit) Encode(o plumbing.EncodedObject) error {
...
+	Name     string
+	Addition int
+	Deletion int
+}
+
+func (fs FileStat) String() string {
+	totalChanges := fs.Addition + fs.Deletion
+	adds := strings.Repeat("+", fs.Addition)
+	dels := strings.Repeat("-", fs.Deletion)
+	return fmt.Sprintf(" %s | %d %s%s", fs.Name, totalChanges, adds, dels)
```
ajnavarro: This output is not correct. `totalChanges` is correct, but the number of `+` and `-` is not. Should be the ratio of added and removed lines. See `man diffstat` for more info.

Also: https://stackoverflow.com/questions/7024848/git-diff-stat-explanation

ajnavarro:
> Figuring out when a file is renamed. I couldn't find any existing code that does it. I came across merkletrie.Action which supports "Insert", "Delete" and "Modify" only. And it made me assume file rename is not supported.

Yes, you're right. Rename is not actually supported.

mcuadros:
We can move it to `Patch` but keep this method, as a shortcut.

plumbing/object/commit.go
```diff
@@ -222,6 +222,32 @@ func (b *Commit) Encode(o plumbing.EncodedObject) error {
...
+func (c *Commit) Stats() (FileStats, error) {
+	// Get the previous commit.
+	ci := c.Parents()
+	parentCommit, err := ci.Next()
+	if err != nil {
+		if err == io.EOF {
+			emptyNoder := treeNoder{}
+			parentCommit = &Commit{
+				Hash: emptyNoder.hash,
+				// TreeHash: emptyNoder.parent.Hash,
```
darkowlzz: @mcuadros I think I'm not using it the way you meant.
This is for the first commit scenario, no parent commit.
Without assigning a `TreeHash` I get `panic: object not found`.
And if I uncomment the above line, it panics with `invalid memory address` error.

Missing something really simple? 😅

darkowlzz:
Thanks for the review.

I have moved `FileStat(s)` to `patch.go` and added `Stats()` to `Patch` as well.

Also, created `printStat()` function to print `FileStats` considering line length to be 72-80. I read that diffstat result varies based on the screen size. Should we too do that?
And, have I made the +/- output too complex? 😅

mcuadros:
Can you rebase @darkowlzz, thanks!

darkowlzz:
Rebased!

mcuadros:
@ajnavarro PTAL

Issue #605, opened 2017-09-24
I would like to request `git-verify-commit` support please


blacktop:
great library btw! 😁

blacktop:
Or at the very least the ability to extract the signature, sign from a commit?

blacktop:
Is this done now? @darkowlzz If so many much thanks! 👍

blacktop:
Is this actually done?  It seems like support for signatures to be included in commits has been added, but not the ability to verify them?

mcuadros:
True, is not implemented, thanks!

PR #616, opened 2017-10-08
Add support for signed commits

### LINKED ISSUE #524: PGP signed commits
Right now, PGP signed commits look like:

```
Author: emersion <contact@emersion.fr>
Date:   Sun Jul 30 12:32:19 2017 +0200

     iQJIBAABCAAyFiEENP+VJs/vDpejQOLkD9574OiPXkgFAll9tbMUHGNvbnRhY3RA
     ZW1lcnNpb24uZnIACgkQD9574OiPXkjPlg/9H60O+GAHRZIY3/B31Y6YynhZl2jz
     lTGLfaLf1quOCkonMFwBxpEZp1Ns9IhFB0lzwFx9LV53pVI9d18tVsOkGRMYRs7G
     6FPk7IEtLeweUMz1iz0Vy2Qxy99jDf6+bQo9nW8J/7xVNb47DCdKTN9TAg2kMevL
     844MFBLFVtaOlyVcKdebjadVzDT/YqQuR9HdqbpzyrlolQsKkHsAw0j9Zvw0tk/v
     sLULwGEYRcMz5AH/y2E/R04O+PRUEH6i2nRvcHjmuUf7IrcIxW4hHDETQiItoRhn
     x8/+RnqPH7IO4G/FsNpV+lF4pBs2/6DI2AXgX9INjrc4vrA4g8mlQPSSjuGWqBvH
     DqMY4GtIkRCggHpvr8uOPbIjeRJv4LQbvxR+qB4W0HX1x67FMgRTz86XnLhbb3/m
     lnI2n1Yq/fiOmmkgrEP/5deSrNY7XmSMkDf7zPcfmMd1T6uk64ScrUIR+Ff0gf0d
     WvbcZasJmF+ZOydS0nhZkE4hIoAFdYKQIkBRGel9kU2wuV5LpiXbUfV8a/49LLay
     XTnBtEx6qMkmI3An6ut/HkrcH0vSu2YMrM9A126HM7/iCy0wdIvI5Ey8Hj0cx/hq
     rVnqKJWI6HK2M1SjKSxeE1cozMuAjLzMZxwbYKvbc+9RCsvpg/Xfubrnx5K71Wpb
     7rEPAgVQ5bBUQug=
     =dXmn
     -----END PGP SIGNATURE-----

    Fix revision in links in /tree and /blob
```

It would be nice to add a new `PGPSignature []byte` field in `object.Commit` to hold the signature, and remove it from the `Message` field. What do you think?

darkowlzz:
This change adds `PGPSignature` field to `Commit` object. This is used
to store the signature of the commit, if any.

Related to #605 
fixes #524

### DIFF
--- a/plumbing/object/commit.go
+++ b/plumbing/object/commit.go
@@ -12,6 +12,11 @@ import (
 	"gopkg.in/src-d/go-git.v4/utils/ioutil"
 )
 
+const (
+	beginpgp string = "-----BEGIN PGP SIGNATURE-----"
+	endpgp   string = "-----END PGP SIGNATURE-----"
+)
+
 // Hash represents the hash of an object
 type Hash plumbing.Hash
 
@@ -28,6 +33,8 @@ type Commit struct {
 	// Committer is the one performing the commit, might be different from
 	// Author.
 	Committer Signature
+	// GPGSignature is the GPG signature of the commit.
+	GPGSignature string
 	// Message is the commit message, contains arbitrary text.
 	Message string
 	// TreeHash is the hash of the root tree of the commit.
@@ -145,10 +152,31 @@ func (c *Commit) Decode(o plumbing.EncodedObject) (err error) {
 	r := bufio.NewReader(reader)
 
 	var message bool
+	var gpgsig bool
 	for {
 		line, err := r.ReadBytes('\n')
 		if err != nil && err != io.EOF {
 			return err
+		}
+
+		if gpgsig {
+			// Check if it's the end of a PGP signature.
+			if bytes.Contains(line, []byte(endpgp)) {
+				c.GPGSignature += string(endpgp) + "\n"
+				gpgsig = false
+			} else {
+				// Trim the left padding.
+				line = bytes.TrimLeft(line, " ")
+				c.GPGSignature += string(line)
+			}
+			continue
+		}
+
+		// Check if it's the beginning of a PGP signature.
+		if bytes.Contains(line, []byte(beginpgp)) {
+			c.GPGSignature += string(beginpgp) + "\n"
+			gpgsig = true
+			continue
 		}
 
 		if !message {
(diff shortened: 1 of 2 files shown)


plumbing/object/commit.go
```diff
@@ -28,6 +33,8 @@ type Commit struct {
 	// Committer is the one performing the commit, might be different from
 	// Author.
 	Committer Signature
+	// GPGSignature is the GPG signature of the commit.
+	GPGSignature string
```
orirawlings: Maybe this should be renamed to `PGPSignature`?

plumbing/object/commit.go
```diff
@@ -145,12 +152,33 @@ func (c *Commit) Decode(o plumbing.EncodedObject) (err error) {
 	r := bufio.NewReader(reader)
 
 	var message bool
+	var gpgsig bool
```
orirawlings: Maybe rename to `pgpsig`?

plumbing/object/commit.go
```diff
@@ -145,12 +152,33 @@ func (c *Commit) Decode(o plumbing.EncodedObject) (err error) {
...
 	for {
 		line, err := r.ReadBytes('\n')
 		if err != nil && err != io.EOF {
 			return err
 		}
 
+		if gpgsig {
+			// Check if it's the end of a PGP signature.
+			if bytes.Contains(line, []byte(endpgp)) {
+				c.GPGSignature += string(endpgp) + "\n"
```
orirawlings: I don't think you need the cast to `string` here

plumbing/object/commit.go
```diff
@@ -145,12 +152,33 @@ func (c *Commit) Decode(o plumbing.EncodedObject) (err error) {
...
+				gpgsig = false
+			} else {
+				// Trim the left padding.
+				line = bytes.TrimLeft(line, " ")
```
orirawlings: Is there a reason we need to Trim the padding? I notice that we add padding back in when we format the Commit. Will this be necessary for handing the signature into another routine that does the verification of the signature?

plumbing/object/commit.go
```diff
@@ -215,6 +243,21 @@ func (b *Commit) Encode(o plumbing.EncodedObject) error {
 		return err
 	}
 
+	if b.GPGSignature != "" {
+		if _, err = fmt.Fprint(w, "gpgsig"); err != nil {
```
orirawlings: Maybe we should try to mimic the original format of the commit content here (i.e. don't print this gpgsig field that didn't exist in the original commit, just print the signature at the end of the commit comment).

If we do leave this in we should probably rename to `pgpsig` anyways.

plumbing/object/commit.go
darkowlzz: The padding is just git's way of presenting it. The public key block at https://www.gnupg.org/signature_key.html shows it without any left padding. So I thought that's how they are actually stored as a single signature. And yes, I thought it would be easier to perform signature verification in the future, without any need to trim the padding.

plumbing/object/commit.go
darkowlzz: Looking at `git cat-file` of a commit with signature, I don't think printing the signature at the end of commit comment would be good. The signature is not needed all the time but only when we want to verify. So, I would say we leave it with its own field. And yes, `pgpsig`, which doesn't matters much because we are using the header and footer of signature to detect them 😊

PR #639, opened 2017-11-08
transport/client: use mutex to synchronize usage of protocol map

erizocosmico:
This allows concurrent usage of `NewClient` and `InstallProtocol`.

### DIFF
--- a/plumbing/transport/client/client.go
+++ b/plumbing/transport/client/client.go
@@ -4,6 +4,7 @@ package client
 
 import (
 	"fmt"
+	"sync"
 
 	"gopkg.in/src-d/go-git.v4/plumbing/transport"
 	"gopkg.in/src-d/go-git.v4/plumbing/transport/file"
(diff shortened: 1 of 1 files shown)


mcuadros:
go-git is not thread safe, so if you need to install a driver please on your side you should do it safely.

erizocosmico:
While I agree that go-git should not strive for thread safety, not having this be concurrent makes it impossible to perform concurrent pushes with go-git if you need to use custom protocols for such task.

Imagine the following scenario:

```
mut.Lock()
client.InstallProtocol(myFancyProto, proto)
mut.Unlock()

mut.RLock()
repo.Push()
mut.RUnlock()
```

`RLock` allows several accesses, as long as they're only read accesses, but before that it needs a write access to install the protocol. So, effectively, until a repo finishes the push another can't install its protocol and perform the push and so on.

So any operation that depends on push can't be done in parallel if you need the installation of protocols for such push, because the `NewClient` is performed deep down the `Push` method and you can't just synchronize that from your code, you're force to synchronize `Push`.

orirawlings:
I think this change makes sense and should be included. It is certainly not obvious from a users perspective that transport operations on separate Repository objects might require synchronization. I think that typically synchronization is achieved by ensuring separate go routines have separate Repository objects, but without this change there are cases which require further synchronization steps. I think it is easier to add and maintain this patch than trying to educate users on how to avoid this particular concurrency issue.

mvpmvh:
why would you choose _not_ to merge this pull request? @erizocosmico use-case above is exactly the problem i just ran into.

PR #640, opened 2017-11-10
utils: merkletrie, filesystem fix symlinks to dir 

dimonomid:
I tried to figure how to write a proper test, but failed in a reasonable time frame, so sending the PR without a test. Nevertheless, here is a tiny repository which reproduces an issue: https://github.com/dimonomid/git-go-symlink-test/blob/master/test_symlink_dir.go

It uses another repository as a test case: https://github.com/dimonomid/git-go-fixture-dir-symlink , which has just two commits; the second one adds this tree:
```
.
├── dir2
│   └── file2
└── dir3
    └── dir2_symlink -> ../dir2
```
And to reproduce the issue, we can just clone the repo (so that we have these `dir2/file2` and symlink to it), and then checkout the previous revision. This is the list of changes it tries to apply:

```
<Delete dir2/file2>
<Delete dir3/dir2_symlink>
<Delete dir3/dir2_symlink/file2>
```

And obviously the last one fails, because `dir2/file2` was already removed before. This PR prevents non-dir items from having children.

### DIFF
--- a/utils/merkletrie/filesystem/node.go
+++ b/utils/merkletrie/filesystem/node.go
@@ -77,6 +77,10 @@ func (n *node) NumChildren() (int, error) {
 }
 
 func (n *node) calculateChildren() error {
+	if !n.IsDir() {
+		return nil
+	}
+
 	if len(n.children) != 0 {
 		return nil
 	}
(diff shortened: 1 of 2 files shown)


mcuadros:
Make sense, but we need a test for merge this.

PR #650, opened 2017-11-20
transport: converts Endpoint interface into a struct

mcuadros:
This PR converts `transport.Endpoint` interface into a struct in order to make easier build third-party transports .

For better understanding of this pr refer to https://github.com/src-d/go-git/pull/650/commits/d9385afb02e18b19102764e460cb445b43d6dc91

Fixes #437, #424


campoy:
Hey, what's the reasoning behind this? The link you show just points to some code change but I don't see why that change is beneficial.

Thanks

PR #652, opened 2017-11-21
plumbing/object: do not eat error on tree decode

ferhatelmas:


ferhatelmas:
@orirawlings Pushed one test in a separate commit, let me know if it makes sense.

PR #656, opened 2017-11-24
plumbing: object, fix Commit.Verify test

darkowlzz:
Commit verification test seems to be passing on my computer still, even on master. This PR is for debugging the failure in Travis.


darkowlzz:
I see the difference in the commit timezone.

```
author Sunny <me@darkowlzz.space> 1511197315 +0000
committer Sunny <me@darkowlzz.space> 1511197315 +0000
```

Since I used `time.Unix()`, it generated the time in relation to my local time, which is different from what the signed commit had.

darkowlzz:
Test was failing because of the relative time zone attached to the commit.
Adding a fixed location to the time fixes the issue.

PR #657, opened 2017-11-24
plumbing: transport/http, Close http.Body reader when needed

ajnavarro:
If we don't do that, all the goroutines opened by the Body Reader are never closed, causing a 'fatal: morestack on g0' error.


ajnavarro:
Related issue: https://github.com/src-d/borges/issues/186

PR #658, opened 2017-11-24
plumbing: object/tag, add signature and verification support

darkowlzz:


plumbing/object/tag.go
```diff
@@ -225,6 +269,31 @@ func (t *Tag) String() string {
...
+	keyring, err := openpgp.ReadArmoredKeyRing(keyRingReader)
+	if err != nil {
+		return nil, err
+	}
+
+	// Extract signature.
+	signature := strings.NewReader(t.PGPSignature)
+
+	// Remove signature. Keep only the tag components.
+	t.PGPSignature = ""
```
orirawlings: I think here and in `Commit.Verify` we should avoid mutating the object in this method. Maybe we should extract a new method `Tag.encode(o plumbing.EncodedObject, includeSig bool)` which we can call from here and `Tag.Encode` with different parameters.

darkowlzz: Sounds good. I've added method `encode()` to both `Commit` and `Tag`, which accepts an extra `includeSig` parameter to include or exclude signature.

darkowlzz:
> I think here and in Commit.Verify we should avoid mutating the object in this method. Maybe we should extract a new method Tag.encode(o plumbing.EncodedObject, includeSig bool) which we can call from here and Tag.Encode with different parameters.

@orirawlings  Sounds good. I've added method `encode()` to both `Commit` and `Tag`, which accepts an extra `includeSig` parameter to include or exclude signature.

Issue #599, opened 2017-09-19
git rev-parse?


ankurmittal:
How do i find commit hash for a particular ref, ref can be a commit hash, branch name, tag name, etc
eg
a891109d49d61fc0792912f0d741b45ec1516525
refs/heads/branch
heads/branch
branch
refs/tags/tag
tag

orirawlings:
There exists a method [(*Repository).ResolveRevision](https://godoc.org/gopkg.in/src-d/go-git.v4#Repository.ResolveRevision) which satisfies part of the behavior of `rev-parse`, including resolving the relative operators (like `~`, `^`, etc.) but it is missing support for other basic scenarios (like the ones you describe).

@mcuadros was the intent that the `internal/revision` package would implement the full [gitrevisions spec](https://www.kernel.org/pub/software/scm/git/docs/gitrevisions.html)? If so, I think we want to add support for full and short hash literals as well as reference name disambiguation.

There are a handful of other `gitrevisions` that require broader feature support which is currently missing in this library. For example, reflog and upstream revisions.

ankurmittal:
any update?

mcuadros:
@orirawlings the goals is cover all the cases. So will be great if any wants to implement it.

antham:
This case ```a891109d49d61fc0792912f0d741b45ec1516525``` isn't taking in account, I didn't implement it, but I think I won't be that hard.
Those 2, ```refs/heads/branch``` and ```refs/tags/tag``` work actually, I tried with a repository of mine and it's resolved with the right sha .
Those cases ```heads/branch```, ```branch``` and ```tag`` are bugs, I tried, they are correctly parsed as Ref but they are not resolved correctly behind, I need to check how to fix that.

antham:
@mcuadros this issue needs to be reopened, this case `a891109d49d61fc0792912f0d741b45ec1516525` is not covered yet, I'm working on it.

ankurmittal:
Also are you planning to cover short hashes?

antham:
Yep I'm going to do so as well

andreimc:
@antham any progress on shorthand hashes ?

PR #660, opened 2017-11-26
fix Repository.ResolveRevision for branch and tag

antham:
Branch and tag revision are properly parsed by revision lexer but not properly solved into references.

#599

### DIFF
--- a/example_test.go
+++ b/example_test.go
@@ -85,36 +85,6 @@ func ExampleRepository_References() {
 
 }
 
-func ExampleRepository_ResolveRevision() {
-	r, _ := git.Clone(memory.NewStorage(), nil, &git.CloneOptions{
-		URL: "https://github.com/git-fixtures/basic.git",
-	})
-
-	// Resolve revision to their corresponding sha
-	revisions := []string {
-		"master~1",
-		"HEAD~2^^~",
-		"HEAD~2^{/binary file}",
-		"refs/heads/master~2^^~",
-	}
-
-	for _, revision := range revisions {
-		h, err := r.ResolveRevision(plumbing.Revision(revision))
-
-		if err != nil {
-			log.Fatalf("%s : %s", revision, err)
-		}
-
-		fmt.Printf("%s  %s\n", h.String(), revision)
-	}
-
-	// Output:
-	// 918c48b83bd081e863dbe1b80f8998f058cd8294  master~1
-	// b029517f6300c2da0f4b651b8642506cd6aaf45d  HEAD~2^^~
-	// 35e85108805c84807bc66a02d91535e1e24b38b9  HEAD~2^{/binary file}
-	// b029517f6300c2da0f4b651b8642506cd6aaf45d  refs/heads/master~2^^~
-}
-
 func ExampleRepository_CreateRemote() {
 	r, _ := git.Init(memory.NewStorage(), nil)
 
--- a/example_test.go
+++ b/example_test.go
@@ -85,6 +85,36 @@ func ExampleRepository_References() {
 
 }
 
+func ExampleRepository_ResolveRevision() {
+	r, _ := git.Clone(memory.NewStorage(), nil, &git.CloneOptions{
+		URL: "https://github.com/git-fixtures/basic.git",
+	})
+
+	// Resolve revision to their corresponding sha
+	revisions := []string {
+		"master~1",
+		"HEAD~2^^~",
+		"HEAD~2^{/binary file}",
+		"refs/heads/master~2^^~",
+	}
+
+	for _, revision := range revisions {
+		h, err := r.ResolveRevision(plumbing.Revision(revision))
+
+		if err != nil {
+			log.Fatalf("%s : %s", revision, err)
+		}
+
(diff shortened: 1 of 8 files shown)


repository.go
```diff
@@ -977,29 +991,6 @@ func (r *Repository) ResolveRevision(rev plumbing.Revision) (*plumbing.Hash, err
...
-				return nil
-			})
-			if err != nil {
-				return &plumbing.ZeroHash, err
-			}
-
-			if c == nil {
-				return &plumbing.ZeroHash, fmt.Errorf(`No commit exists prior to date "%s"`, date.String())
-			}
-
```
antham: I removed this part because it's a bad implementation,I misread the spec, it must follows this : `This suffix may only be used immediately following a ref name and the ref must have an existing log ($GIT_DIR/logs/<ref>)`

repository.go
```diff
@@ -897,15 +897,29 @@ func (r *Repository) ResolveRevision(rev plumbing.Revision) (*plumbing.Hash, err
...
+			revisionRef := item.(revision.Ref)
+			var ref *plumbing.Reference
+
+			for _, rule := range []string{
+				"%s",
+				"refs/%s",
+				"refs/tags/%s",
+				"refs/heads/%s",
+				"refs/remotes/%s",
+				"refs/remotes/%s/HEAD",
```
antham: This part exists already in plumbing, I was wondering if it would be possible to move all this code related to revision in plumbing ?

example_test.go
```diff
@@ -85,6 +85,36 @@ func ExampleRepository_References() {
...
+		h, err := r.ResolveRevision(plumbing.Revision(revision))
+
+		if err != nil {
+			log.Fatalf("%s : %s", revision, err)
+		}
+
+		fmt.Printf("%s  %s\n", h.String(), revision)
+	}
+
+	// Output:
```
antham: I saw in others example, you have `// Example Output` instead of `// Output`, so examples are not compiled, is it a mistake or something you explicitly did ?

repository_test.go
```diff
@@ -1262,24 +1262,32 @@ func (s *RepositorySuite) TestWorktreeBare(c *C) {
...
+		"refs/remotes/origin/master": "6ecf0ef2c2dffb796033e5a02219af86ec6584e5",
+		"refs/remotes/origin/HEAD":   "6ecf0ef2c2dffb796033e5a02219af86ec6584e5",
+		"HEAD~2^^~":                  "b029517f6300c2da0f4b651b8642506cd6aaf45d",
+		"HEAD~3^2":                   "a5b8b09e2f8fcb0bb99d3ccb0958157b40890d69",
+		"HEAD~3^2^0":                 "a5b8b09e2f8fcb0bb99d3ccb0958157b40890d69",
+		"HEAD~2^{/binary file}":      "35e85108805c84807bc66a02d91535e1e24b38b9",
+		"HEAD~^{/!-some}":            "1669dce138d9b841a518c64b10914d88f5e488ea",
+		"master":                     "6ecf0ef2c2dffb796033e5a02219af86ec6584e5",
+		"branch":                     "e8d3ffab552895c19b9fcf7aa264d277cde33881",
+		"v1.0.0":                     "6ecf0ef2c2dffb796033e5a02219af86ec6584e5",
```
antham: This tag exists here bu I can't find it on https://github.com/git-fixtures/basic.git

antham:
@mcuadros can you review this PR pls ?

example_test.go
mcuadros: If the examples are in _examples folder are tested in a different way. But if is a example in other folder is an error

example_test.go
```diff
@@ -85,6 +85,36 @@ func ExampleRepository_References() {
 
 }
 
+func ExampleRepository_ResolveRevision() {
```
mcuadros: I prefer having this example on the `_examples` folder, following the format.

repository.go
mcuadros: Maybe we can make this public:
 https://github.com/src-d/go-git/blob/902347aef9a7cd618c48e3baa1121b72e2beab68/plumbing/reference.go#L18-L27

repository_test.go
mcuadros: Extrange... but trust on the fixture.

antham: maybe you could add it to the original repository ? It's convenient when you need to test on the repository and it will fit what is defined in fixtures.

mcuadros:
@antham ping

repository_test.go
antham: @mcuadros

repository.go
antham: Don't you think we could replace those litteral strings with constants defined just right above to avoid duplication ?

antham:
@mcuadros you can come back, AppVeyor crashs but it doesn't seem related to my tests

antham:
Now AppVeyor works but there is an error with one of the build with travis....

antham:
@mcuadros did you see my last comments in this PR about RefRevParseRules and the tag on basic repository ?

repository_test.go
mcuadros: Maybe, I will check it.

PR #663, opened 2017-11-27
storage: filesystem, add support for git alternates

### LINKED ISSUE #325: add support for git alternates
Simple test to retrieve an object from a lightweight clone results in "object not found" errors.

darkowlzz:
This change adds a new method `Alternates()` in `DotGit` to check and
query alternate source.
```
func (d *DotGit) Alternates() (string, error)
```
This is used by `ObjectStorage.EncodedObject` to check for alternates
and get the queried object from the alternate source.

I've manually tested this with the scenario and code in #325 but I'm not sure what would be the right way to setup an automated test for this. **Need help with testing**.

Fixes #325

### DIFF
--- a/storage/filesystem/internal/dotgit/dotgit.go
+++ b/storage/filesystem/internal/dotgit/dotgit.go
@@ -3,6 +3,7 @@ package dotgit
 
 import (
 	"bufio"
+	"bytes"
 	"errors"
 	"fmt"
 	"io"
@@ -756,9 +757,27 @@ func (d *DotGit) PackRefs() (err error) {
 	return nil
 }
 
-// Module return a billy.Filesystem poiting to the module folder
+// Module return a billy.Filesystem pointing to the module folder
 func (d *DotGit) Module(name string) (billy.Filesystem, error) {
 	return d.fs.Chroot(d.fs.Join(modulePath, name))
+}
+
+// Alternates returns the content of objects/info/alternates if available.
+// This can be used to checks if it's a shared repository.
+func (d *DotGit) Alternates() (string, error) {
+	altpath := d.fs.Join("objects", "info", "alternates")
+	f, err := d.fs.Open(altpath)
+	if os.IsNotExist(err) {
+		return "", err
+	}
+
+	buf := new(bytes.Buffer)
+	_, err = buf.ReadFrom(f)
+	if err != nil {
+		return "", err
+	}
+
+	return buf.String(), nil
 }
 
 func isHex(s string) bool {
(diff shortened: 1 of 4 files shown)


mcuadros:
You will need to create a fixture with altnernates and pull request it to: https://github.com/src-d/go-git-fixtures

storage/filesystem/object.go
```diff
@@ -133,6 +135,23 @@ func (s *ObjectStorage) EncodedObject(t plumbing.ObjectType, h plumbing.Hash) (p
...
+	// If the error is still object not found, check if it's a shared object
+	// repository.
+	if err == plumbing.ErrObjectNotFound {
+		alt, e := s.dir.Alternates()
+		if e == nil {
+			// Create a new object storage pointing to the path in the
+			// alternates file.
+			fs := osfs.New(filepath.Dir(alt))
+			dir := dotgit.New(fs)
+			o, oe := newObjectStorage(dir)
```
orirawlings: Maybe this secondary storage object should be cached rather than recreated each time we decode an object stored in the alternates?

storage/filesystem/internal/dotgit/dotgit.go
```diff
@@ -502,11 +503,29 @@ func (d *DotGit) readReferenceFile(path, name string) (ref *plumbing.Reference,
...
 
-// Module return a billy.Filesystem poiting to the module folder
+// Module return a billy.Filesystem pointing to the module folder
 func (d *DotGit) Module(name string) (billy.Filesystem, error) {
 	return d.fs.Chroot(d.fs.Join(modulePath, name))
 }
 
+// Alternates returns the content of objects/info/alternates if available.
+// This can be used to checks if it's a shared repository.
+func (d *DotGit) Alternates() (string, error) {
```
orirawlings: Maybe it would be better if this returned `(*DotGit, error)` instead of just the file contents?

darkowlzz:
> Maybe this secondary storage object should be cached rather than recreated each time we retrieve an object stored in the alternates?

@orirawlings some hints on how to cache storage object would be great :)

darkowlzz:
NOTE: The test `TestAlternatesRepo` depends on the fixture in https://github.com/src-d/go-git-fixtures/pull/8 .

mcuadros:
@darkowlzz Windows :(
https://ci.appveyor.com/project/mcuadros/go-git/build/693#L356

darkowlzz:
Alright, so there were two issues, both related to the paths on Windows. Glad that it was caught. 

Failure in dotgit_test.go was due to the hard-coded separators. Replaced them with `path/filepath`'s `Separator` and added Windows specific conditional in-line fixtures. Seems to work fine now.

Issue in worktree_test.go was due to the alternates in the worktree fixture being in *nix style slash path, which lead to unknown path on windows. Added an internal conversion of relative paths to slash path to solve this issue.

mcuadros:
Awesome!

PR #665, opened 2017-11-27
remote: support for non-force, fast-forward-only fetches

strib:
This PR adds support for a missing "+" in a fetch refspec, which should indicate a fast-forward-only fetch.  (Alternatively, the caller could put `Force=true` into the `FetchOptions` to indicate that all fetches should be treated as forces.)

It relies on file-level locking provided by the storage layer, so that distributed storage layers (being accessed concurrently by multiple independent nodes each running go-git) can be supported.

Code mostly written by @taruti, based on an early version by @lupine in #490.

### DIFF
--- a/plumbing/storer/reference.go
+++ b/plumbing/storer/reference.go
@@ -16,6 +16,10 @@ var ErrMaxResolveRecursion = errors.New("max. recursion level reached")
 // ReferenceStorer is a generic storage of references.
 type ReferenceStorer interface {
 	SetReference(*plumbing.Reference) error
+	// CheckAndSetReference sets the reference `new`, but if `old` is
+	// not `nil`, it first checks that the current stored value for
+	// `old.Name()` matches the given reference value in `old`.  If
+	// not, it returns an error and doesn't update `new`.
 	CheckAndSetReference(new, old *plumbing.Reference) error
 	Reference(plumbing.ReferenceName) (*plumbing.Reference, error)
 	IterReferences() (ReferenceIter, error)
--- a/plumbing/storer/reference.go
+++ b/plumbing/storer/reference.go
@@ -16,6 +16,7 @@ var ErrMaxResolveRecursion = errors.New("max. recursion level reached")
 // ReferenceStorer is a generic storage of references.
 type ReferenceStorer interface {
 	SetReference(*plumbing.Reference) error
+	CheckAndSetReference(new, old *plumbing.Reference) error
 	Reference(plumbing.ReferenceName) (*plumbing.Reference, error)
 	IterReferences() (ReferenceIter, error)
 	RemoveReference(plumbing.ReferenceName) error
--- a/storage/filesystem/internal/dotgit/dotgit.go
+++ b/storage/filesystem/internal/dotgit/dotgit.go
@@ -294,6 +294,10 @@ func (d *DotGit) SetRef(r, old *plumbing.Reference) error {
 
 	defer ioutil.CheckClose(f, &err)
 
+	// Lock is unlocked by the deferred Close above. This is because Unlock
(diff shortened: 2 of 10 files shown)


plumbing/storer/reference.go
```diff
@@ -16,6 +16,7 @@ var ErrMaxResolveRecursion = errors.New("max. recursion level reached")
 // ReferenceStorer is a generic storage of references.
 type ReferenceStorer interface {
 	SetReference(*plumbing.Reference) error
+	CheckAndSetReference(new, old *plumbing.Reference) error
```
mcuadros: I know that the order methods are updocumented too, I belive because looks like obviuos (even with that we should document the behaviour at some point), but we should document this one now, sinse for me it`s not a obviuos the goal.

storage/filesystem/internal/dotgit/dotgit.go
```diff
@@ -239,7 +240,39 @@ func (d *DotGit) Object(h plumbing.Hash) (billy.File, error) {
...
+
+	line := strings.TrimSpace(string(b))
+	return plumbing.NewReferenceFromStrings(name, line), nil
+}
+
+func (d *DotGit) checkReferenceAndTruncate(f billy.File, old *plumbing.Reference) error {
+	if old == nil {
+		return nil
+	}
+	ref, err := d.readReferenceFrom(f, old.Name().String())
```
mcuadros: This function is not covered by the test.

storage/memory/storage.go
```diff
@@ -202,6 +203,20 @@ func (r ReferenceStorage) SetReference(ref *plumbing.Reference) error {
 	return nil
 }
 
+func (r ReferenceStorage) CheckAndSetReference(ref, old *plumbing.Reference) error {
+	if ref != nil {
```
mcuadros: Can be change this to:
``` 
if ref == nil {
    return nil
}
````

File: storage/memory/storage.go [hunk 4]
```diff
@@ -202,6 +203,20 @@ func (r ReferenceStorage) SetReference(ref *plumbing.Reference) error {
 	return nil
 }
 
+func (r ReferenceStorage) CheckAndSetReference(ref, old *plumbing.Reference) error {
```
mcuadros: This function is not covered by the tests.

storage/filesystem/internal/dotgit/dotgit.go
strib: `readReferenceFrom` is definitely covered by the tests, since it's called by `readReferenceFile` below.  But if you're referring to `checkReferenceAndTruncate`, sure I'll add a test for it.

storage/memory/storage.go
strib: Ah, I didn't notice that the remote tests don't use filesystem storage.  I'll add a test.

strib:
@mcuadros I've added more tests and addressed your comments.  Please take another look, thanks!

plumbing/storer/reference.go
```diff
@@ -16,6 +16,11 @@ var ErrMaxResolveRecursion = errors.New("max. recursion level reached")
 // ReferenceStorer is a generic storage of references.
 type ReferenceStorer interface {
 	SetReference(*plumbing.Reference) error
+	// CheckAndSetReference sets the reference `new`, but if `old` is
+	// not `nil`, it first checks that the current stored value for
+	// `old.Name()` matches the given reference value in `old`.  If
+	// not, it returns an error and doesn't update `new`.
+	CheckAndSetReference(new, old *plumbing.Reference) error
```
orirawlings: Maybe it is a good time to add doc comments to the other methods in this interface?

PR #666, opened 2017-11-28
dotgit: handle refs that exist in both packed-refs and a loose ref file

strib:
If a repo starts out with a `packed-refs` file, and later a reference is updated by creating a new reference file for a name that already exists in `packed-refs`, then the same reference will be listed in two places (possibly with two different ref values).  This is legal in git; the loose ref file always takes precedence over the packed-ref file.  However, to support this properly, go-git needs to do two things:

1. On ref removal, remove it from both places.  First remove the ref file, and then remove the packed-refs file (under a lock).  Note that if the underlying storage layer is distributed and might be accessed concurrently, we'd likely require stronger locking semantics to make sure the (possibly out-of-date) packed-refs value isn't accessed after the loose ref is deleted but before packed-refs is cleaned up, but I leave that for future work.
2. When listing refs, don't double-report refs in `packed-refs`.  This used to be the case, but recently (#651) it was changed, so this restores the original logic (without adding the packed-refs cache back in).

### DIFF
--- a/storage/filesystem/internal/dotgit/dotgit.go
+++ b/storage/filesystem/internal/dotgit/dotgit.go
@@ -268,7 +268,7 @@ func (d *DotGit) Refs() ([]*plumbing.Reference, error) {
 		return nil, err
 	}
 
-	if err := d.addRefsFromPackedRefs(&refs); err != nil {
+	if err := d.addRefsFromPackedRefs(&refs, seen); err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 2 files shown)


strib:
(If #665 is merged first, we'll need to update the test in this PR to get it to compile, so please don't blindly merge it.)

mcuadros:
@strib merge your selft when you are ready.

PR #667, opened 2017-11-29
all: simplification

ferhatelmas:
- no length for map initialization
 - don't check for boolean/error return
 - don't format string
 - use string method of bytes buffer instead of converting bytes to string
 - use `strings.Contains` instead of `strings.Index`
 - use `bytes.Equal` instead of `bytes.Compare`

### DIFF
--- a/blame.go
+++ b/blame.go
@@ -147,10 +147,7 @@ func (b *blame) fillRevs() error {
 	var err error
 
 	b.revs, err = references(b.fRev, b.path)
-	if err != nil {
-		return err
-	}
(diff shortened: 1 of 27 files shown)


ferhatelmas:
@orirawlings I used [gosimple](https://github.com/dominikh/go-tools). I don't think we have more code for defined rules but there might be missing rules. 

@mcuadros any plan to integrate [gometalinter](https://github.com/alecthomas/gometalinter) into CI ? IMHO, lint rules must be fixed before thinking about it.

PR #669, opened 2017-11-29
storage/repository: add new functions for garbage collection

strib:
This PR lays the groundwork for garbage collection.  It adds separate functions for each of the major pieces of garbage collection:

* `ReferenceStorer.PackRefs()` packs all existing refs into a single packed-refs file, and deletes all the other loose refs.  Right now the `dotgit` implementation is equivalent to `git pack-refs --all`.
* `Repository.Prune()` deletes loose, unreferenced objects that are older than a certain specified time.
* `Repository.RepackObjects()` re-packs all referenced objects (including those in packfiles) into a single packfile, and then deletes all other packfiles older than a certain specified time.

Note that for object walking, we've included an alternative implementation from the existing `revlist` implementation, that is more memory-efficient for large repos.

I know the PR is light on tests, but I wanted to get it up quickly to see what the feedback would be.  We've written app-specific tests outside of this repo and are working on more general tests with large repos. Please let me know what types of unit tests you'd like to see added when you review.  Thanks!

### DIFF
--- a/plumbing/storer/object.go
+++ b/plumbing/storer/object.go
@@ -43,10 +43,12 @@ type EncodedObjectStorer interface {
 	// ForEachObjectHash iterates over all the (loose) object hashes
 	// in the repository without necessarily having to read those objects.
 	// Objects only inside pack files may be omitted.
+	// If ErrStop is sent the iteration is stop but no error is returned.
 	ForEachObjectHash(func(plumbing.Hash) error) error
 	// LooseObjectTime looks up the (m)time associated with the
-	// loose object (that is not in a pack file). Implementations
-	// may
+	// loose object (that is not in a pack file). Some
+	// implementations (e.g. without loose objects)
+	// always return an error.
 	LooseObjectTime(plumbing.Hash) (time.Time, error)
 	// DeleteLooseObject deletes a loose object if it exists.
 	DeleteLooseObject(plumbing.Hash) error
--- a/plumbing/storer/object.go
+++ b/plumbing/storer/object.go
@@ -3,6 +3,7 @@ package storer
 import (
 	"errors"
 	"io"
+	"time"
 
 	"gopkg.in/src-d/go-git.v4/plumbing"
 )
@@ -36,6 +37,19 @@ type EncodedObjectStorer interface {
 	//
 	// Valid plumbing.ObjectType values are CommitObject, BlobObject, TagObject,
 	IterEncodedObjects(plumbing.ObjectType) (EncodedObjectIter, error)
+	// HasEncodedObject returns ErrObjNotFound if the object doesn't
+	// exist.  If the object does exist, it returns nil.
+	HasEncodedObject(plumbing.Hash) error
+	// ForEachObjectHash iterates over all the (loose) object hashes
+	// in the repository without necessarily having to read those objects.
+	// Objects only inside pack files may be omitted.
+	ForEachObjectHash(func(plumbing.Hash) error) error
+	// LooseObjectTime looks up the (m)time associated with the
+	// loose object (that is not in a pack file). Implementations
+	// may
+	LooseObjectTime(plumbing.Hash) (time.Time, error)
+	// DeleteLooseObject deletes a loose object if it exists.
+	DeleteLooseObject(plumbing.Hash) error
 }
 
 // DeltaObjectStorer is an EncodedObjectStorer that can return delta
--- a/plumbing/storer/object.go
+++ b/plumbing/storer/object.go
@@ -52,6 +52,11 @@ type EncodedObjectStorer interface {
 	LooseObjectTime(plumbing.Hash) (time.Time, error)
 	// DeleteLooseObject deletes a loose object if it exists.
 	DeleteLooseObject(plumbing.Hash) error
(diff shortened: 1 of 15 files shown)


plumbing/storer/object.go
```diff
@@ -36,6 +37,27 @@ type EncodedObjectStorer interface {
...
+	// ForEachObjectHash iterates over all the (loose) object hashes
+	// in the repository without necessarily having to read those objects.
+	// Objects only inside pack files may be omitted.
+	// If ErrStop is sent the iteration is stop but no error is returned.
+	ForEachObjectHash(func(plumbing.Hash) error) error
+	// LooseObjectTime looks up the (m)time associated with the
+	// loose object (that is not in a pack file). Some
+	// implementations (e.g. without loose objects)
+	// always return an error.
+	LooseObjectTime(plumbing.Hash) (time.Time, error)
```
mcuadros: For avoid force the implementation of complex storers, I rather add this methods to a new interface, like, 
`LooseObjectStorer` or similar where includes all the related methods, and check if the current storer implement it, and if not, return a not supported error. 

Similar to `Transactioner` interface.

prune.go
```diff
@@ -0,0 +1,56 @@
...
+				return nil
+			}
+			// Skip too new objects.
+			if !t.Before(opt.OnlyObjectsOlderThan) {
+				return nil
+			}
+		}
+		return opt.Handler(hash)
+	})
+	if err != nil {
```
mcuadros: just return the err from the for each

storage/filesystem/object.go
```diff
@@ -479,3 +506,34 @@ func hashListAsMap(l []plumbing.Hash) map[plumbing.Hash]bool {
 
 	return m
 }
+
+func (s *ObjectStorage) ForEachObjectHash(fun func(plumbing.Hash) error) error {
+	err := s.dir.ForEachObjectHash(fun)
+	if err != nil {
```
mcuadros:
```
if err == nil {
return nil
}
```

storage/memory/storage.go
```diff
@@ -114,6 +115,14 @@ func (o *ObjectStorage) SetEncodedObject(obj plumbing.EncodedObject) (plumbing.H
 	return h, nil
 }
 
+func (o *ObjectStorage) HasEncodedObject(h plumbing.Hash) (err error) {
+	_, ok := o.Objects[h]
+	if !ok {
```
mcuadros: `if _, ok := o.Objects[h]: !ok {`

mcuadros:
In general looks awesome, just the detail of the interfaces. And of course more test.

strib:
I've addressed your comments and added some more tests @mcuadros.  Please take another look, thanks!

mcuadros:
@strib https://travis-ci.org/src-d/go-git/jobs/309254658 test are failling

strib:
Oops sorry.  I pushed a commit but that's not the only fix.  One sec...

strib:
Ok should be fixed now.  Sorry, I made some last minute changes and thought I had run the tests again, but I guess I messed it up somehow.  Ready for a look now.

mcuadros:
Windows test still failling

strib:
> Windows test still failling

Is that my fault?  It's happening on a lot of other tests too that I didn't touch.  Looks like maybe some concurrency thing with appveyor.

strib:
Oh, it's a packed-refs thing, hrm weird.  Not getting unlocked somehow.  Will look into it.

mcuadros:
@strib still one missing :(

storage/filesystem/internal/dotgit/dotgit.go
```diff
@@ -576,6 +664,105 @@ func (d *DotGit) readReferenceFile(path, name string) (ref *plumbing.Reference,
...
+// ref update could also lock packed-refs, so only one lock is
+// required during ref-packing.  But that would worsen performance in
+// the common case.
+//
+// TODO: add an "all" boolean like the `git pack-refs --all` flag.
+// When `all` is false, it would only pack refs that have already been
+// packed, plus all tags.
+func (d *DotGit) PackRefs() (err error) {
+	// Lock packed-refs, and create it if it doesn't exist yet.
+	f, err := d.fs.OpenFile(packedRefsPath, os.O_RDWR|os.O_CREATE, 0600)
```
mcuadros: Ok, I found why Windows tests are failling for this operation. Since you are open a file and keeping it open, until the end of this function, the Rename at 748 line fails. 

Why you are keep a Lock over the file and you are not writting directly on it? Instead of this are you writting in a temporal file and the replacing it. 

Maybe we can simplify this code and just lock the file and writter over it, no?

strib: Ugh Windows is the worst.  I really want to do a rename so that the operation is atomic, and if there's a crash we don't risk having a partially-written packed-refs file.

I'm done working for the night but I'll try to come up with a solution in the morning.

strib:
@mcuadros: ok, Windows passes now.  My solution was to write directly to the locked file if we're using a storage layer that doesn't support renaming over a locked file (i.e., the Windows filesystem).

Not sure what's up with the travis failures but I don't _think_ they're related.  Please take another look, thanks!

PR #670, opened 2017-11-29
worktree.go: fix checking out branches on empty repositories

eiso:
The previous behaviour of `createBranch` when checking out a branch on a git repository with no branches (directly after `git init`) and subsequently making a commit: was to ignore the branch that was checked out and commit to master. Because the case of `plumbing.ErrReferenceNotFound == true` was not handled. This PR now handles that case and creates the branch as expected. 

**Need help with testing**

### DIFF
--- a/worktree.go
+++ b/worktree.go
@@ -140,9 +140,9 @@ func (w *Worktree) Checkout(opts *CheckoutOptions) error {
 	}
 
 	if opts.Create {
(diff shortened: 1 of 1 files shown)


eiso:
The code can most likely be rearranged in a more elegant way but I am not familiar with the API enough to do so.

PR #671, opened 2017-11-30
worktree.go: fixing the behaviour of switching branches inside an empty git repo

eiso:
When you have an empty git repo (`git init`), `go-git` was creating a branch, with it's associated references. When a repository is empty (before its first commit), only a symbolic reference (entry in `.git/HEAD`) is allowed since there is no parent commit. This fix changes it to a symbolic reference for empty repositories.

*Help with testing is needed*


mcuadros:
Replciating this issue at git, I observer exactly the same behaviour that we have right now:

```
>mkdir foo 
> cd foo/  
> git init
Initialized empty Git repository in /tmp/foo/.git/
> git remote add foo git@github.com:git-fixtures/empty.git
> git fetch foo
> git checkout master
error: pathspec 'master' did not match any file(s) known to git.
```

eiso:
The issue is not with what you have tried. The issue is with having no branches (after git init) and then checking out a new branch that is not master. `go-git` fails on this behaviour currently, since it does not create a symbolic reference but instead tries to create a hard reference instead.

worktree.go
```diff
@@ -140,6 +140,13 @@ func (w *Worktree) Checkout(opts *CheckoutOptions) error {
 		return err
 	}
 
+	c, err := w.getCommitFromCheckoutOptions(opts)
+	if err == plumbing.ErrReferenceNotFound {
```
orirawlings: Should we handle other values for `err`?

orirawlings: I think we also want to check `opts.Create` here. If false, I think we want to continue to error out.

```
$ git init
$ cat .git/HEAD
ref: refs/heads/master
$ cat .git/refs/heads/master
cat: .git/refs/heads/master: No such file or directory
$ git checkout aBranch
error: pathspec 'aBranch' did not match any file(s) known to git.
$ cat .git/HEAD
ref: refs/heads/master
$ git checkout -b aBranch
Switched to a new branch 'aBranch'
$ cat .git/HEAD
ref: refs/heads/aBranch
$ cat .git/refs/heads/aBranch
cat: .git/refs/heads/aBranch: No such file or directory
```

PR #686, opened 2017-12-09
git: worktree, add Grep() method for git grep

darkowlzz:
This change implemented grep on worktree with options to **~match ignoring case~, invert match and specify pathspec**. Also, a **commit hash or reference** can be used to specify the worktree to search.

### DIFF
--- a/options.go
+++ b/options.go
@@ -2,6 +2,7 @@ package git
 
 import (
 	"errors"
+	"regexp"
 
 	"gopkg.in/src-d/go-git.v4/config"
 	"gopkg.in/src-d/go-git.v4/plumbing"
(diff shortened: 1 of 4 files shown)


options.go
```diff
@@ -365,3 +365,42 @@ type ListOptions struct {
 type CleanOptions struct {
 	Dir bool
 }
+
+// GrepOptions describes how a grep should be performed.
+type GrepOptions struct {
+	// Pattern is the match pattern.
+	Pattern string
```
mcuadros: Itstead of accepting a string, I would prefer a `*regexp.Regexp` with this the people can speed that the process with precompilled expressions.

options.go
```diff
@@ -365,3 +365,42 @@ type ListOptions struct {
...
+	// IgnoreCase enables case insensitive match.
+	IgnoreCase bool
```
mcuadros: If we accept the type instead of the string, this is not need it.

worktree.go
```diff
@@ -711,6 +713,118 @@ func (w *Worktree) Clean(opts *CleanOptions) error {
...
+	}
+
+	// Create a Regexp object using the PathSpec.
+	regPathSpec, err := regexp.Compile(opts.PathSpec)
+	if err != nil {
+		return nil, err
+	}
+
+	// Iterate through the files and look for any matches.
+	err = fileiter.ForEach(func(file *object.File) error {
```
mcuadros: can we put this code in a private function?

worktree_test.go
```diff
@@ -1317,3 +1317,194 @@ func (s *WorktreeSuite) TestAlternatesRepo(c *C) {
...
+					TreeName:   "6ecf0ef2c2dffb796033e5a02219af86ec6584e5",
+				},
+				{
+					FileName:   "vendor/foo.go",
+					LineNumber: 3,
+					Content:    "import \"fmt\"",
+					TreeName:   "6ecf0ef2c2dffb796033e5a02219af86ec6584e5",
+				},
+			},
+		},
```
mcuadros: I prefer the open and close brace in the same line.

darkowlzz:
* Changed `GrepOptions` to accept precompiled regexp objects for `Pattern` and `PathSpec`
* Moved the file iteration and pattern matching code to a separate private function `findMatchInFiles()`
* Changed the table test struct format
* Updated compatibility doc

PR #687, opened 2017-12-09
check .ssh/config for host and port overrides; fixes #629

### LINKED ISSUE #629: Usage with a host alias set ssh config
I'm getting a DNS resolution error to github when I have an alias for a particular remote origin url.

For example, let's say I have two git configs:
```
remote.origin.url=git@git1:user1/foo
remote.origin.url=git@git2:user2/bar
```

These have corresponding ssh configs:
```
Host git1
Hostname github.com
IdentityFile ~/.ssh/foo
User user1

Host git2
Hostname github.com
IdentityFile ~/.ssh/bar
User user2
```

Operations using go-git (I've only tried Pull but I assume others too), don't respect the ssh configs and fail with `dial tcp: lookup git1 on 169.254.169.254:53: no such host`. I'm not passing an Auth into the PullOptions. Am I doing something wrong or is this expected?

smithrobs:
I ran into this issue and figured I'd send a PR of what I ended up doing.

Per https://github.com/golang/go/issues/18781 there is no built-in parser for `.ssh/config` so I introduce a dependency on https://github.com/kevinburke/ssh_config

Let me know if you had something else in mind or don't agree with the additional dep.

### DIFF
--- a/plumbing/transport/ssh/common.go
+++ b/plumbing/transport/ssh/common.go
@@ -4,11 +4,13 @@ package ssh
 import (
 	"fmt"
 	"reflect"
+	"strconv"
 
 	"gopkg.in/src-d/go-git.v4/plumbing/transport"
 	"gopkg.in/src-d/go-git.v4/plumbing/transport/internal/common"
 
 	"golang.org/x/crypto/ssh"
+	"github.com/kevinburke/ssh_config"
 )
 
 // DefaultClient is the default SSH client.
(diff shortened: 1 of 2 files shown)


mcuadros:
Thanks @smithrobs, I added a test and fixed a couple of bugs.

smithrobs:
Excellent! Thanks for cleaning up my commit and maintaining this lib! 😄

PR #698, opened 2017-12-20
plumbing: cache, enforce the use of cache in packfile decoder

jfontan:
Decoder object can make use of an object cache to speed up processing.
Previously the only way to specify it was changing manually the struct
generated by NewDecodeForFile. This lead to some instances to be created
without it and penalized performance.

Now the cache should be explicitly passed to the constructor function.
NewDecoder now creates objects with a cache using the default size.

A new helper function was added to create cache objects with the default
size as this becomes a common task now:

    cache.NewObjectLRUDefault()

### DIFF
--- a/plumbing/cache/object_lru.go
+++ b/plumbing/cache/object_lru.go
@@ -24,6 +24,11 @@ func NewObjectLRU(maxSize FileSize) *ObjectLRU {
 	return &ObjectLRU{MaxSize: maxSize}
 }
 
+// NewObjectLRUDefault creates a new ObjectLRU with the default cache size.
+func NewObjectLRUDefault() *ObjectLRU {
+	return &ObjectLRU{MaxSize: DefaultMaxSize}
+}
+
 // Put puts an object into the cache. If the object is already in the cache, it
 // will be marked as used. Otherwise, it will be inserted. A single object might
 // be evicted to make room for the new object.
--- a/plumbing/format/packfile/decoder.go
+++ b/plumbing/format/packfile/decoder.go
@@ -89,9 +89,9 @@ func NewDecoder(s *Scanner, o storer.EncodedObjectStorer) (*Decoder, error) {
 // the specified one, nil will be returned. This is intended to avoid the content
 // deserialization of all the objects.
 //
-// cacheObject is an ObjectLRU that is used to speed up the process. If cache
-// is not needed you can pass nil. To create a cache object with the default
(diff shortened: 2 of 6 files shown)


plumbing/format/packfile/decoder.go
```diff
@@ -80,15 +80,20 @@ type Decoder struct {
...
+		cache.NewObjectLRUDefault())
 }
 
 // NewDecoderForType returns a new Decoder but in this case for a specific object type.
 // When an object is read using this Decoder instance and it is not of the same type of
 // the specified one, nil will be returned. This is intended to avoid the content
-// deserialization of all the objects
+// deserialization of all the objects.
+//
+// cacheObject is an ObjectLRU that is used to speed up the process. If cache
```
ajnavarro: The cache instance can be any implementation of cache.Object

plumbing/format/packfile/decoder.go
```diff
@@ -80,15 +80,20 @@ type Decoder struct {
...
+// is not needed you can pass nil. To create a cache object with the default
+// size you an use the helper cache.ObjectLRUDefault().
```
ajnavarro: s/an/can/

plumbing/format/packfile/decoder.go
```diff
@@ -101,8 +106,9 @@ func NewDecoderForType(s *Scanner, o storer.EncodedObjectStorer,
 	}
 
 	return &Decoder{
-		s: s,
-		o: o,
+		s:              s,
+		o:              o,
+		DeltaBaseCache: cacheObject,
```
ajnavarro: can you make DeltaBaseCache private?

jfontan: I've addressed these issues in the following commits.

plumbing/cache/object_lru.go
```diff
@@ -24,6 +24,11 @@ func NewObjectLRU(maxSize FileSize) *ObjectLRU {
 	return &ObjectLRU{MaxSize: maxSize}
 }
 
+// NewObjectLRUDefault creates a new ObjectLRU with the default cache size.
+func NewObjectLRUDefault() *ObjectLRU {
```
mcuadros: no tested function

jfontan: Now cache tests are done twice. One with the previous cache and another time with a cache created with `NewObjectLRUDefault`.

PR #699, opened 2017-12-20
Add a setRef version that supports siva fs

jfontan:
siva FS only support openning files in read only or write only mode. The
method SetRef is split in files with an extra version that only writes
the reference. It can be activated with `-tags siva` on building.

### DIFF
--- a/storage/filesystem/internal/dotgit/dotgit_setref.go
+++ b/storage/filesystem/internal/dotgit/dotgit_setref.go
@@ -1,0 +1,43 @@
+// +build !siva
+
+package dotgit
+
+import (
+	"os"
+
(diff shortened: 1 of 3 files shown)


storage/filesystem/internal/dotgit/dotgit_setref.go
```diff
@@ -0,0 +1,43 @@
+// +build !siva
```
ajnavarro: should use a more generic key, like can_truncate or something like this?

jfontan: Yes. I'm going to another pull request more generic.

PR #700, opened 2017-12-21
Add a setRef and rewritePackedRefsWhileLocked versions that supports non rw fs

jfontan:
There are some filesystems that do not support opening the files in read
and write modes at the same time. The method SetRef is split in files with
an extra version that only writes the reference. It can be activated with
`-tags norwfs` on building.

### DIFF
--- a/storage/filesystem/internal/dotgit/dotgit_setref_norwfs.go
+++ b/storage/filesystem/internal/dotgit/dotgit_setref_norwfs.go
@@ -2,9 +2,13 @@
 
 package dotgit
 
-import "gopkg.in/src-d/go-git.v4/plumbing"
+import (
+	"fmt"
 
-// There are some filesystems tha don't support opening files in RDWD mode.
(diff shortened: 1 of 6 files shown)


storage/filesystem/internal/dotgit/dotgit_setref_norwfs.go
```diff
@@ -0,0 +1,17 @@
+// +build norwfs
+
+package dotgit
+
+import "gopkg.in/src-d/go-git.v4/plumbing"
+
+func (d *DotGit) setRef(fileName, content string, old *plumbing.Reference) error {
```
ajnavarro: Can you add some comments here explaining the reason of this method (like on the PR description)?

jfontan:
Do not merge. There are still some changes needed to make it fully compatible.

jfontan:
The implementation is ready. Tests also pass with `go test -tags norwfs ./...`.

PR #712, opened 2018-01-09
Set default pack window size in config

jfontan:
Config is not initialized with the default window size. Without this the window size is 0 and packfile code is unable to generate deltas.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -70,7 +70,7 @@ func NewConfig() *Config {
 		Raw:        format.New(),
 	}
 
-	config.Pack.Window = defaultPackWindow
+	config.Pack.Window = DefaultPackWindow
 
 	return config
 }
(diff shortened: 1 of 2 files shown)


config/config.go
```diff
@@ -64,11 +64,15 @@ type Config struct {
...
 // NewConfig returns a new empty Config.
 func NewConfig() *Config {
-	return &Config{
+	config := &Config{
 		Remotes:    make(map[string]*RemoteConfig),
 		Submodules: make(map[string]*Submodule),
 		Raw:        format.New(),
 	}
+
+	config.Pack.Window = defaultPackWindow
```
mcuadros: please add a test

mcuadros: Can you make `defaultPackWindow` public and document it?

jfontan: Done

PR #715, opened 2018-01-10
Clean reconstructed objects outside pack window

jfontan:
Object walk reconstructs delta objects but these are not cleaned up
after they got out the pack window. Without this change all
reconstructed objects reside in memory.


jfontan:
This change has some side effects that I'm investigating. I'll open a new PR when all is fixed.

PR #716, opened 2018-01-11
Clean reconstructed objects outside pack window

jfontan:
Object walk reconstructs delta objects but these are not cleaned up
after they got out the pack window. Without this change all
reconstructed objects reside in memory.

restoreOriginal call is moved before calling Size(). Now we can not
guarantee that the object is already undeltified.

### DIFF
--- a/plumbing/format/packfile/delta_selector.go
+++ b/plumbing/format/packfile/delta_selector.go
@@ -222,10 +222,16 @@ func (dw *deltaSelector) walk(
 ) error {
 	indexMap := make(map[plumbing.Hash]*deltaIndex)
 	for i := 0; i < len(objectsToPack); i++ {
-		// Clean up the index map for anything outside our pack
-		// window, to save memory.
+		// Clean up the index map and reconstructed delta objects for anything
+		// outside our pack window, to save memory.
 		if i > int(packWindow) {
-			delete(indexMap, objectsToPack[i-int(packWindow)].Hash())
+			obj := objectsToPack[i-int(packWindow)]
+
+			delete(indexMap, obj.Hash())
+
+			if obj.IsDelta() {
+				obj.Original = nil
+			}
 		}
 
 		target := objectsToPack[i]
@@ -261,6 +267,16 @@ func (dw *deltaSelector) walk(
 }
 
 func (dw *deltaSelector) tryToDeltify(indexMap map[plumbing.Hash]*deltaIndex, base, target *ObjectToPack) error {
+	// Original object might not be present if we're reusing a delta, so we
+	// ensure it is restored.
+	if err := dw.restoreOriginal(target); err != nil {
+		return err
+	}
+
+	if err := dw.restoreOriginal(base); err != nil {
+		return err
+	}
+
 	// If the sizes are radically different, this is a bad pairing.
 	if target.Size() < base.Size()>>4 {
 		return nil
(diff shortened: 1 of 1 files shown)


plumbing/format/packfile/delta_selector.go
```diff
@@ -261,6 +267,16 @@ func (dw *deltaSelector) walk(
 }
 
 func (dw *deltaSelector) tryToDeltify(indexMap map[plumbing.Hash]*deltaIndex, base, target *ObjectToPack) error {
+	// Original object might not be present if we're reusing a delta, so we
+	// ensure it is restored.
+	if err := dw.restoreOriginal(target); err != nil {
```
erizocosmico: isn't this making it actually slower than before (even if it uses way less memory)?

jfontan: I've just run `borges pack` with repository https://github.com/numpy/numpy (downloaded locally) and the results are quite close.

* Original version:
  * 203.83user 12.34system 2:28.80elapsed 2.4 Gb RAM
  * 204.40user 12.27system 2:29.30elapsed 2.4 Gb RAM
* With proposed changes:
  * 217.82user 11.97system 2:25.72elapsed 1 Gb RAM
  * 220.06user 11.81system 2:26.28elapsed 1 Gb RAM

It seems that it's a bit slower (user time is bigger) but it spends more system time. Maybe heap allocation.

erizocosmico:
Then lgtm. That little slowdown is a nice tradeoff for half the memory
On Thu, 11 Jan 2018 at 19:19, Javi Fontan <notifications@github.com<mailto:notifications@github.com>> wrote:

@jfontan commented on this pull request.

________________________________

In plumbing/format/packfile/delta_selector.go<https://github.com/src-d/go-git/pull/716#discussion_r161034780>:

> @@ -261,6 +267,16 @@ func (dw *deltaSelector) walk(
 }

 func (dw *deltaSelector) tryToDeltify(indexMap map[plumbing.Hash]*deltaIndex, base, target *ObjectToPack) error {
+       // Original object might not be present if we're reusing a delta, so we
+       // ensure it is restored.
+       if err := dw.restoreOriginal(target); err != nil {


I've just run borges pack with repository https://github.com/numpy/numpy (downloaded locally) and the results are quite close.

  *   Original version:
     *   203.83user 12.34system 2:28.80elapsed 2.4 Gb RAM
     *   204.40user 12.27system 2:29.30elapsed 2.4 Gb RAM
  *   With proposed changes:
     *   217.82user 11.97system 2:25.72elapsed 1 Gb RAM
     *   220.06user 11.81system 2:26.28elapsed 1 Gb RAM

It seems that it's a bit slower (user time is bigger) but it spends more system time. Maybe heap allocation.

mcuadros:
Can we test this with more diverse repository? Not only with the `numpy` repository?

jfontan:
I've did new tests with the following repos:

* **upsilon**: small repository, https://github.com/numpy/numpy
* **numpy**: average repository, https://github.com/numpy/numpy
* **tensorflow**, average repository, https://github.com/tensorflow/tensorflow
* **bismuth**, some files are 100Mb in size, https://github.com/hclivess/Bismuth

The times and memory are only from the push action from local to a new repository. Each test is executed twice and the smaller value for time and memory is selected.

| repository | master | fix memory |
| --------------|-----------|----------------|
| upsilon | 54.2s 387MiB | 53.4s 387MiB |
| numpy | 1m25s 1.9GiB | 1m19s 0.65GiB |
| tensorflow | 3m16s 3.9GiB | 2m59s 1.28GiB |
| bismuth | 12m16s 17.7GiB | 10m37s 3.13GiB |

The code to do the benchmark is here: https://gist.github.com/jfontan/42fbfe0761e5280012d285c1a4290f5d

erizocosmico:
So it's actually faster? And uses half the memory? Nice, great job!

mcuadros:
Networking is envolved, so speed number aren't reliable

jfontan:
I've just updated the table with two small repos.

PR #720, opened 2018-01-16
plumbing: cache, modify cache to delete more than one item to free space

jfontan:
The previous version could only delete the oldest used object. If the
object to cache was bigger than the space freed it could not be added.

Also the decoder adds bases to the cache when they are needed.

This change increases the speed creating indexes 2x.

### DIFF
--- a/plumbing/cache/object_lru.go
+++ b/plumbing/cache/object_lru.go
@@ -51,7 +51,7 @@ func (c *ObjectLRU) Put(obj plumbing.EncodedObject) {
 
 	objSize := FileSize(obj.Size())
 
-	if objSize >= c.MaxSize {
+	if objSize > c.MaxSize {
 		return
 	}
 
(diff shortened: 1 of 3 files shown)


plumbing/cache/object_lru.go
```diff
@@ -55,18 +55,14 @@ func (c *ObjectLRU) Put(obj plumbing.EncodedObject) {
 		return
 	}
 
-	if c.actualSize+objSize > c.MaxSize {
+	for c.actualSize+objSize > c.MaxSize {
```
ajnavarro: Should we check if the object to add is bigger than the cache size to do not remove everithing?

mcuadros:
Will be nice have a test here.

plumbing/cache/object_lru.go
ajnavarro: nevermind, is already checked before the for.

Issue #705, opened 2018-01-01
git clone -n 


Nhoya:
Hi, i was wondering if was possible to reproduce the -n option of git clone

mcuadros:
Should be that easy as skip this line: https://github.com/src-d/go-git/blob/v4.0.0/worktree.go#L181

rykov:
@mcuadros I didn't find where Repository's `clone` func calls `Checkout` on the worktree that you've linked. 

Do you think that it could by implemented by adding a flag to `CloneOptions` that would disable this section of Repository's `clone` even if `r.wt != nil`?

https://github.com/src-d/go-git/blob/v4.0.0/repository.go#L452-L478

rykov:
Something like this: https://github.com/src-d/go-git/compare/master...rykov:master

mcuadros:
Sorry is not the Checkout is the Reset, https://github.com/src-d/go-git/blob/v4.0.0/repository.go#L463

rykov:
@mcuadros I've taken another look at the documentation for `git-clone` and I think the [changes that I've made on my fork](https://github.com/src-d/go-git/compare/master...rykov:master) follow the functionality closer than putting a condition just around `Reset`.

Here's what the docs say for `--recurse-submodules`:

> This option is ignored if the cloned repository does not have a worktree/checkout (i.e. if any of --no-checkout/-n, --bare, or --mirror is given)

https://git-scm.com/docs/git-clone#git-clone---recurse-submodulesltpathspec

I'll add some tests and submit [this](https://github.com/src-d/go-git/compare/master...rykov:master) as a PR

Nhoya:
Pretty interesting, can you please provide an example usage?

rykov:
@Nhoya So the behavior is similar to `git clone --bare`, in that the worktree is not created.  However, with `-n` option, the metadata is still placed under the `.git` subdirectory, rather than in root.

Here's a modification of the basic example from the README with this flag enabled:

```go
Info("git clone -n https://github.com/src-d/go-git")

_, err := git.PlainClone("/tmp/foo", false, &git.CloneOptions{
    URL:        "https://github.com/src-d/go-git",
    Progress:   os.Stdout,
    NoCheckout: true,
})

CheckIfError(err)
```

Nhoya:
So retriving commits with

```
r, _ := git.Clone(memory.NewStorage(), nil, &git.CloneOptions{
    URL:        "https://github.com/src-d/go-git",
[...]

PR #721, opened 2018-01-17
Support for clone without checkout (git clone -n)

rykov:
Based on  #705 issue discussion about supporting `git clone -n` option that doesn't checkout HEAD after cloning a remote repository.  This follows documented `git-clone` behavior of ignoring `--recurse-submodules` when no-checkout option is enabled:

> This option is ignored if the cloned repository does not have a worktree/checkout (i.e. if any of --no-checkout/-n, --bare, or --mirror is given)

src: https://git-scm.com/docs/git-clone#git-clone---recurse-submodulesltpathspec

### DIFF
--- a/repository_test.go
+++ b/repository_test.go
@@ -448,6 +448,28 @@ func (s *RepositorySuite) TestPlainCloneWithRecurseSubmodules(c *C) {
 	c.Assert(cfg.Submodules, HasLen, 2)
 }
 
+func (s *RepositorySuite) TestPlainCloneNoCheckout(c *C) {
+	dir, err := ioutil.TempDir("", "plain-clone-no-checkout")
+	c.Assert(err, IsNil)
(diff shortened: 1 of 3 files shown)


repository_test.go
```diff
@@ -448,6 +448,24 @@ func (s *RepositorySuite) TestPlainCloneWithRecurseSubmodules(c *C) {
 	c.Assert(cfg.Submodules, HasLen, 2)
 }
 
+func (s *RepositorySuite) TestPlainCloneNoCheckout(c *C) {
```
mcuadros: can we ensure that HEAD is pointing to the commit should be pointing?

rykov: @mcuadros Take a look now - I've added a HEAD check

rykov:
@mcuadros When you get a chance, can you please tag this into a release?

PR #724, opened 2018-01-19
plumbing: packfile, Add a buffer to crc writer

jfontan:
crc update with block smaller than 16 bytes uses a slower version of the
function. ReadByte is heavily used by zlib inflate so most of the time
crc is updated byte by byte.

A new Flush method is added to the scanner to flush this crc writer
cache. It is only called when the Scanner reader is a teeReader.

Times to push localy from different repositories:

| repo | unbuffered | buffered |
|------|------------|----------|
| cangallo | 139.21ms | 152.13ms |
| octoprint-tft | 2.84s | 2.92s |
| numpy | 47.29s | 44.94s |
| tensorflow | 2m17.94s | 1m56.92s |
| bismuth | 8m13.79s | 7m01.46s |

I will try to make more tests with small repos if needed. The numbers in these repos changed a lot when using the laptop to test.

### DIFF
--- a/plumbing/format/packfile/scanner.go
+++ b/plumbing/format/packfile/scanner.go
@@ -63,10 +63,7 @@ func NewScanner(r io.Reader) *Scanner {
 
 	crc := crc32.NewIEEE()
 	return &Scanner{
-		r: &teeReader{
-			newByteReadSeeker(seeker),
-			crc,
-		},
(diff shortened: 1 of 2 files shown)


mcuadros:
@jfontan please read `CONTRIBUTING.md`

jfontan:
I've updated commit message to conform to contributing guidelines. An extra commit was added with explicit checking of CRC calculated by the scanner.

PR #731, opened 2018-01-24
plumbing: format/packfile, fix crash with cycle deltas

jfontan:
Resolving cycles relied on `ObjectToPack` objects having Original. This is no longer true with the changes from #720. This commit changes:

* Save original type, hash and size in `ObjectToPack`
* Use `SetObject` to set both Original and resolved type, hash and size
* Restore original object before using `BackToOriginal` (cycle resolution)
* Update encoder test to check this case

### DIFF
--- a/plumbing/format/packfile/object_pack.go
+++ b/plumbing/format/packfile/object_pack.go
@@ -53,7 +53,7 @@ func newDeltaObjectToPack(base *ObjectToPack, original, delta plumbing.EncodedOb
 
 // BackToOriginal converts that ObjectToPack to a non-deltified object if it was one
 func (o *ObjectToPack) BackToOriginal() {
-	if o.IsDelta() {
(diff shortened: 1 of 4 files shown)


plumbing/format/packfile/object_pack.go
```diff
@@ -71,11 +77,24 @@ func (o *ObjectToPack) WantWrite() bool {
 	return o.Offset == 1
 }
 
+// SetOriginal sets both Original and saves size, type and hash
+func (o *ObjectToPack) SetOriginal(obj plumbing.EncodedObject) {
+	o.Original = obj
+	o.originalSize = obj.Size()
```
ajnavarro: check if obj is nil

jfontan: Added checks to SetOriginal and  BackToOriginal. The test now also uses it to set Original to nil.

Issue #456, opened 2017-06-30
add support for directories on git add (was: git add . doesn't seem to work)


jucardi:
I'm trying to add all contents in a cloned repo. Normally I would do this with the command line by doing a
```
git add .
```

I'm trying to do it by running

```go
if _,err = workTree.Add("."); err != nil {
```

But that actually returns an error saying

```
read [path]: is a directory
```

Do I need to add file by file by hand?

mcuadros:
[Worktree.Add](https://godoc.org/gopkg.in/src-d/go-git.v4#Worktree.Add) only works with files, directories are not handled, but PRs are welcome.

jucardi:
@mcuadros sounds good, thanks for your help, I'll submit directory a PR with directory support sometime next week.

halfcrazy:
@jucardi What's the current status of your PR?

joshbetz:
I've been doing this, but it's pretty slow.

```
	status, err := worktree.Status()
	if err != nil {
		return err
	}

	if status.IsClean() {
		return nil
	}

	for path, _ := range status {
		w.Add(path)
	}
```

Labutin:
Yes, it really TOO slow.
I need to commit about 500 updated files and calling worktree.Add(path) for all files takes more than 3 minutes :(
Using CommitOptions.All doing the same loop over all modified files and works the same slow.
Any ideas how to commit 500 updated files faster?

mcuadros:
The complexity of the `Add` is almost trivial, the time consumed is on doing the hashing. How big are the files? How many time takes on git command, @Labutin?

Labutin:
I have about 500 modified files. Each file about 22KB size. Not so big.
```
dlabutin@dlabutin-mbp $ git status| wc
     529    1085   26114
dlabutin@dlabutin-mbp $ time git add .
real	0m0.092s
user	0m0.041s
sys	0m0.046s
```

Console `git add .` takes less than 0.1 sec.
But
```
	if _, err := w.Commit("Commin message", &git.CommitOptions{
		All: true,
		Author: &object.Signature{
			Name: "octopus",
			Email: "octopus@server",
			When: time.Now(),
		},
	}); err != nil {
		return err
	}

```
works about 3 minutes.
I am testing on Macbook Pro (SSD drive).

Labutin:
Looks like problem is here: https://github.com/src-d/go-git/blob/master/worktree_status.go#L248

[...]

PR #739, opened 2018-02-01
new methods Worktree.[AddGlob|RemoveBlob] and recursive Worktree.[Add|Remove]

mcuadros:
Fixes: https://github.com/src-d/go-git/issues/456

### DIFF
--- a/worktree_status.go
+++ b/worktree_status.go
@@ -5,6 +5,8 @@ import (
 	"errors"
 	"io"
 	"os"
+	"path"
+	"path/filepath"
 
 	"gopkg.in/src-d/go-billy.v4/util"
 	"gopkg.in/src-d/go-git.v4/plumbing"
@@ -24,10 +26,8 @@ var (
 	// the worktree.
 	ErrDestinationExists = errors.New("destination exists")
 	// ErrGlobNoMatches in an AddGlob if the glob pattern does not match any
-	// file in the worktree.ErrNotDirectory
+	// files in the worktree.
 	ErrGlobNoMatches = errors.New("glob pattern did not match any files")
-	// ErrNotDirectory  in an AddDirectory if the path is not a directory.
-	ErrNotDirectory = errors.New("path is not a directory")
 )
 
 // Status returns the working tree status.
@@ -252,8 +252,12 @@ func diffTreeIsEquals(a, b noder.Hasher) bool {
 
 // Add adds the file contents of a file in the worktree to the index. if the
 // file is already staged in the index no error is returned. If a file deleted
-// from the Workspace is given, the file is removed from the index.
+// from the Workspace is given, the file is removed from the index. If a
+// directory given, adds the files and all his sub-directories recursively in
+// the worktree to the index. If any of the files is already staged in the index
+// no error is returned. When path is a file, the blob.Hash is returned.
 func (w *Worktree) Add(path string) (plumbing.Hash, error) {
+	// TODO(mcuadros): remove plumbing.Hash from signature at v5.
 	s, err := w.Status()
 	if err != nil {
 		return plumbing.ZeroHash, err
@@ -264,7 +268,16 @@ func (w *Worktree) Add(path string) (plumbing.Hash, error) {
 		return plumbing.ZeroHash, err
 	}
 
-	added, h, err := w.doAdd(idx, s, path)
+	var h plumbing.Hash
+	var added bool
+
+	fi, err := w.Filesystem.Lstat(path)
+	if err != nil || !fi.IsDir() {
+		added, h, err = w.doAddFile(idx, s, path)
+	} else {
+		added, err = w.doAddDirectory(idx, s, path)
+	}
+
 	if err != nil {
 		return h, err
 	}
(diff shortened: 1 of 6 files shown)


worktree_status.go
```diff
@@ -243,30 +251,170 @@ func diffTreeIsEquals(a, b noder.Hasher) bool {
...
+		return h, nil
+	}
+
+	return h, w.r.Storer.SetIndex(idx)
+}
+
+// AddDirectory adds the files contents of a directory and all his
+// sub-directories recursively in the worktree to the index. If any of the
+// file is already staged in the index no error is returned.
+func (w *Worktree) AddDirectory(path string) error {
```
orirawlings: This API seems a little clumsy to me. Is there a reason we can simply detect and support directory paths as part of the current `Add` function? A `path string` is already sufficient for a user to describe either directories or files.

Moreover, if we are going to provide separate functions for directories on `Add` we should also include separate functions `MoveDirectory` and `RemoveDirectory` for consistency.

In any case, I think we should simply extend the existing `Add`, `Move`, and `Remove` to support directory paths.

worktree_status.go
```diff
@@ -243,30 +251,170 @@ func diffTreeIsEquals(a, b noder.Hasher) bool {
...
+	}
+
+	return
+
+}
+
+// AddGlob given a glob pattern adds all the matching files content and all his
+// sub-directories recursively in the worktree to the index. If any of the
+// file is already staged in the index no error is returned.
+func (w *Worktree) AddGlob(pattern string) error {
```
orirawlings: If we are going to support `AddGlob` we should also add a `RemoveGlob` for consistency in the API.

mcuadros: Nice point.

worktree_status.go
mcuadros: what should I return when I `Add` a repository? Because the signature returns a `Hash`.

orirawlings: It seems like `Add` returns the hash of the blob object that was added to the repository. If one is adding a directory, it might make sense to return the hash of the corresponding tree object.

mcuadros: `Add` doesn't create any tree object.

orirawlings: Ah yes. I see that. I guess this works then.

mcuadros:
@orirawlings PTAL

mcuadros:
@orirawlings PTAL

worktree_status.go
```diff
@@ -243,30 +250,147 @@ func diffTreeIsEquals(a, b noder.Hasher) bool {
...
-			h, err = w.deleteFromIndex(path)
+			added = true
+			h, err = w.deleteFromIndex(idx, path)
 		}
-		return h, err
+
+		return
 	}
 
 	if s.File(path).Worktree == Unmodified {
```
orirawlings: Would it be an optimization to move this check before the `w.copyFileToStorage(path)` call?

worktree_status.go
```diff
@@ -378,26 +488,92 @@ func (w *Worktree) doUpdateFileToIndex(e *index.Entry, filename string, h plumbi
...
+	return w.Filesystem.Remove(path)
+}
+
+func (w *Worktree) doRemoveFile(idx *index.Index, path string) (plumbing.Hash, error) {
+	hash, err := w.deleteFromIndex(idx, path)
 	if err != nil {
 		return plumbing.ZeroHash, err
 	}
 
+	if err := w.deleteFromFilesystem(path); err != nil {
```
orirawlings: Maybe just:

```
return hash, w.deleteFromFilesystem(path)
```

worktree_status.go
```diff
@@ -409,6 +585,43 @@ func (w *Worktree) deleteFromFilesystem(path string) error {
 	return err
 }
 
+// RemoveGlob given a glob pattern removes all the matching files content and
```
orirawlings: We should probably avoid the masculine pronoun, "his", here too:

> RemoveGlob removes all paths, matching pattern, from the index. If pattern matches a directory path, all directory contents are removed from the index recursively.

worktree_status.go
```diff
@@ -409,6 +585,43 @@ func (w *Worktree) deleteFromFilesystem(path string) error {
...
+// all his sub-directories recursively in the worktree to the index.
+func (w *Worktree) RemoveGlob(pattern string) error {
+	files, err := util.Glob(w.Filesystem, pattern)
+	if err != nil {
+		return err
+	}
+
+	if len(files) == 0 {
```
orirawlings: I'm curious here how we might use this method for a paths that might still exist in the repository or index, but might have already been removed from the working directory. Will this check cause this method to return an error without attempting to remove them from the index?

Example:
```
$ git init
$ touch a; git add a; git commit -m first
[master (root-commit) 5d36496] first
 1 file changed, 0 insertions(+), 0 deletions(-)
 create mode 100644 a
$ rm a; git status
On branch master
Changes not staged for commit:
  (use "git add/rm <file>..." to update what will be committed)
  (use "git checkout -- <file>..." to discard changes in working directory)

        deleted:    a

no changes added to commit (use "git add" and/or "git commit -a")
$ git rm -- a*; git status
rm 'a'
On branch master
Changes to be committed:
  (use "git reset HEAD <file>..." to unstage)

        deleted:    a

```

worktree_status.go
```diff
@@ -409,6 +585,43 @@ func (w *Worktree) deleteFromFilesystem(path string) error {
...
+			return err
+		}
+	}
+
+	return w.r.Storer.SetIndex(idx)
+}
+
 // Move moves or rename a file in the worktree and the index, directories are
 // not supported.
 func (w *Worktree) Move(from, to string) (plumbing.Hash, error) {
```
orirawlings: Are we planning to wait for now on adding support for directories to this method?

worktree_status.go
```diff
@@ -18,9 +20,14 @@ import (
...
-// ErrDestinationExists in an Move operation means that the target exists on
-// the worktree.
-var ErrDestinationExists = errors.New("destination exists")
+var (
+	// ErrDestinationExists in an Move operation means that the target exists on
+	// the worktree.
+	ErrDestinationExists = errors.New("destination exists")
+	// ErrGlobNoMatches in an AddGlob if the glob pattern does not match any
+	// file in the worktree.ErrNotDirectory
+	ErrGlobNoMatches = errors.New("glob pattern did not match any files")
```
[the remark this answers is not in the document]
mcuadros: What match is files, so I thing is ok with files

worktree_status.go
mcuadros: Add TODO

mcuadros:
@orirawlings PTAL

worktree_status.go
orirawlings: @mcuadros this seems like a bug to me. Can you confirm?

In `git add`, you can modify entries in the index even if the glob matches no paths in the current working directory. But this check looks like it would return an error before trying to update the index if the glob has no matches in the working directory.
```
$ git init
$ touch a; git add a
$ rm a; ls a*
$ git add -- a*; echo $?
0
```

worktree_status.go
```diff
@@ -243,30 +251,147 @@ func diffTreeIsEquals(a, b noder.Hasher) bool {
...
+		return err
+	}
+
+	idx, err := w.r.Storer.Index()
+	if err != nil {
+		return err
+	}
+
+	var saveIndex bool
+	for _, file := range files {
```
orirawlings: This also looks like we aren't trying to match patterns on things in the index, only paths that exist in the working directory. This is slightly different behavior than `git add -- <glob>` in cases where index entries should be removed (i.e. `git add` supports both adding AND removing entries from the index).

mcuadros: Add, only adds files from the woktree to the index. The function `Add` should not do exactly the behavior of `git add`. This is a programmatically interface, so I rather keep it simple.

worktree_status.go
mcuadros: I think this is a very complex behavior, the goal of this PR was speed up multiple `Add` calls.

worktree_status.go
orirawlings: I think that makes sense. I'm just anticipating that this discrepancy will cause confusion for users and drive more issues being opened.

inge4pres:
Hi 😄 
can't understand if this is adding the support for recursive deletion of not empty folders, I am trying to do it with 
```worktree.Filesystem.Remove(path)``` where path is a folder containing some files and I'm getting error ```remove /tmp/path: directory not empty``` (using the clone with OS storage)

I looked a bit into billy's code and there is a `RemoveAll` function mimicking the standard lib's `os.RemoveAll` but it looks like in worktree.go it's only used in the `rmFileAndDirIfEmpty` function...

Any ideas?
Thanks

PR #743, opened 2018-02-08
Fix TestUploadPackWithContextOnRead test in plumbing/transport/test/upload_pack.go

kuba--:
Signed-off-by: Kuba Podgórski <kuba.podgorski@gmail.com>


kuba--:
Discard is an io.Writer on which all Write calls succeed without doing anything.

kuba--:
context was canceled (we should wait till context is `Done`)

PR #744, opened 2018-02-09
plumbing: format/packfile, fix panic retrieving object hash.

jfontan:
In some cases the original data is not saved before it is cleaned and forces a panic when it's retrieved.

The change adds `ObjectToPack.CleanOriginal` to be used to clean original object instead of:

    object.Original = nil

Now when the Original data is freed because it's no longer in the pack window a `SetOriginal` call is done to make sure that Size, Hash and Size data is not lost.

Fixes https://github.com/src-d/borges/issues/229

### DIFF
--- a/plumbing/format/packfile/delta_selector.go
+++ b/plumbing/format/packfile/delta_selector.go
@@ -103,7 +103,7 @@ func (dw *deltaSelector) objectsToPack(
 
 		otp := newObjectToPack(o)
 		if _, ok := o.(plumbing.DeltaObject); ok {
-			otp.Original = nil
+			otp.CleanOriginal()
 		}
 
 		objectsToPack = append(objectsToPack, otp)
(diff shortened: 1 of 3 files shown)


plumbing/format/packfile/delta_selector.go
```diff
@@ -231,7 +231,8 @@ func (dw *deltaSelector) walk(
 			delete(indexMap, obj.Hash())
 
 			if obj.IsDelta() {
-				obj.Original = nil
+				obj.SetOriginal(obj.Original)
+				obj.CleanOriginal()
```
ajnavarro: this is a little bit strange, maybe is worth it to add a method that just add the original metadata to the objectToPack, something like `obj.SetOriginalMetadata(obj.Original)`

jfontan: I've adede a `SaveOriginalMetadata` function. It uses current `Original` object.

Issue #750, opened 2018-02-14
Clone fails for https://gitlab.com/agamigo/material


dave:
The git command line tool is able to clone this repo:

```
$ git clone https://gitlab.com/agamigo/material
Cloning into 'material'...
warning: redirecting to https://gitlab.com/agamigo/material.git/
remote: Counting objects: 1016, done.
remote: Compressing objects: 100% (559/559), done.
remote: Total 1016 (delta 504), reused 679 (delta 306)
Receiving objects: 100% (1016/1016), 216.49 KiB | 564.00 KiB/s, done.
Resolving deltas: 100% (504/504), done.
```

... however go-git fails:

```go
func TestClone(t *testing.T) {
	_, err := git.Clone(memory.NewStorage(), memfs.New(), &git.CloneOptions{
		URL:               "https://gitlab.com/agamigo/material",
		SingleBranch:      true,
		Depth:             1,
		RecurseSubmodules: git.DefaultSubmoduleRecursionDepth,
	})
	if err != nil {
		t.Fatal(err)
	}
}
```

Output:

```
getter_test.go:23: repository not found
```

dave:
I assume it's something to do with the line `warning: redirecting to https://gitlab.com/agamigo/material.git/` in the command line `git` output?

PR #751, opened 2018-02-14
transport: http, fix services redirecting only info/refs

mcuadros:
Some git HTTP services (like GitLab), do not redirect all the git HTTP endpoints, they only do for the `info/refs` path. In order to solve this problem, we change the endpoint for the subsequence requests info the info/refs` requests was redirected.

Fixes: https://github.com/src-d/go-git/issues/750


dave:
I can confirm this fixes https://github.com/src-d/go-git/issues/750

PR #759, opened 2018-02-25
plumbing: ssh, return error when creating public keys from invalid PEM

mdelillo:
If the `pemBytes` passed in is not valid, [`pem.Decode` will return nil](https://golang.org/src/encoding/pem/pem.go?s=2199:2247#L68). Passing that to `x509.IsEncryptedPEMBlock` will panic when it [tries to read `Headers`](https://golang.org/src/crypto/x509/pem_decrypt.go?s=2314:2357#L89). This fix returns an error early rather than panicking.


mcuadros:
Please, can you Signed-off your commit? @mdelillo

mdelillo:
Sure thing

PR #762, opened 2018-02-28
plubming: transport, Escape the user and pswd for endpoint. Fixes #723

### LINKED ISSUE #723: Basic Auth: Encode special characters in password
Dear Maintainers,

first of all, thanks a lot for your work. This library is awesome!
I'm not quite sure if my finding is really a bug or intended. Feel free to close this issue if it's intended.

For my current project I like to ask the user for the git repository url and credentials.
Before I actually clone the git repo, I would like to show the user if the given url and credentials are valid and if I'm able to clone the repository in the next step. 
This was actually quite straight forward with small information from other (closed) issues. 

The main issue I had was the authentication via basic auth. I use a very "secure" password which also includes special characters like "@" and "?" and so on. Every time I used my "secure" password with special characters, I got an access denied error. Please see this code as an example:

```go
// Create new endpoint
ep, err := transport.NewEndpoint(repo.URL)
if err != nil {
        os.Exit(1)
}

if repo.Username != "" && repo.Password != "" {
	ep.User = repo.Username
	ep.Password = repo.Password
}

// Create client
cl, err := client.NewClient(ep)
if err != nil {
        os.Exit(1)
}

// Open new session
s, err := cl.NewUploadPackSession(ep, nil)
if err != nil {
        os.Exit(1)
}
defer s.Close()

// Get advertised references (e.g. branches)
// ep.User = ""
// ep.Password = ""
ar, err := s.AdvertisedReferences()
...
```

If you run this code with a password that has special characters, you will get a nice access denied.
But if you clear the password and username before the command `s.AdvertisedReferences()` is executed, it will work. (see the last two lines which are commented)

Why is that happening? 
The problem is that the basic auth credentials will be added multiple times and for one routine the password is not URL encoded which is required. 
The first time it will be added:
https://github.com/src-d/go-git/blob/f6aca0820fb0a44d4c4875a21581a09609d63a6b/plumbing/transport/http/common.go#L116

And the second time:
https://github.com/src-d/go-git/blob/f6aca0820fb0a44d4c4875a21581a09609d63a6b/plumbing/transport/http/common.go#L34

The second time is actually the bad part. This will generate the complete URL without actually encoding the password and is adding the basic auth credentials the second time without being required.

zkry:
When endpoint is generating a url string I made sure that the username and password are escaped properly using the url.PathEscape function. There is also a test to make sure the string representation of the Git URL is a valid URL by parsing it.

Signed-off-by: Zachary Romero <zacromero3@gmail.com>

### DIFF
--- a/plumbing/transport/common.go
+++ b/plumbing/transport/common.go
@@ -128,10 +128,10 @@ func (u *Endpoint) String() string {
 		buf.WriteString("//")
 
 		if u.User != "" || u.Password != "" {
-			buf.WriteString(u.User)
+			buf.WriteString(url.PathEscape(u.User))
 			if u.Password != "" {
 				buf.WriteByte(':')
-				buf.WriteString(u.Password)
+				buf.WriteString(url.PathEscape(u.Password))
 			}
 
 			buf.WriteByte('@')
--- a/plumbing/transport/common_test.go
+++ b/plumbing/transport/common_test.go
@@ -1,6 +1,7 @@
 package transport
 
 import (
+	"net/url"
 	"testing"
 
 	"gopkg.in/src-d/go-git.v4/plumbing/protocol/packp/capability"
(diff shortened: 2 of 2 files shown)


mcuadros:
Nice!

PR #768, opened 2018-03-03
Added support for non-symlink checkouts on Windows when elevated rights are missing

fkollmann:
On Windows, elevated rights are required for symlinks.

If they are missing, which is true for almost all users, the library will fail with "A required privilege is not held by the client." (ERROR_PRIVILEGE_NOT_HELD).

This change will handle the error and write the file, instead of trying to create a symlink.

### DIFF
--- a/worktree.go
+++ b/worktree.go
@@ -9,6 +9,7 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
+	"syscall"
 
 	"gopkg.in/src-d/go-git.v4/config"
 	"gopkg.in/src-d/go-git.v4/plumbing"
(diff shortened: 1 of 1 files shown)


mcuadros:
This is how behaves git on windows?

fkollmann:
No, there is an ongoing fix at Go: https://github.com/golang/go/issues/22874

This will simply work around the issue until fixed in Go itself.

fkollmann:
I will close this PR in favor of an implementation that mimics the behavior of Go on Windows: https://github.com/src-d/go-git/pull/776

PR #771, opened 2018-03-05
repository.Log: add alternatives for commit traversal order

ilius:
The new option `Order` to `repo.Log` allows customizing the order of commits.
Specially because the current algorithm is not compatible with `git log`, the current algorithm use a stack (Depth-first Search?), while `git log` uses timestamp of commit (committer time).

So if we want to have the same order as `git log` we specify `Order: git.LogOrderCommitterTime`.
In the CommitterTime implementation, the **stack of iterators** is replaced with a **BinaryHeap of commits** (to avoid sorting the commits every time a commit is added). Also `start` property is removed (we start with a heap of length 1) so the code is simpler and cleaner.

For the BinaryHeap we use this library: https://github.com/emirpasic/gods

The default behavior (commits order) of `repo.Log` is unchanged.

### DIFF
--- a/options.go
+++ b/options.go
@@ -307,12 +307,27 @@ func (o *ResetOptions) Validate(r *Repository) error {
 	return nil
 }
 
+type LogOrder int8
+
+const (
+	LogOrderDefault LogOrder = iota
+	LogOrderDFS
+	LogOrderDFSPost
+	LogOrderBSF
+	LogOrderCommitterTime
+)
+
 // LogOptions describes how a log action should be performed.
 type LogOptions struct {
 	// When the From option is set the log will only contain commits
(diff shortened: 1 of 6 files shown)


ilius:
The failing test does not seem to be related to my changes

```
/home/travis/gopath/src/gopkg.in/src-d/go-git.v4/plumbing/transport/test/upload_pack.go:55:
    c.Assert(err, Equals, transport.ErrRepositoryNotFound)
... obtained *errors.errorString = &errors.errorString{s:"unexpected EOF"} ("unexpected EOF")
... expected *errors.errorString = &errors.errorString{s:"repository not found"} ("repository not found")

```

PR #776, opened 2018-03-09
Added support for symlink checkouts on Windows

fkollmann:
On Windows, elevated rights are required for symlinks.

If they are missing, which is true for almost all users, the library will fail with "A required privilege is not held by the client." (ERROR_PRIVILEGE_NOT_HELD).

This implementation mimics the behavior of Git on Windows.


worktree.go
```diff
@@ -540,6 +541,20 @@ func (w *Worktree) checkoutFile(f *object.File) (err error) {
 	return
 }
 
+func isSymlinkWindowsNonAdmin(err error) bool {
+	const ERROR_PRIVILEGE_NOT_HELD syscall.Errno = 1314
+
+	if err != nil {
+		if x, ok := err.(*os.LinkError); ok {
```
mcuadros: can you give a proper name to the variable?

worktree.go
```diff
@@ -540,6 +541,20 @@ func (w *Worktree) checkoutFile(f *object.File) (err error) {
...
+			if xx, ok := x.Err.(syscall.Errno); ok {
```
mcuadros: same here

mcuadros:
Sorry @fkollmann the "Signed-off-by" is done using the "-s" flag in all the PR commits.

fkollmann:
Okay, so every commit has to be signed-off... I will copy the changes to another branch and create a new pull request, later.

fkollmann:
Closed in favor of https://github.com/src-d/go-git/pull/778

PR #783, opened 2018-03-14
Fix RefSpec.Src()

maguro:
Previously, the Src() function was assuming there are no “+” characters in the refspec src
and erroneously used the strings.Index() function to compute the start index of src in the
refspec.  This change now uses the IsForceUpdate() method to decide how to elide the
force update token.

Signed-off-by: Alan Cabrera <adc@toolazydogs.com>


maguro:
So, I don't have karma to merge the request.  Can someone do that for me?

maguro:
May thanks!

Issue #780, opened 2018-03-14
Tests in the root package are too slow


mvdan:
They should either be made faster (parallel tests, or other fixes), or the slowest should be skipped on `testing.Short()`.

At the moment, this is far from ideal:

```
$ go test -short
OK: 197 passed
PASS
ok      gopkg.in/src-d/go-git.v4        58.933s
```

I would have expected 10 seconds at most. Ideally, 2-3 seconds or less.

mvdan:
Here's another - the tests aren't standard Go, so I can't use out of the box features like test's `-v` and `-run`. Is there a reason for that?

mcuadros:
We want to change it to use asserty, as any other modern development at source{d}. But the current test suite is built on top of [gocheck](https://labix.org/gocheck), this decision was made 3yrs ago, when was the default test framework for us, now we have a bunch of tests to change it, so its a lot of work for a little benefit.

mvdan:
I honestly don't mind whatever test framework is used. But in my opinion, the end result is that regular Go tests are run, so that `go test` flags like `-v` and `-run` are useful. Otherwise, `go test` is little other than running some arbitrary Go program that fails or succeeds.

mcuadros:
Yep, but sadly this, flags are "ignored" by gocheck. Since the whole testsuite runs under the same `T`

PR #784, opened 2018-03-14
add PlainOpen variant to find .git in parent dirs

### LINKED ISSUE #765: A version of PlainOpen that detects the root of the repository
If I have a repository at `~/foo`, I can run `git status` in `~/foo/bar` without a problem.

However, with any little tool that I write with this library, I use `git.PlainOpen(".")`. This works fine if I'm at the root of the git repo, but quickly breaks down if I'm in any of its subdirectories.

I propose that a new `Open` variant be added that would iterate over the parent directories, stopping at the first containing a `.git` directory or file. This could be `func PlainOpenDetect(from string) (*Repository, error)`, for example, to not have it depend on `os.Getwd`.

There's also the question of whether anyone would require the current `PlainOpen` versus this improved one. If the answer is not, perhaps the behavior of `PlainOpen` could simply be changed.

Happy to submit a patch with tests if this seems like a good idea.

/cc @mcuadros @smola

mvdan:
This is the git tool's behavior that people are used to; if one runs a
git command in a repository's subdirectory, git still works.

Fixes #765.

### DIFF
--- a/options.go
+++ b/options.go
@@ -420,3 +420,14 @@ func (o *GrepOptions) Validate(w *Worktree) error {
 
 	return nil
 }
+
+// PlainOpenOptions describes how opening a plain repository should be
+// performed.
+type PlainOpenOptions struct {
+	// DetectDotGit defines whether parent directories should be
+	// walked until a .git directory or file is found.
+	DetectDotGit bool
+}
+
+// Validate validates the fields and sets the default values.
+func (o *PlainOpenOptions) Validate() error { return nil }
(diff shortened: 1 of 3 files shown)


mvdan:
I've given up on the tests at this point; I'm on slow airport wifi and they've been running for about five minutes. I would normally just run the tests I'm interested in, or hope for faster and more robust tests, but see https://github.com/src-d/go-git/issues/780.

mcuadros:
@mvdan any update?

mvdan:
Thanks for the ping - still need to follow-up on this.

mvdan:
@mcuadros should be fine now; tests pass locally.

mcuadros:
awesome!

PR #789, opened 2018-03-23
blame: add commit hash to blame result

shanedasilva:


mcuadros:
Can you add a test for this?

PR #790, opened 2018-03-24
Makefile: -, Fixed to retry test cases when the test cases has failed.

TsutomuNakamura:
Tests sometimes fails due to "unexpected EOF" like the URL.
https://travis-ci.org/src-d/go-git/jobs/357483531#L1546-L1549

I found that this error occured by test cases under "plumbing/transport/ssh" directory.
But I haven't been able to find cause of this error yet(I guess, the cause of the error is something due to test environment).
Although it is a provisional countermeasure, fixed to retry test cases when the error occurred.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -38,9 +38,27 @@ test-coverage:
 	@cd $(WORKDIR); \
 	echo "" > $(COVERAGE_REPORT); \
 	for dir in `find . -name "*.go" | grep -o '.*/' | sort | uniq`; do \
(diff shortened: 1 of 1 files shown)


TsutomuNakamura:
After fixed:
https://travis-ci.org/TsutomuNakamura/go-git/jobs/357783620#L1583

mcuadros:
I don't think is the best solution, we should work on fixing the error and not hide it. Thanks

TsutomuNakamura:
Certainly. Thank you for your consideration.

Issue #791, opened 2018-03-26
CheckClose should be used with named returns


jfontan:
If `defer CheckClose(closer, &err)` is used without named returns any close error is lost. This happens here:

https://github.com/src-d/go-git/blob/master/plumbing/format/packfile/common.go#L50

Check if there are any other place where `CheckClose` is not used with a named return.

PR #794, opened 2018-03-27
Use CheckClose with named returns and fix tests

jfontan:
Previously some close errors were losts. This is especially problematic in go-git as lots of work is done here like generating indexes and moving packfiles.

Some tests write to an already existing repository retrieved from fixtures. The permissions of these files are read only and make receive pack fail. This was shadowed before as close errors were lost.

Fixes: #791

### DIFF
--- a/storage/filesystem/internal/dotgit/dotgit.go
+++ b/storage/filesystem/internal/dotgit/dotgit.go
@@ -375,7 +375,7 @@ func (d *DotGit) findPackedRefsInFile(f billy.File) ([]*plumbing.Reference, erro
 	return refs, s.Err()
 }
 
-func (d *DotGit) findPackedRefs() (r []*plumbing.Reference, error error) {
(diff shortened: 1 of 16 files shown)


storage/filesystem/internal/dotgit/dotgit.go
```diff
@@ -375,7 +375,7 @@ func (d *DotGit) findPackedRefsInFile(f billy.File) ([]*plumbing.Reference, erro
 	return refs, s.Err()
 }
 
-func (d *DotGit) findPackedRefs() ([]*plumbing.Reference, error) {
+func (d *DotGit) findPackedRefs() (r []*plumbing.Reference, error error) {
```
ajnavarro: this one is called `error` instead of `err`, and you are using err on the code. Is a leftover or am I missing something?

jfontan: It's my sloppy fingers, thanks! Just fixed it.

PR #795, opened 2018-03-28
Add commit hash to blame result

shanedasilva:


smola:
@shanedasilva I don't know, we'll have to wait for maintainers to check it.
cc @mcuadros @strib @orirawlings

Issue #793, opened 2018-03-26
Clone is slow when repo contains a lot of files


glibsm:
## Problem
I've noticed that `PlainClone` was taking a long time in my use case.

In order to try and reproduce the problem I created a repo containing 17k files
with randomly generated data in them: https://github.com/glibsm/alphabet.

Below is the code used and the resulting performance numbers.

Any help is appreciated. Haven't dug into the go-git library code just yet, figured
I'd share the info I have first.

## Sample code

```go
package main

import (
        "flag"
        "fmt"
        "log"
        "os"
        "runtime/pprof"

        git "gopkg.in/src-d/go-git.v4"
)

var cpuprofile = flag.String("cpuprofile", "", "write cpu profile to file")

func main() {
        flag.Parse()
        if *cpuprofile != "" {
                f, err := os.Create(*cpuprofile)
                if err != nil {
                        log.Fatal(err)
                }
                pprof.StartCPUProfile(f)
                defer pprof.StopCPUProfile()
        }

        cloneOpts := &git.CloneOptions{
                URL:      "https://github.com/glibsm/alphabet", // repo with 17k files
                Progress: os.Stdout,
        }
        repo, err := git.PlainClone("./repo", false, cloneOpts)
        if err != nil {
                log.Panic(err)
        }

        head, err := repo.Head()
        if err != nil {
                log.Panic(err)
        }
        fmt.Println(head)
}
```

## Running the sample code

```
Counting objects: 36579, done.
Compressing objects: 100% (1423/1423), done.
Total 36579 (delta 4), reused 36574 (delta 3), pack-reused 0
978d628f9d82456dffef0bb0579b8a617750c8db refs/heads/master
./plain --cpuprofile plain.cpu  22.18s user 14.97s system 140% cpu **26.404 total**
```

## Command line clone for speed comparison

```
$ time git clone https://github.com/glibsm/alphabet
Cloning into 'alphabet'...
remote: Counting objects: 36579, done.
remote: Compressing objects: 100% (1423/1423), done.
remote: Total 36579 (delta 4), reused 36574 (delta 3), pack-reused 0
Receiving objects: 100% (36579/36579), 1.56 MiB | 5.98 MiB/s, done.
Resolving deltas: 100% (4/4), done.
Checking out files: 100% (17579/17579), done.

real    0m3.569s
user    0m0.508s
sys     **0m2.826s**
```

As we can see, the same repo clone times are **2.8 vs 26** seconds, which is quite a big gap.

[...]

PR #800, opened 2018-04-04
format: packfile, avoid allocating in teeReader.ReadByte

nussjustin:
ReadByte allocates a byte slice on each call which showed up in memory profiles for the code in #793. Avoiding the allocation saves 3 seconds on my machine.

Updates #793


nussjustin:
Seems like the problem was already fixed on tip, ops

PR #803, opened 2018-04-09
config: adds branches to config for tracking branches against remotes…

### LINKED ISSUE #313: Setup tracking for default branch when cloned.
Is it possible to set up tracking of remote branches, similarly to how `git clone` does for the default branch? I've found the relevant configuration [mentioned here](https://github.com/src-d/go-git/blob/867b10692e5f8a34a82cc0a783bdb63e2b5ff398/config/config_test.go#L20), but I *think* that's just testing that configuration is marshalled and unmarshalled without losing information.

It this is not possible currently, what would be the most appropriate way to implement this? Should `Config` gain methods for manipulating these particular settings, should `Clone` / `PlainClone` do this by default, etc, etc.

Any guidance would be much appreciated :)

TheHipbot:
Cloning against a branch will also track that branch against the remote. Allows the ability to create and modify branch tracking in git config.

* Adds Branches to Config
* Updates Repository.clone to include tracking against branch when clone is against a remote branch
* Updates Repository.clone to track against master if no Reference is specified in CloneOptions
* Adds Repository.CreateBranch to create new branch tracking on repository
* Adds Repository.Branch to lookup branch tracking on repository
* Adds Repository.DeleteBranch to delete branch tracking on repository
…, updates clone to track when cloning a branch. Fixes #313

Signed-off-by: Jeremy Chambers <jeremy@thehipbot.com>


config/branch.go
```diff
@@ -0,0 +1,71 @@
...
+	raw *format.Subsection
+}
+
+// Validate validates fields of branch
+func (b *Branch) Validate() error {
+	if b.Name == "" {
+		return errBranchEmptyName
+	}
+
+	if b.Merge != "" && !strings.HasPrefix(b.Merge, "refs/heads/") {
```
mcuadros: Check this using [`ReferenceName.IsBranch`](https://godoc.org/gopkg.in/src-d/go-git.v4/plumbing#ReferenceName.IsBranch)

config/config.go
```diff
@@ -56,6 +56,10 @@ type Config struct {
 	// of the submodule, should equal to Submodule.Name.
 	Submodules map[string]*Submodule
 
+	// Branches list of branches, the key is the branch name and should
+	// equal Branch.Name
+	Branches map[string]*Branch
+
```
mcuadros: Please remove this extra line and also the 58.

config/config.go
```diff
@@ -87,19 +92,31 @@ func (c *Config) Validate() error {
 		}
 	}
 
+	for name, b := range c.Branches {
```
mcuadros: This is not covered by tests.

config/branch.go
```diff
@@ -0,0 +1,71 @@
...
+
+// Branch contains information on the
+// local branches and which remote to track
+type Branch struct {
+	// Name name of branch
+	Name string
+	// Remote name of remote to track
+	Remote string
+	// Merge is the local refspec for the branch
+	Merge string
```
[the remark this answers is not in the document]
TheHipbot: Thanks for the quick feedback, I'll update the PR as soon as I can

PR #807, opened 2018-04-13
dotgit: ignore filenames that don't match a hash

strib:
For both packfiles and object files.

Issue: keybase/client#11366


mcuadros:
Can you add a test?

strib:
@mcuadros added a simple test.

PR #819, opened 2018-04-27
Fix host redirect handling

DisposaBoy:
If you try to clone from a url that redirects to another host e.g. `example.com/repo` -> `github.com/org/repo.git`, it results in a redirect to `example.com/org/repo.git` because ModifyEndpointIfRedirect only takes into account the scheme and path.

### DIFF
--- a/plumbing/transport/http/common.go
+++ b/plumbing/transport/http/common.go
@@ -4,6 +4,7 @@ package http
 import (
 	"bytes"
 	"fmt"
+	"net"
 	"net/http"
 	"strconv"
 	"strings"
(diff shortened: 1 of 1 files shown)


DisposaBoy:
I'm unable to sign the DCO because the sign-off process is weird, so feel free to consider this a bug report and and use the code as you wish.

DisposaBoy:
@smola IMO, the whole process could be simplified to not require contributors to add their email address to commit messages and/or only do it once.
But to be clear, the weird part that I object(ed) to is the bolded part of the following paragraph from your [CONTRIBUTING.md](https://github.com/src-d/go-git/blob/master/CONTRIBUTING.md#certificate-of-origin)

> In order to show your agreement with the DCO you should include at the end of commit message, the following line: Signed-off-by: John Doe <john.doe@example.com>, **using your real name**.

smola:
@DisposaBoy Thanks for the clarification. I understand that contributing under a pseudonym is a very legitimate use case. But so far, we are not able to accept that under our current legal guidelines.

I have moved this to an issue: https://github.com/src-d/go-git/issues/820
And opened an issue for ourselves to check if we can accept pseudonymous contributions in the future: https://github.com/src-d/guide/issues/208

However, according to our present policy, I have to close this PR.

PR #821, opened 2018-05-01
plumbing: packp, Report Status should continue decoding after band 1 is flushed. Fixes #787

### LINKED ISSUE #787: No sideband data stored for push
This seems like it could be a regression since https://github.com/src-d/go-git/pull/573. I haven't been able to see any sideband data for push for a few months now, while clone stores sideband data just fine. I should've reported this earlier, I thought it was user error at first (hey, maybe it still is!)

Steps to reproduce:

1. Run [the progress example](https://github.com/src-d/go-git/blob/a39fae69325f5a1faf8fe1b512124f3b77e8612d/_examples/progress/main.go), and see that sideband works fine with clone.
2. Add `Progress: os.Stdout` to the push options in [the push example](https://github.com/src-d/go-git/blob/a39fae69325f5a1faf8fe1b512124f3b77e8612d/_examples/push/main.go) and see no sideband data written to `os.Stdout`.
3. Push to GitHub using [the original implementation's example](https://github.com/src-d/go-git/pull/573#issue-138816868) and see no sideband data written to `os.Stdout`.

Let me know if I can provide any more details. Thanks!

Werelds:
As described on #787 ReportStatus breaks out before decoding all bands in responses. While I understand the reasoning behind it, it also means that anything on band 2 will never be decoded.

This PR changes that behaviour to keep decoding after the first flush, allowing band 2 to still be output.

I decided to keep this within ReportStatus, as that will be used across transports and is in line with the official Git client. That also always decodes sidebands if support is enabled.

### DIFF
--- a/plumbing/protocol/packp/report_status.go
+++ b/plumbing/protocol/packp/report_status.go
@@ -19,11 +19,23 @@ const (
 type ReportStatus struct {
 	UnpackStatus    string
 	CommandStatuses []*CommandStatus
+	withSideband    bool
 }
 
 // NewReportStatus creates a new ReportStatus message.
+// without support for reading additional withSideband data.
 func NewReportStatus() *ReportStatus {
-	return &ReportStatus{}
+	return &ReportStatus{
+		withSideband: false,
+	}
+}
+
+// NewReportStatusWithSideband creates a new ReportStatus message,
+// with support for reading additional withSideband data.
+func NewReportStatusWithSideband() *ReportStatus {
+	return &ReportStatus{
+		withSideband: true,
+	}
 }
 
 // Error returns the first error if any.
@@ -57,8 +69,7 @@ func (s *ReportStatus) Encode(w io.Writer) error {
 	return e.Flush()
 }
 
-// Decode reads from the given reader and decodes a report-status message. It
-// does not read more input than what is needed to fill the report status.
+// Decode reads from the given reader and decodes a report-status message.
 func (s *ReportStatus) Decode(r io.Reader) error {
 	scan := pktline.NewScanner(r)
 	if err := s.scanFirstLine(scan); err != nil {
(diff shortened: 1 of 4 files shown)


mcuadros:
Rebase and take a look to the ` ReceivePackSuite.TestSendPackAddDeleteReference` because is failing due to this change.

Werelds:
Rebased it and ran the tests, but it seems to be fine here?

Attached my shell output and a code coverage file created ~2 minutes ago.

- [go-test.log](https://github.com/src-d/go-git/files/1968854/go-test.log)
- [coverage.txt](https://github.com/src-d/go-git/files/1968853/coverage.txt)

I was on the latest commit anyway, only your commit today wasn't in my fork.

Werelds:
Strange, it does fail on Travis. I can't seem to reproduce that. I'll see if I can run it in one of the images Travis uses and debug it that way.

mcuadros:
I have the same fail failure on my laptop are you using Linux or OSX? I am using Arch.

Werelds:
Ubuntu via WSL, actually. I'll fire up my Macbook and give it a go on that.

Clearly something must be different under WSL that prevents this from working. Wonder what though, it can't be the TCP stack itself. Threading perhaps.

Werelds:
Alright, getting the issue on my Mac. I'll debug on that and see what's going on.

Werelds:
Ok, figured out what's going on. The test in question is without sideband support, so the connection gets closed before it attempts reading sideband support, without it emitting an EOF.

I'll make some amends to be able to differentiate between the two scenarios without breaking backwards compatibility.

Werelds:
I'm going to close this PR, do more digging and re-submit at a later date. When sideband support is enabled and the capabilities are set accordingly, it still fails due to the connection closing too early.

I'll have to dig deeper into the internals of the library to see where exactly this happens. The connection should not be closed if there is still data to be read, but somewhere some bit of code decides that it's seen enough and does exactly that.

Issue #772, opened 2018-03-05
ResolveRevision doesn't resolve annotated tags


rykov:
Looking through the `revision.Ref` case for the switch statement in `ResolveRevision`, the tag is resolved, however, calling `r.CommitObject` fails with "object not found" error because the resolved commit is not a commit Hash. Seems like one needs to call `r.resolveToCommitHash` either before `CommitObject` or inside of it (depending on the intended API)

https://github.com/src-d/go-git/blob/ecda5c1512bcb19e1802d629b18872ec995e23cf/repository.go#L914-L930

mcuadros:
Can you provide an example repository? Thanks

rykov:
Here's a quick way to reproduce this:

```sh
$ git init 
$ git commit --allow-empty -m "Initial commit"
$ git tag -a ann-tag -m "Annotated!"
$ git tag reg-tag
```

And then run:

```go
package main

import(
  "gopkg.in/src-d/go-git.v4"
  "gopkg.in/src-d/go-git.v4/plumbing"
  "fmt"
)

func main() {
  r, _ := git.PlainOpen(".")
  rev1, err := r.ResolveRevision(plumbing.Revision("reg-tag"))
  fmt.Println("Regular tag: ", rev1, err)
  rev2, err := r.ResolveRevision(plumbing.Revision("ann-tag"))
  fmt.Println("Annotated tag: ", rev2, err)
}
```

The output that I see with v4.1.1 is:

```
Regular tag:  7e454763ed760efa060937a9fd603a60b9fd2879 <nil>
Annotated tag:  0000000000000000000000000000000000000000 object not found
```

orirawlings:
It seems like `git rev-parse` would resolve to the hash of the `tag` object in this case. So we probably want to mimic that behavior as well.

rykov:
If you mean the hash of the annotated tag object, then it makes sense, but we also then need a public method to unpeel that to a commit hash.  Something similar to `r.resolveToCommitHash`

pinoniq:
In the meantime, is there another way to resolve annotated tags using go-git?

pinoniq:
I did some digging and I managed to get it working "kind of".

repository.ResolveRevision expects the resolved reference to be a commit. This is the case for lightweight tags. But not at all for annotated tags.

replacing
```
			if ref != nil {
				refCommit, rErr = r.CommitObject(ref.Hash())
			} else {
				rErr = plumbing.ErrReferenceNotFound
			}
```
[...]

PR #830, opened 2018-05-11
Teach ResolveRevision how to look up annotated tags

novas0x2a:
This should address #772. The existing tests aren't broken by this, but I'm not quite sure how to add an annotated tag to the fixtures to add a new test for this.


mcuadros:
Well, the tests are passing. Means that you managed to solve it, or is that were failing in your local environment?

Can you cover this change by a test?

novas0x2a:
Any updates on this or on https://github.com/src-d/go-git-fixtures/issues/1 , which this PR is kind of blocked on?

novas0x2a:
@mcuadros?

mcuadros:
Sorry for the late response, the fixtures can't be regenerated are immutable, if you want to cover a new case that can't be done using any of the current fixtures, you need to submit a PR with the new fixture. 

The dotgit folders are a tarball of the `.git` and the name is the sha1 of the tar file. The packfiles are just the packfile.

The fixtures are based on https://github.com/git-fixtures/

PR #848, opened 2018-05-31
config: modules, worktree: Submodule fixes for CVE-2018-11235

josephvusich:
This PR contains two fixes addressing [CVE-2018-11235](https://nvd.nist.gov/vuln/detail/CVE-2018-11235):

1. Ignore submodules containing "`..`" in the module path.
2. Do not allow `.gitmodules` to be a symlink.

Both of these changes mirror fixes made by the canonical Git project in response to the CVE.

References:
  * https://blogs.msdn.microsoft.com/devops/2018/05/29/announcing-the-may-2018-git-security-vulnerability/
  * https://security-tracker.debian.org/tracker/CVE-2018-11235
  * https://github.com/git/git/commit/10ecfa76491e4923988337b2e2243b05376b40de
  * https://github.com/git/git/commit/0383bbb9015898cbc79abd7b64316484d7713b44

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -135,7 +135,7 @@ func (c *Config) Unmarshal(b []byte) error {
 	if err := c.unmarshalPack(); err != nil {
 		return err
 	}
-	c.unmarshalSubmodules()
+	unmarshalSubmodules(c.Raw, c.Submodules)
 
 	if err := c.unmarshalBranches(); err != nil {
 		return err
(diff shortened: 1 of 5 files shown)


dustin-decker:
Thanks for the quick patching on this. 

I'm just starting to use this library and i'm curious: Does go-git support hooks or ever plans to? If someone were to recursively clone a malicious repository using go-git, could code exec have been possible? It'd be nice to be using a git library that doesn't have the surface area of hooks to worry about. There have already been a few code execution exploits that relied on git hooks and when building software to process git repositories, more of them is a concern.

smola:
@dustin-decker We do not support hooks and we  have no plans to do it.

We might consider supporting hooks if someone is willing to contribute it. But I think that, in that case, it would be opt-in.

PR #849, opened 2018-06-01
blame: fix edge case with missing \n in content length causing mismat…

shanedasilva:
…ched length error


mcuadros:
Can you provide a test?

PR #861, opened 2018-06-12
setAuth (lowercase) seals the interface preventing extension

keynan:


.gitignore
```diff
@@ -2,3 +2,4 @@ coverage.out
 *~
 coverage.txt
 profile.out
+.idea
```
smola: Please, do not add further entries in gitignore. You can use a [global gitignore](https://help.github.com/articles/ignoring-files/#create-a-global-gitignore) for this.

smola:
I think we should do this test, but this needs to pass tests as well as signing-off the commit.

PR #864, opened 2018-06-15
storage/filesystem: avoid norwfs build flag

smola:
norwfs build flag was used to work on filesystems that do not support neither opening a file in read/write mode or renaming a file (e.f. sivafs).

This had two problems:

- go-git could not be compiled to work properly both with regular filesystems and limited filesystems at the same time.
- the norwfs trick was not available on Windows.

This PR removes the norwfs build flag, as well as the windows conditional flag on the dotgit package.

For the file open mode, we use the new billy capabilities, to check at runtime if the filesystem supports opening a file in read/write mode or not.

For the renaming, we just try and fallback to alternative methods if billy.ErrNotSupported is returned.

Signed-off-by: Santiago M. Mola <santi@mola.io>


smola:
This depends on billy capabilities, which should be released soon (cc @jfontan). go-billy-siva also needs to be updated to report capabilities properly for this to work properly.

PR #869, opened 2018-06-21
packfile: optimise NewIndexFromIdxFile for a very common case

dsymonds:
Loading from an on-disk idxfile will usually already have the idxfile
entries in order, so check that before wasting time on sorting.


dsymonds:
FYI, I’m running go-git in a large production environment and it spends a
large (double digit) percent of its CPU time doing the sort here, which I
believe is almost always unnecessary.

smola:
@dsymonds Right. Packfiles generated with go-git and git should be sorted. Although packfiles as seen in the wild can break a lot of common assumptions.

Thanks for the PR!

PR #870, opened 2018-06-21
Remote.Fetch: error on missing remote reference

mcuadros:
A Fetch from a remote with a non-wildcard refspec should fail.

 ```
git fetch origin "+refs/pull/1/head:refs/pull/1/head"
fatal: Couldn't find remote ref refs/pull/1/head
```

### DIFF
--- a/remote.go
+++ b/remote.go
@@ -630,14 +630,29 @@ func calculateRefs(
 		spec = append(spec, refspecTag)
 	}
 
-	iter, err := remoteRefs.IterReferences()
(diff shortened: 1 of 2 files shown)


smola:
@fkorotkov This looks like a regression. Would you mind opening a new issue describing your problem? If possible, with a code snippet to reproduce it, too. Thanks!

fkorotkov:
@smola I proposed a fix here: https://github.com/src-d/go-git/pull/910

PR #872, opened 2018-06-25
plumbing: Reduce unbounded recursion in packfile gen

nalanj:
While working with some large-ish repos I noticed that repacking (and packfile generation in general) can hit some really nasty issues due to unbounded recursion. This PR cuts out the two cases I noticed by flattening the recursions.

Signed-off-by: Alan Johnson <alan.johnson@gmail.com>

### DIFF
--- a/plumbing/format/packfile/delta_selector.go
+++ b/plumbing/format/packfile/delta_selector.go
@@ -135,12 +135,17 @@ func (dw *deltaSelector) encodedObject(h plumbing.Hash) (plumbing.EncodedObject,
 
 func (dw *deltaSelector) fixAndBreakChains(objectsToPack []*ObjectToPack) error {
 	m := make(map[plumbing.Hash]*ObjectToPack, len(objectsToPack))
-	for _, otp := range objectsToPack {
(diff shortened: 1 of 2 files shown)


commondream:
Looking at that coverage reduction.

On the seen handling I was doing, it seems like you are using `WantsWrite()` for similar, but when I was repacking real repos I still hit a delta cycle. Do you think that points toward an issue in `WantsWrite()` or is my seen handling a reasonable way to go forward?

commondream:
@mcuadros @smola Is there a good way to review total coverage on my own computer (like codecov would do) so I can see if a patch hits that section?

smola:
@commondream Some editors show coverage per line. I use [Visual Studio Code](https://code.visualstudio.com/), where you can toggle code coverage visibility after you run tests right from the editor. [IntelliJ IDEA](https://www.jetbrains.com/idea/) and [Atom](https://atom.io/) also do this if setup correctly.

mcuadros:
`go test -coverprofile=cover.out  -v  && go tool cover -html=cover.out`

PR #880, opened 2018-07-03
repository: added cleanup for PlainClone. Fixes #741

### LINKED ISSUE #741: After failing PlainClone .git directory nevertheless exists
### What version of Go are you using (`go version`)?
go version go1.9.3 darwin/amd64

### What did you do?

```go
package main

import (
	"log"
	"os"
	"gopkg.in/src-d/go-git.v4"
	"gopkg.in/src-d/go-git.v4/plumbing"
)

func main() {
	_, err := git.PlainClone("repository", false, &git.CloneOptions{
		URL:           "https://github.com/git-fixtures/basic.git",
		ReferenceName: plumbing.ReferenceName("notexistingbranch"),
	})
	if err != nil {
		if _, err := os.Stat("repository/.git"); err != nil {
			if os.IsNotExist(err) {
				log.Print("Not Exists")
			} else {
			}
		}
		log.Print("Exists")
	}
}
```

### What did you expect to see?
I expexted to see, that the .git directory does not exist after a failing pull

### What did you see instead?
I saw that the directory exist, when the error is not nil.

bjaroszewski:
This PR introduces the cleanup for the PlainClone().

While using PlainClone, if the clone will raise an error (f.e. incorrect credentials used for authentication) we have work dir left.


repository.go
```diff
@@ -334,7 +335,23 @@ func dotGitFileToOSFilesystem(path string, fs billy.Filesystem) (bfs billy.Files
 // if the new repository will be bare or normal. If the path is not empty
 // ErrRepositoryAlreadyExists is returned.
 func PlainClone(path string, isBare bool, o *CloneOptions) (*Repository, error) {
-	return PlainCloneContext(context.Background(), path, isBare, o)
+	r, err := PlainCloneContext(context.Background(), path, isBare, o)
+	if err != nil {
```
smola: Cleanup logic should be the same for `PlainCloneContext`

repository.go
```diff
@@ -334,7 +335,23 @@ func dotGitFileToOSFilesystem(path string, fs billy.Filesystem) (bfs billy.Files
...
+		case transport.ErrAuthorizationFailed:
+			os.RemoveAll(path)
+			return nil, err
+		case ErrRepositoryAlreadyExists:
+			return nil, err
+		default:
+			os.Remove(path)
+			return nil, err
+		}
+	}
```
smola: nitpick: add a space here (we usually add a space after every if/for block.

repository.go
```diff
@@ -334,7 +335,23 @@ func dotGitFileToOSFilesystem(path string, fs billy.Filesystem) (bfs billy.Files
...
+		case transport.ErrAuthenticationRequired:
+			os.RemoveAll(path)
+			return nil, err
+		case transport.ErrAuthorizationFailed:
+			os.RemoveAll(path)
+			return nil, err
+		case ErrRepositoryAlreadyExists:
+			return nil, err
+		default:
+			os.Remove(path)
```
smola: Why the default case is `Remove` and auth errors are `RemoveAll`?

repository.go
```diff
@@ -334,7 +334,13 @@ func dotGitFileToOSFilesystem(path string, fs billy.Filesystem) (bfs billy.Files
 // if the new repository will be bare or normal. If the path is not empty
 // ErrRepositoryAlreadyExists is returned.
 func PlainClone(path string, isBare bool, o *CloneOptions) (*Repository, error) {
-	return PlainCloneContext(context.Background(), path, isBare, o)
+	r, err := PlainCloneContext(context.Background(), path, isBare, o)
```
smola: Keep it in a single place. If you implement it in `PlainCloneContext`, then that's it, you don't need it here.

repository_test.go
```diff
@@ -577,6 +577,24 @@ func (s *RepositorySuite) TestPlainCloneContext(c *C) {
...
+
+	r, err := PlainCloneContext(ctx, dir, false, &CloneOptions{
+		URL: "incorrectOnPurpose",
+	})
+	c.Assert(r, IsNil)
+	c.Assert(err, NotNil)
+
+	_, err = os.Stat(dir)
+	dirNotExist := os.IsNotExist(err)
+	c.Assert(dirNotExist, Equals, true)
```
smola: The behaviour does not seem correct: it removes a directory that already existed before. go-git shouldn't remove anything it didn't create during clone.

repository_test.go
```diff
@@ -577,6 +578,23 @@ func (s *RepositorySuite) TestPlainCloneContext(c *C) {
...
+	tmpDir := c.MkDir()
+	repoDir := path.Join(path.Dir(tmpDir), "repoDir")
+	r, err := PlainCloneContext(ctx, repoDir, false, &CloneOptions{
+		URL: "incorrectOnPurpose",
+	})
+	c.Assert(r, IsNil)
+	c.Assert(err, NotNil)
+
+	_, err = os.Stat(repoDir)
+	dirNotExist := os.IsNotExist(err)
```
smola: There should be another test to check that if the directory exists, it is not removed by PlainClone.

bjaroszewski:
@mcuadros as per discussion on the slack, I will introduce the changes as @smola suggested for this cleanup on PlainCloneContext(). I should have some free evening this week, so I'll update the PR. Sorry for the delay with it.

repository.go
```diff
@@ -347,12 +348,64 @@ func PlainClone(path string, isBare bool, o *CloneOptions) (*Repository, error)
 // operation is complete, an error is returned. The context only affects to the
 // transport operations.
 func PlainCloneContext(ctx context.Context, path string, isBare bool, o *CloneOptions) (*Repository, error) {
+	dirEmpty := false
+	dirExist := false
+
+	file, err := os.Stat(path)
+	pathNotExist := os.IsNotExist(err)
```
smola: You can write here:

```
file, err := os.Stat(path)
if os.IsNotExist(err) {
  dirExist = true
} else if err != nil {
  return nil, err
}
```

If `os.Stat`returns an error and it is not a not exist error, something is wrong (e.g. permissions) and we should not continue.

repository.go
```diff
@@ -347,12 +348,64 @@ func PlainClone(path string, isBare bool, o *CloneOptions) (*Repository, error)
...
+	if dirExist {
+		fh, err := os.Open(path)
+		if err != nil {
+			return nil, err
+		}
+		defer fh.Close()
+
+		names, err := fh.Readdirnames(1)
+
+		if err == io.EOF && len(names) == 0 {
```
smola: Here if `err` is not nil or `io.EOF`, an error should be returned.

repository.go
```diff
@@ -347,12 +348,64 @@ func PlainClone(path string, isBare bool, o *CloneOptions) (*Repository, error)
...
+				return nil, err
+			}
+
+			for _, name := range names {
+				err = os.RemoveAll(filepath.Join(path, name))
+				if err != nil {
+					return nil, err
+				}
+			}
+		} else if !dirExist {
```
smola: Avoid additional indentation by reversing condition order:

```
if !dirExist {
    os.RemoveAll(path)
    return nil, err
}

// continue here

File: repository_test.go [hunk 10]
```diff
@@ -579,6 +579,38 @@ func (s *RepositorySuite) TestPlainCloneContext(c *C) {
 	c.Assert(err, NotNil)
 }
 
+func (s *RepositorySuite) TestPlainCloneContextWithIncorrectRepo(c *C) {
+	ctx, cancel := context.WithCancel(context.Background())
+	cancel()
+
+	tmpDir := c.MkDir()
+	repoDir := filepath.Join(tmpDir, "repoDir") //path.Join(path.Dir(tmpDir), "repoDir")
```
smola: Remove commented out code.

smola:
CI fails:

```
# gopkg.in/src-d/go-git.v4
./repository.go:373:16: undefined: ErrDirNotEmpty
```

repository.go
smola: Sorry, my snippet was wrong.
Leave your logic as it is if behavior with permission errors is correct.
Do not use `pathNotExist` variable here only to use it for the if condition. Just write `if !os.IsNotExist(err)`.

repository.go
smola: Meh, I confused `dirExist` and  `dirEmpty`, sorry.

repository_test.go
```diff
@@ -576,7 +576,7 @@ func (s *RepositorySuite) TestPlainCloneContextWithProperParameters(c *C) {
 		URL: s.GetBasicLocalRepositoryURL(),
 	})
 
-	c.Assert(r, NoNill)
+	c.Assert(r, NotNill)
```
PaluMacil: still a typo here... you want NotNil, not NotNill

PR #882, opened 2018-07-03
worktree: sort the tree object.  Fixes #881

### LINKED ISSUE #881: tree object not sorted correctly
After making changes to an in-memory clone and then pushing, the remote server (github in this case) complains about the tree sort order and rejects the push.  This issue can be reproduced with "git fsck" as well.

spottybenny:


Xjs:
I just stumbled over this as well.

Thank you, @spottybenny, for that fix, which addresses the problem by sorting a tree object when it is constructed out of an index.

You could also consider sorting the index already. The index is currently being sorted upon encoding, causing index *files* to have the correct order. (Cf. `encodeEntries` in [encoder.go](https://github.com/src-d/go-git/blob/fcfd239c49baee9195f7b8234f46288a29c3951c/plumbing/format/index/encoder.go#L65)
It might be worthwhile to already sort the in-memory representation of the index, which could be done in `(i *Index) Add` in [index.go](https://github.com/src-d/go-git/blob/fcfd239c49baee9195f7b8234f46288a29c3951c/plumbing/format/index/index.go#L57). That would remove the need for sorting in all places where the index is used. (I haven't checked how the git utility does this, this might be interesting as well.)

worktree_commit.go
```diff
@@ -11,6 +11,7 @@ import (
 	"gopkg.in/src-d/go-git.v4/storage"
 
 	"gopkg.in/src-d/go-billy.v4"
+	"sort"
```
smola: Put this import in the first block, along with `"path"` and `"strings"`

worktree_commit_test.go
```diff
@@ -8,9 +8,15 @@ import (
...
 	"gopkg.in/src-d/go-git.v4/storage/memory"
 
+	"bytes"
 	. "gopkg.in/check.v1"
 	"gopkg.in/src-d/go-billy.v4/memfs"
+	"gopkg.in/src-d/go-billy.v4/osfs"
 	"gopkg.in/src-d/go-billy.v4/util"
+	"gopkg.in/src-d/go-git.v4/storage/filesystem"
+	"io/ioutil"
+	"os"
```
smola: Same here, but standard library imports in the first block, together with the others.
Same for go-git imports, put them in the same block as the other go-git imports.

smola:
Sorting at construction would, indeed, be desirable. But the entries slice in the index is public, so we cannot really guarantee that it is always sorted.

spottybenny:
I have addressed the import-related comments, thank you.  With respect to doing the sort at other locations in the code, I'm probably missing something, but correctly sorting requires knowing whether the tree entry is a directory or a not (the correct sort will sort directories as if they have a / at the end, even though they don't actually have the / in their name).  The place I added the sort was the only place I could see where the information about whether the entry was a directory or file could be determined.  I believe this means that any sorts being done at other locations (such as the sort at https://github.com/src-d/go-git/blob/fcfd239c49baee9195f7b8234f46288a29c3951c/plumbing/format/index/encoder.go#L65) are incorrect.

spottybenny:
Looking some more, I see that the sort in encoder.go does have the appropriate information, and so the sort there should probably be fixed.  However, iirc the directory versus file sorting strangeness wasn't involved in the original case I was solving (which happened some months ago, only now getting around to figuring out how to contribute my patch), so I think the sorting also needs to be elsewhere, and in the place where the tree actually was encoded for the commit seemed an appropriate place.  However, if there is a more appropriate place that works I'd be happy to move it.

glibsm:
Looks like the PR has stalled, yet this would be great to get in for next release.
What's currently holding this back (apart from a single failed build)?

smola:
@spottybenny Could you solve the merge conflict?

PR #885, opened 2018-07-05
plumbing: object, return ErrFileNotFound in FindEntry. Fixes #883

### LINKED ISSUE #883: object.Tree.FindEntry should return ErrFileNotFound rather than errEntryNotFound
If directory is not found, it will return the public error `ErrDirNotFound`. For a missing file, it returns `errEntryNotFound`: https://github.com/src-d/go-git/blob/f01958913fab6e1967c1317b7222d1160212371c/plumbing/object/tree.go#L178

Since the error is private there isn't a way to programmatically check for that condition.

jsravn:
FindEntry will return ErrDirNotFound if the directory doesn't exist. But
it doesn't return the corresponding error if the file itself is missing.
This adds the logic to return ErrEntryNotFound, so users can
programmatically check for this condition.


jsravn:
Thanks!

Issue #886, opened 2018-07-06
Get diff of head and incoming git object for pre-receive hook


noxora:
Hi guys,
I've been trying to figure out how to write a pre-receive hook that will figure out the difference between the head that is being pushed to and the incoming changes, and scan the insertions that are made.

I have found that there is a "quarantine" folder that git creates while the pre-receive hook is being run that contains the objects that are to be migrated into the repo. 

My current problem is that I do not know how to create an object using the go-git library that I can use to compare against the most recent commit in the tree.

I wrote a quick python script to verify that this is what is happening 
```[master 4cdaaf6] a
 1 file changed, 1 insertion(+)
Counting objects: 3, done.
Delta compression using up to 8 threads.
Compressing objects: 100% (3/3), done.
Writing objects: 100% (3/3), 282 bytes | 282.00 KiB/s, done.
Total 3 (delta 1), reused 0 (delta 0)
remote: git diff e060e7a47b6d054d8fa6f487822a719270ac7e7b 4cdaaf6268e6ccc9414983421bc97465bbb97cc6
remote: diff --git a/stuff.txt b/stuff.txt
remote: index 556dac7..f951eee 100644
remote: --- a/stuff.txt
remote: +++ b/stuff.txt
remote: @@ -72,3 +72,4 @@ s
remote:  s
remote:  s
remote:  s
remote: +s
remote: 
remote: incoming dir...
remote: 4c
remote: ba
remote: f9
remote: pack
remote: 
To localhost:ldecker/something.git
   e060e7a..4cdaaf6  master -> master```

First, I just exec "git diff" against the two object hashes that I am given on stdin which takes the format
`<old-value> SP <new-value> SP <ref-name> LF`
an example of this is `959c8c3185242eee9fc3d03479e2270d9f8967ca 45cda90346ac64ad726945f6c49b65372db488ac refs/heads/master`

I figure there is a way to do this with the library, and I would like to package my hook into a neat binary with golang. Any help would be much appreciated

**Comment by noxora** [comment]
So I fixed this locally by changing the aforementioned `objectPath` function to something like 
```
func (d *DotGit) objectPath(h plumbing.Hash) string {
	hString := h.String()
	_, err1 := d.fs.Stat(d.fs.Join(objectsPath, hString[0:2], hString[2:40])) //check if file exists
	if err1 != nil {                                                          //if not, look for an "incoming" directory
		directoryContents, err := d.fs.ReadDir(objectsPath)
		if err != nil {
			return ""
[...]

PR #887, opened 2018-07-10
added support for quarantine directory

noxora:
This is a part of issue #886 
I have tested this on my local installation and it has allowed me to create a "commitObject" from an object sitting in a quarantine directory as per the git-receive-pack

This code might be put in the wrong place, or implemented differently than you all would like, so feel free to move/alter it.
The idea of it is to check to see if the file whose path you're returning is actually there, and if not to see if there is an `incoming-XXXXXXX` directory and look for the file in there

### DIFF
--- a/storage/filesystem/dotgit/dotgit.go
+++ b/storage/filesystem/dotgit/dotgit.go
@@ -279,19 +279,66 @@ func (d *DotGit) objectPath(h plumbing.Hash) string {
 	return d.fs.Join(objectsPath, hash[0:2], hash[2:40])
 }
 
+//incomingObjectPath is intended to add support for a git pre-recieve hook to be written
+//it adds support for go-git to find objects in an "incoming" directory, so that the library
+//can be used to write a pre-recieve hook that deals with the incoming objects.
+//More on git hooks found here : https://git-scm.com/docs/githooks
+//More on 'quarantine'/incoming directory here : https://git-scm.com/docs/git-receive-pack
+func (d *DotGit) incomingObjectPath(h plumbing.Hash) string {
+	hString := h.String()
+	directoryContents, err := d.fs.ReadDir(objectsPath)
+	if err != nil {
+		return d.fs.Join(objectsPath, hString[0:2], hString[2:40])
+	}
+	var incomingDirName string
+	for _, file := range directoryContents {
+		if strings.Split(file.Name(), "-")[0] == "incoming" && file.IsDir() {
+			incomingDirName = file.Name()
+		}
+	}
+	if incomingDirName == "" {
+		return d.fs.Join(objectsPath, hString[0:2], hString[2:40])
+	}
+	return d.fs.Join(objectsPath, incomingDirName, hString[0:2], hString[2:40])
+}
+
 // Object returns a fs.File pointing the object file, if exists
 func (d *DotGit) Object(h plumbing.Hash) (billy.File, error) {
-	return d.fs.Open(d.objectPath(h))
+	obj1, err1 := d.fs.Open(d.objectPath(h))
+	if os.IsNotExist(err1) {
+		obj2, err2 := d.fs.Open(d.incomingObjectPath(h))
+		if err2 != nil {
+			return obj1, err1
+		}
+		return obj2, err2
+	}
+	return obj1, err1
 }
 
 // ObjectStat returns a os.FileInfo pointing the object file, if exists
 func (d *DotGit) ObjectStat(h plumbing.Hash) (os.FileInfo, error) {
-	return d.fs.Stat(d.objectPath(h))
+	obj1, err1 := d.fs.Stat(d.objectPath(h))
+	if os.IsNotExist(err1) {
+		obj2, err2 := d.fs.Stat(d.incomingObjectPath(h))
+		if err2 != nil {
+			return obj1, err1
+		}
+		return obj2, err2
+	}
+	return obj1, err1
 }
 
 // ObjectDelete removes the object file, if exists
 func (d *DotGit) ObjectDelete(h plumbing.Hash) error {
-	return d.fs.Remove(d.objectPath(h))
+	err1 := d.fs.Remove(d.objectPath(h))
+	if os.IsNotExist(err1) {
+		err2 := d.fs.Remove(d.incomingObjectPath(h))
+		if err2 != nil {
+			return err1
+		}
+		return err2
+	}
+	return err1
 }
 
 func (d *DotGit) readReferenceFrom(rd io.Reader, name string) (ref *plumbing.Reference, err error) {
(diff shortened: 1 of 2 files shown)


noxora:
So the test error is because I am changing the output in some cases to
`.git/objects/incoming-XXXXXX/hash[0:2]/hash[2:40]`
instead of 
`.git/objects/hash[0:2]/hash[2:40]`
What's confusing me is that this should only change the output in cases where the object IS in the "incoming" folder, it shouldn't change it in scenarios where the object isn't in either.

smola:
@noxora Thanks for the pull request!

It would be good to mention _quarantine environment_ somewhere in the code / comments, with a link to https://git-scm.com/docs/git-receive-pack#_quarantine_environment

This way those of us who didn't heard about this before will not be confused ;)

I still do not see how the whole thing would work to mimick a pre-receive hook. We would need to know more about that before merging this.

For example:
- When would incoming objects be written?
- This should probably be optional to not impact performance when we don't use it.
- When an object does not exist, is it correct to read its incoming counterpart in any operation? Or it should be seen only by certain operations?

noxora:
@smola Thanks for the response, I should have been clearer

This is not intended to mimic a pre-receive hook, it only enables the writing of one using this library. I started writing a hook using this library and when I went to run it, I found that go-git was not finding the object I wanted to compare with inside the quarantine directory. 

This pull request is intended to add support to look for a 'quarantine' directory containing objects so that a pre-receive hook created with this library can work with the incoming objects. 

As per your questions
*In a normal push operation, the remote server receives the objects and puts them into the 'quarantine' directory. It then runs the pre-receive hook, if that hook exits with status code 0, it runs the update hook for each reference to be updated. If that also passes, then it updates the reference by migrating the object out of the quarantine folder (at least, that's how I understand it)

*This should probably only be checked if a 'quarantine' directory might exist, which should only be during a git-receive-pack operation, which I think is only git push. 

*It is probably not correct to attempt to read it in all situations, as it will cause some slight lag. Again, I think it is only required while receiving the contents of a git push. The quarantine directory only exists while the pre-receive and update hooks are being run. However, I am not familiar enough with this library to know exactly how to correct this

noxora:
@smola 
I have spent some time researching what exactly we should be doing in this case, and have reached out to the git mailing list. The conversation is publicly available [here](https://public-inbox.org/git/20180802183924.GB23690@sigill.intra.peff.net/T/) 

Without the environment variables described in Jeff King's response, I believe this is currently the best option available to support quarantine directories.

If there are any changes you would like me to make, I would be happy to, but I would like to know if you would like to move forward with this pull request. If you would, please let me know if there is anything I can do to help with it

noxora:
@mcuadros I apologize if this bothers you, but I haven't heard back from anyone in quite some time. Would you be able to take a look at this?

mcuadros:
@noxora Will be great if you can cover the new functionality with some tests.
Also if you can squash and sign-off your commits.

storage/filesystem/dotgit/dotgit.go
```diff
@@ -279,19 +279,66 @@ func (d *DotGit) objectPath(h plumbing.Hash) string {
...
+		return obj2, err2
+	}
+	return obj1, err1
 }
 
 // ObjectDelete removes the object file, if exists
 func (d *DotGit) ObjectDelete(h plumbing.Hash) error {
-	return d.fs.Remove(d.objectPath(h))
+	err1 := d.fs.Remove(d.objectPath(h))
+	if err1 != nil {
```
smola: Only fallback to incoming object if `os.IsNotExist(err1)`.

storage/filesystem/dotgit/dotgit.go
```diff
@@ -279,19 +279,66 @@ func (d *DotGit) objectPath(h plumbing.Hash) string {
...
+		return obj2, err2
+	}
+	return obj1, err1
 }
 
 // ObjectStat returns a os.FileInfo pointing the object file, if exists
 func (d *DotGit) ObjectStat(h plumbing.Hash) (os.FileInfo, error) {
-	return d.fs.Stat(d.objectPath(h))
+	obj1, err1 := d.fs.Stat(d.objectPath(h))
+	if err1 != nil {
```
smola: Use `os.IsNotExist`

storage/filesystem/dotgit/dotgit.go
```diff
@@ -279,19 +279,66 @@ func (d *DotGit) objectPath(h plumbing.Hash) string {
...
+		return d.fs.Join(objectsPath, hString[0:2], hString[2:40])
+	}
+	return d.fs.Join(objectsPath, incomingDirName, hString[0:2], hString[2:40])
+}
+
 // Object returns a fs.File pointing the object file, if exists
 func (d *DotGit) Object(h plumbing.Hash) (billy.File, error) {
-	return d.fs.Open(d.objectPath(h))
+	obj1, err1 := d.fs.Open(d.objectPath(h))
+	if err1 != nil {
```
smola: Use `os.IsNotExist`

noxora:
Alright, so for some reason `TestObjectDelete` isn't passing on appveyor, and I think it's to do with appveyor's environment. I have commented the test out for now, but it is very similar to `TestObject` and `TestObjectStat` which both pass. 

All three of these tests pass on my macbook, so I am not sure why it is failing there

I would also like to clarify how this change helps with a git-hook.

By allowing the library to see a quarantined directory, one could write a git pre-receive hook using this library to do something like diff the HEAD with the incoming changes and perform some checks on any added/changed logic. 

It is not part of triggering a hook to be run, but I would be happy to add that later in a different PR

storage/filesystem/dotgit/dotgit_test.go
```diff
@@ -537,6 +537,48 @@ func (s *SuiteDotGit) TestObject(c *C) {
...
+	dir := New(fs)
+
+	hash := plumbing.NewHash("03db8e1fbe133a480f2867aac478fd866686d69e")
+	err := dir.ObjectDelete(hash)
+	c.Assert(err, IsNil)
+	//	incomingHash := "9d25e0f9bde9f82882b49fe29117b9411cb157b7" //made up hash
+	//	incomingDirPath := fs.Join("objects", "incoming-123456")
+	//	incomingFilePath := fs.Join(incomingDirPath, incomingHash[0:2], incomingHash[2:40])
+	//	fs.MkdirAll(incomingDirPath, os.FileMode(0755))
+	//	fs.Create(incomingFilePath)
```
smola: You did not create the `incomingHash[0:2]` directory before creating `incomingFilePath`.

Also assert `fs.MkdirAll` and `fs.Create` do not return errors.

Maybe the problem is related to that. Or maybe something related to `os.IsNotExist` on Windows.

noxora: I tried those changes, and it also failed so I'm not exactly sure what's going on.

The changes are included, but commented out

PR #892, opened 2018-07-18
plumbing/object: fix pgp signature encoder/decoder

jfontan:
The way of reading pgp signatures was searching for pgp begin line in the header. This caused problems when this string appeared and was not part of the signature. For example if it appears in the message as an example or is part of the author name the decoder starts treating it as a signature. In this state the code was not able to notice then the header ended so it entered in an infinite loop searching for pgp end string.

Now it uses the same method as original git. Searches for gpgsig section in header and starts getting all lines until the next part.

In encoder the string used to add signatures was incorrect. It is now changed to the proper `gpgsig` string instead of `pgpsig`.

### DIFF
--- a/plumbing/object/commit.go
+++ b/plumbing/object/commit.go
@@ -17,8 +17,9 @@ import (
 )
 
 const (
-	beginpgp string = "-----BEGIN PGP SIGNATURE-----"
-	endpgp   string = "-----END PGP SIGNATURE-----"
+	beginpgp  string = "-----BEGIN PGP SIGNATURE-----"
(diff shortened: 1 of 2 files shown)


ajnavarro:
ping @jfontan

jfontan:
@smola I believe it does. Now gpgsign header read is stopped when the line is empty or its first character is not space. https://github.com/src-d/go-git/pull/892/files#diff-7eac18321111ba95fe06592bca6b5112R185

This fixes the infinite loop past EOF we had before. I can not think on other way to exploit signatures.

PR #898, opened 2018-07-19
Feature/new packfile parser

jfontan:
Needs #896

It needs tests and index writer is still not complete:

* 64 bit offsets
* Index checksum

Some improvements can be done changing the cache to hold byte slices and index by offset. We can also skip reading headers in `resolveDeltas` phase.

### DIFF
--- a/plumbing/format/idxfile/writer.go
+++ b/plumbing/format/idxfile/writer.go
@@ -1,0 +1,132 @@
+package idxfile
+
+import (
+	"bytes"
+	"math"
+	"sort"
+
+	"gopkg.in/src-d/go-git.v4/plumbing"
+	"gopkg.in/src-d/go-git.v4/utils/binary"
+)
+
+type object struct {
+	hash   plumbing.Hash
+	offset int64
+	crc    uint32
+}
+
+type objects []object
+
+// Writer implements a packfile Observer interface and is used to generate
+// indexes.
+type Writer struct {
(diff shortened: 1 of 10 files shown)


plumbing/format/idxfile/writer.go
```diff
@@ -0,0 +1,132 @@
...
+import (
+	"bytes"
+	"math"
+	"sort"
+
+	"gopkg.in/src-d/go-git.v4/plumbing"
+	"gopkg.in/src-d/go-git.v4/utils/binary"
+)
+
+type object struct {
```
erizocosmico: we already have `Entry` in this package with this exact same struct, maybe you can use that?

plumbing/format/idxfile/writer.go
```diff
@@ -0,0 +1,132 @@
...
+	hash   plumbing.Hash
+	offset int64
+	crc    uint32
+}
+
+type objects []object
```
erizocosmico: are we using the `objects` type for something specific (i don't see any methods for it)? If not, I'd say just go with `[]object` (or `[]Entry`)

erizocosmico: Nvm, saw the methods at the end of the file

jfontan:
Now it supports 64 bit offsets but needs more testing. Index checksum is already generated by the encoder.

PR #899, opened 2018-07-20
 plumbing: packfile, new Packfile representation 

### LINKED ISSUE #897: Packfile representation
Packfile is intended to be used when something needs information from the packfile, and it needs to retrieve that information as fast as possible.

To make it possible we need to use in a good way the several files generated from the packfile parser.

We will have a Packfile object that uses an Index instance mandatorily to be able to get random objects in the fastest way.

Methods for that object will be:

```go
type Packfile struct {
    Index Index
    file os.File
}

func (*Packfile) Get(h plumbing.ObjectHash) (plumbing.EncodedObject, error) {}

func (*Packfile) GetByOffset(o int64) (plumbing.EncodedObject, error) {}

func (*Packfile) GetAll() (EncodedObjectIter, error) {}

func (*Packfile) ID() (plumbing.Hash, error) {}

func (*Packfile) Close() error {}
```

Description from: https://docs.google.com/document/d/1I2Vte4LhMdpx0kvlrg-k6hNA-0kUGtUkC1-kfwbCmeI

erizocosmico:
Closes #897 

Depends on #898 and #896 

Introduces the new `Packfile` representation that can retrieve `plumbing.EncodedObject`s from the packfile file given an index.

The status of this is very WIP, just as #898 and #896. A lot of pieces still don't fit very well together and we have to figure out what to deprecate and remove and how to put it all together better.

The packfile representation, specifically, is working and passing tests. It's duplicating a lot of logic from the packfile decoder right now, but until we know better if we should remove it or make it work with this I preferred to just duplicate it. When we're more sure about which direction we'll be moving in, then we can think about either removing the decoder or making some sort of component that contains this shared logic between the two components.


plumbing/format/packfile/decoder.go
```diff
@@ -120,6 +122,7 @@ func NewDecoderForType(s *Scanner, o storer.EncodedObjectStorer,
 
 		idx:          idxfile.NewMemoryIndex(),
 		offsetToType: make(map[int64]plumbing.ObjectType),
+		offsetToHash: make(map[int64]plumbing.Hash),
```
jfontan: We can use `idx.FindHash` instead of having this duplicated here.

plumbing/format/packfile/packfile.go
```diff
@@ -0,0 +1,249 @@
...
+	"gopkg.in/src-d/go-git.v4/plumbing/storer"
+)
+
+// Packfile allows retrieving information from inside a packfile.
+type Packfile struct {
+	idxfile.Index
+	billy.File
+	s              *Scanner
+	deltaBaseCache cache.Object
+	offsetToHash   map[int64]plumbing.Hash
```
jfontan: Same as with decoder, this can live in the index.

erizocosmico:
@jfontan but the index should already contain all this mapping, right?

erizocosmico:
Fixed, now it works with `DiskObject` objects and is lazy (as much as possible)

PR #904, opened 2018-07-27
Tests and indexes in packfile decoder

jfontan:
* Add tests to 64 bit offset index creation and `FindHash`
* Add index generation to `packfile.Decoder`
* Remove duplicate `filesystem.IndexStorage`


erizocosmico:
Should tests be failing?

jfontan:
They also fail in my computer so yes, there are still tests failing. I still haven't look why they fail.

PR #908, opened 2018-08-05
plumbing: Change final return to a default in the switch

SMTullis:
The final return in ReferenceType.String() was implemented to return an 
empty string only when the switch did not have a match. Properly, this 
should be a default within the switch statement.


smola:
@SMTullis Could you sign-off the commit (`git commit -s` or `git commit --amend -s`), see https://github.com/src-d/go-git/blob/master/CONTRIBUTING.md#certificate-of-origin

mcuadros:
Please re-open when the DCO once signed, thanks

PR #916, opened 2018-08-14
Improvement/memory consumption new packfile parser

jfontan:
* plumbing: add buffer cache and use it in packfile parser: It uses less memory and is faster as slices don't have to be converted from/to `MemoryObject` and they are indexed by offset.

* plumbing/pacfile: tidy up objectInfo struct
  * a new hasher is created when needed
  * delete unused fields
  * base content is no longer kept in memory

Testing clone of a local repo with a 1.1 Gb packfile and 3 million objects:

Before:

```
295.85user 19.33system 3:57.77elapsed 132%CPU (0avgtext+0avgdata 7045108maxresident)k
7312inputs+3482616outputs (0major+1836239minor)pagefaults 0swaps
```

Now:

```
397.70user 19.12system 4:48.93elapsed 144%CPU (0avgtext+0avgdata 2460856maxresident)k
0inputs+3482616outputs (0major+650858minor)pagefaults 0swaps
```

### DIFF
--- a/plumbing/cache/buffer_lru.go
+++ b/plumbing/cache/buffer_lru.go
@@ -1,0 +1,98 @@
+package cache
+
+import (
+	"container/list"
+	"sync"
+)
+
+// BufferLRU implements an object cache with an LRU eviction policy and a
+// maximum size (measured in object size).
+type BufferLRU struct {
+	MaxSize FileSize
+
+	actualSize FileSize
+	ll         *list.List
+	cache      map[int64]*list.Element
+	mut        sync.Mutex
+}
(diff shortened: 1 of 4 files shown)


jfontan:
Improved a bit the performance skipping deltas that already had the hash computed. This happens when the delta is used again as base but is not in cache.

Now:

```
346.68user 18.48system 4:05.61elapsed 148%CPU (0avgtext+0avgdata 2664700maxresident)k
0inputs+3486544outputs (0major+913970minor)pagefaults 0swaps
```

PR #920, opened 2018-08-17
git: Add ability to PGP sign commits

vancluever:
This adds the ability to sign commits by adding the `SignKey` field to
`CommitOptions`. If present, the commit will be signed during the
`WorkTree.Commit` call.

The supplied `SignKey` must already be decrypted by the caller.

`Signed-off-by: Chris Marchesi <chrism@vancluevertech.com>`

-----

There is also a small bugfix in here as well that was affecting the ability to sign commits:

```
plumbing: object, Don't add new line at end of commit signature

The way that commit signatures were being written out was causing an
extra newline to be written at the end of the commit when the message
encoding was already taking care of this. Ultimately, this results in a
corrupt object, rendering the object unverifiable with the signature in
the commit.

Signed-off-by: Chris Marchesi <chrism@vancluevertech.com>
```

### DIFF
--- a/worktree_commit.go
+++ b/worktree_commit.go
@@ -1,6 +1,7 @@
 package git
 
 import (
+	"bytes"
 	"path"
 	"strings"
 
@@ -117,7 +118,7 @@ func (w *Worktree) buildCommitSignature(commit *object.Commit, signKey *openpgp.
 	if err != nil {
 		return "", err
 	}
-	var b strings.Builder
+	var b bytes.Buffer
 	if err := openpgp.ArmoredDetachSign(&b, signKey, r, nil); err != nil {
 		return "", err
 	}
--- a/worktree_commit.go
+++ b/worktree_commit.go
@@ -4,6 +4,7 @@ import (
 	"path"
 	"strings"
 
+	"golang.org/x/crypto/openpgp"
 	"gopkg.in/src-d/go-git.v4/plumbing"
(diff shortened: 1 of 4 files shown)


vancluever:
Looks like tests are failing due to use of `strings.Builder`.

I'll switch this up to using `bytes.Buffer` to maintain the backwards compatibility with Go 1.9.

worktree_commit.go
```diff
@@ -92,13 +94,37 @@ func (w *Worktree) buildCommitObject(msg string, opts *CommitOptions, tree plumb
...
 		return plumbing.ZeroHash, err
 	}
 	return w.r.Storer.SetEncodedObject(obj)
 }
 
+func (w *Worktree) buildCommitSignature(commit *object.Commit, signKey *openpgp.Entity) (string, error) {
+	encoded := &plumbing.MemoryObject{}
+	if err := commit.Encode(encoded); err != nil {
+		return "", err
+	}
```
smola: Add spaces after each if block (this is a [convention that we generally follow all over go-git codebase](https://github.com/src-d/guide/blob/master/engineering/conventions-go.md#group-code-blocks-with-blank-lines)).

smola:
I triggered manually the failing job in CI. It is an intermittent failure, not related to this PR.

vancluever:
Thanks @smola! I'll keep that in mind for some future tagging work I'm working on.

PR #928, opened 2018-08-22
git: Add tagging support

vancluever:
This adds a few new higher-level functions:

* `CreateTag`, which can be used to create both lightweight and annotated
tags with a supplied `TagObjectOptions` struct. PGP signing is possible as
well.
* `Tag`, to fetch a single tag ref.
* `DeleteTag`, to delete a tag. This simply deletes the ref. The object is
left orphaned to be GCed later (maybe, see below).

-----

I'm not 100% sure if `DeleteTag` is the correct behavior - looking for
details on exactly *what* happens to a tag object if you delete the ref
and not the tag were sparse, and groking the Git source did not really
produce much insight to the untrained eye. This may be something that
comes up in review. If deletion of the object is necessary, the
in-memory storer may require some updates to allow `DeleteLooseObject` to
be supported.

-----

Also, similar to how commit encoding needed to be corrected to avoid corruption of the commit for verification, tag objects needed the same thing. That has been added in another commit.


vancluever:
Looks like some encoding-related tests may be failing. I'll check these out when I have a chance.

It should be noted that `Decode` and `Verify` have not been touched, so more than likely the expected data in the failing tests may just need to be adjusted.

vancluever:
So the issue was not necessarily with the encoding/decoding.

Ultimately, the encoder/decoder expects the message to be in a canonicalized format where there is _always_ a newline, even for short messages (no body). Assuming this, the (new) encoder was ultimately adding on extra newlines when it should not have been.

To correct this I've started to do canonicalization in the higher level - `TagObjectOptions.Message` is now  run through `strings.TrimSpace` with a newline added at the end, to make sure that any incoming message is corrected, regardless of what shape it's in.

vancluever:
Made a small update today so you don't need to actually pass in the object type for an annotated tag. This will now be fetch from the passed in hash.

vancluever:
I'm kind of finding how much fetching a tag object is not necessarily needed. As such, I may adjust `Tag()` so that it does not return the tag object as well, leaving that to be fetched/followed by the caller.

Contrary to the docs too, all tags have references, so the current `Tags()` iterator will actually work for fetching annotated tags too. The only difference is that the caller has to pull the object data manually by following the hash in the ref.

I've seen a lot of this in preliminary use though, which is why I'm thinking it's not really necessary to return both:

```go
ref, _, err := r.Tag(...)
_, obj, err := r.Tag(...)
```

repository.go
```diff
@@ -484,6 +492,153 @@ func (r *Repository) DeleteBranch(name string) error {
...
+}
+
+// DeleteTag deletes a tag from the repository.
+func (r *Repository) DeleteTag(name string) error {
+	ref, err := r.Tag(name)
+	if err != nil {
+		return err
+	}
+
+	obj, err := r.TagObject(ref.Hash())
```
smola: Git does not delete the object even if it's a loose object. It just defers to prune as you pointed out. I think we should do as git, not just for compatibility, but also to avoid divergences with different storage implementations.

It would be nice to have a test for the `Prune` function that checks that if you create an annotated tag and delete the lightweight one, the annotated tag is effectively deleted.

repository.go
```diff
@@ -845,9 +1000,31 @@ func (r *Repository) Log(o *LogOptions) (object.CommitIter, error) {
 	return nil, fmt.Errorf("invalid Order=%v", o.Order)
 }
 
-// Tags returns all the References from Tags. This method returns only lightweight
-// tags. Note that not all the tags are lightweight ones. To return annotated tags
-// too, you need to call TagObjects() method.
+// Tags returns all the tag References in a repository.
```
smola: Thanks for the correction, this was, indeed, confusing at least!

repository.go
```diff
@@ -484,6 +492,153 @@ func (r *Repository) DeleteBranch(name string) error {
...
+	default:
+		// Some other error
+		return nil, err
+	}
+
+	var target plumbing.Hash
+	if opts != nil {
+		target, err = r.createTagObject(name, hash, opts)
+		if err != nil {
+			return nil, err
```
smola: I verified with plain git that if the tag object creation succeeds and the lightweight creation fails, it does not try to clean up and just leaves a dangling tag object behind. So we're good here (I had doubts initially).

repository.go
vancluever: I actually thought I removed this, because as mentioned in the original PR message it's actually not even possible to test this with an annotated tag right now due to limitations of the in-memory storage interface.

I'll get it removed and add a test to delete an annotated tag with a `Prune` on the end.

repository.go
```diff
@@ -484,6 +492,153 @@ func (r *Repository) DeleteBranch(name string) error {
...
+
+	var b bytes.Buffer
+	if err := openpgp.ArmoredDetachSign(&b, signKey, rdr, nil); err != nil {
+		return "", err
+	}
+
+	return b.String(), nil
+}
+
+// Tag fetches a tag from the repository.
```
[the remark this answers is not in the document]
vancluever: Done!

vancluever:
Hey @smola! All the stuff has been added now.

`Prune` was actually not working for tags as tag objects were not supported for walking, incidentally enough. This has been added now. One of the annotated tag deletion tests exclusively tests unpacked objects too, so we know that `Prune` is working 100% correctly.

options.go
```diff
@@ -377,6 +379,41 @@ func (o *CommitOptions) Validate(r *Repository) error {
...
 }
 
+var (
+	ErrMissingName    = errors.New("name field is required")
+	ErrMissingTagger  = errors.New("tagger field is required")
+	ErrMissingMessage = errors.New("message field is required")
+)
+
+// TagObjectOptions describes how a tag object should be created.
+type TagObjectOptions struct {
```
smola: `s/TagObjectOptions/CreateTagOptions/` to match current convention of `<Function>Options`

vancluever:
@smola I mentioned it in Slack but just wanted to note (versus just resolving the conversation) that I made the naming updates as well. Thanks!

smola:
cc @mcuadros

mcuadros:
@vancluever thanks for the contribution.

vancluever:
Thanks @mcuadros!

PR #932, opened 2018-08-24
Fix `fatal: corrupt patch` error in unified diff format

distorhead:
There may be negative number in the range of unified diff. Negative number occurs when patch changes for example first or second line of file. Bug related to ContextLines number.

Demonstration repo and program to generate wrong patch available here:

https://github.com/flant/go-git-test/

Resulting patch:

```
diff --git a/README.md b/README.md
index 94954abda49de8615a048f8d2e64b5de848e27a1..f3dad9514629b9ff9136283ae331ad1fc95748a8 100644
--- a/README.md
+++ b/README.md
@@ -1,2 +-1,2 @@
 hello
-world
+bug
```

`+-1,2` causes `git apply` to complain:

```
fatal: corrupt patch at line 5
```


mcuadros:
@ajnavarro PTAL

ajnavarro:
@mcuadros the change looks good to me, but I will appreciate a test for it.

PR #935, opened 2018-08-25
storage/dotgit: search for incoming dir only once

jfontan:
Search for incoming object directory was done once each time objects were accessed. This means a ReadDir of the objects path that is expensive. Now incoming directory is searched the first time an object is accessed and its name kept in DotGit to be reused.

This has the drawback that if the incoming object directory is created after the repo is opened it won't be found.

Related to #887

### DIFF
--- a/storage/filesystem/dotgit/dotgit.go
+++ b/storage/filesystem/dotgit/dotgit.go
@@ -283,11 +283,14 @@ func (d *DotGit) objectPath(h plumbing.Hash) string {
 	return d.fs.Join(objectsPath, hash[0:2], hash[2:40])
 }
 
-//incomingObjectPath is intended to add support for a git pre-recieve hook to be written
-//it adds support for go-git to find objects in an "incoming" directory, so that the library
-//can be used to write a pre-recieve hook that deals with the incoming objects.
-//More on git hooks found here : https://git-scm.com/docs/githooks
-//More on 'quarantine'/incoming directory here : https://git-scm.com/docs/git-receive-pack
+// incomingObjectPath is intended to add support for a git pre-receive hook
+// to be written it adds support for go-git to find objects in an "incoming"
+// directory, so that the library can be used to write a pre-receive hook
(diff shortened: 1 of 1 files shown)


jfontan:
@noxora, can you check if this change is compatible with pre-commit hook?

noxora:
I do not see why this would not be compatible, it appears to be a solid change.

smola:
I think new incoming directories can appear in subsequent calls?

jfontan:
@smola, yes, that can happen if this is the use case.

* Open repository with go-git
* Execute `git pull` that triggers pre hook
* While the pre hook is running read from repository with go-git

While this is possible I don't think it's the common usage. Most probably go-git will be used inside the pre hook.

* Execute `git clone` that triggers pre hook
* pre hook executes a binary that uses go-git to read the repo

storage/filesystem/dotgit/dotgit.go
```diff
@@ -286,26 +290,37 @@ func (d *DotGit) objectPath(h plumbing.Hash) string {
...
+}
+
+// hasIncomingObjects searches for an incoming directory and keeps its name
+// so it doesn't have to be found each time an object is accessed.
+func (d *DotGit) hasIncomingObjects() bool {
+	if !d.incomingChecked {
+		directoryContents, err := d.fs.ReadDir(objectsPath)
+		if err == nil {
+			for _, file := range directoryContents {
+				if strings.Split(file.Name(), "-")[0] == "incoming" && file.IsDir() {
```
kuba--: Is it the same as `strings.HasPrefix(file.Name(), "incoming")`

jfontan: Changed

noxora:
@jfontan I thought similarly, it is possible that it could change, but if it is being used to write a hook, it shouldn't. However, if we were to treat this as volatile (which it could be, in some uncommon use cases) then this would be problematic.

Issue #877, opened 2018-06-28
newknownhostscallback always uses default known_hosts files


thehowl:
https://github.com/src-d/go-git/blob/fcfd239c49baee9195f7b8234f46288a29c3951c/plumbing/transport/ssh/auth_method.go#L246-L258

`files` get overwritten locally with the result of `getDefaultKnownHosts`. There is no check `if len(files) == 0` at the top, so no matter what arguments are passed to the function, the defaults are always used.

PR #940, opened 2018-08-30
Check if list of files is empty.

kuba--:

Closes https://github.com/src-d/go-git/issues/877


mcuadros:
Can we add a test for this?

kuba--:
I'll try to generate some fake `know_hosts`  files. We'll see how it goes

kuba--:
@mcuadros - I've added `TestNewKnownHostsCallback` which mocks a `known_hosts` file with `github` host.

kuba--:
rebased

PR #941, opened 2018-08-30
plumbing/storer: add ExclusiveAccess option to Storer

jfontan:
Adds `ExclusiveAccess` configuration to `filesystem.Storage` and `filesystem.DotGit`. This option means that the git repository is not expected to be modified while open and enables some optimizations.

Each time a file is accessed the storer tries to open an object file for the requested hash. When this is done for a lot of objects it is expensive. With `ExclusiveAccess` option a list of object files is generated the first time an object is accessed and used to check if exists instead of using system calls.

A similar optimization is done for packfiles.

### DIFF
--- a/options.go
+++ b/options.go
@@ -9,6 +9,7 @@ import (
 	"gopkg.in/src-d/go-git.v4/plumbing"
 	"gopkg.in/src-d/go-git.v4/plumbing/object"
 	"gopkg.in/src-d/go-git.v4/plumbing/protocol/packp/sideband"
+	"gopkg.in/src-d/go-git.v4/plumbing/storer"
 	"gopkg.in/src-d/go-git.v4/plumbing/transport"
 )
 
@@ -428,11 +429,12 @@ func (o *GrepOptions) Validate(w *Worktree) error {
 // PlainOpenOptions describes how opening a plain repository should be
 // performed.
 type PlainOpenOptions struct {
+	// Storage layer options.
+	Storage storer.Options
+
 	// DetectDotGit defines whether parent directories should be
 	// walked until a .git directory or file is found.
 	DetectDotGit bool
-	// Static means that the repository won't be modified while open.
-	Static bool
 }
 
 // Validate validates the fields and sets the default values.
--- a/options.go
+++ b/options.go
@@ -9,7 +9,6 @@ import (
 	"gopkg.in/src-d/go-git.v4/plumbing"
 	"gopkg.in/src-d/go-git.v4/plumbing/object"
 	"gopkg.in/src-d/go-git.v4/plumbing/protocol/packp/sideband"
-	"gopkg.in/src-d/go-git.v4/plumbing/storer"
 	"gopkg.in/src-d/go-git.v4/plumbing/transport"
 )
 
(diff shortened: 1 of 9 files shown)


storage/filesystem/dotgit/dotgit.go
```diff
@@ -241,9 +304,29 @@ func (d *DotGit) Objects() ([]plumbing.Hash, error) {
 	return objects, nil
 }
 
-// Objects returns a slice with the hashes of objects found under the
-// .git/objects/ directory.
+// ForEachObjectHash iterates over the hashes of objects found under the
+// .git/objects/ directory and executes the provided .
```
ajnavarro: the provided function?

jfontan:
I'm also going to move `StorageOptions` to `options.go` as do not make sense repeating the same structure in `repository`, `filesystem.Storage` and `DotGit`.

options.go
```diff
@@ -431,6 +431,8 @@ type PlainOpenOptions struct {
 	// DetectDotGit defines whether parent directories should be
 	// walked until a .git directory or file is found.
 	DetectDotGit bool
+	// Static means that the repository won't be modified while open.
+	Static bool
```
kuba--: so maybe readonly

kuba--: readonly?

storage/filesystem/dotgit/dotgit.go
```diff
@@ -241,9 +304,29 @@ func (d *DotGit) Objects() ([]plumbing.Hash, error) {
...
 func (d *DotGit) ForEachObjectHash(fun func(plumbing.Hash) error) error {
```
kuba--: yeah, it should be `fun` ;)

options.go
jfontan: I thought about read only but what it really means is that it is not modified from the outside. I've added packfile and hash cache invalidation when new objects or packfiles are created. Mainly because I wanted to pass all tests in this mode.

Anyway, I agree that `Static` may not be the best name but could not find any other that reflected that is not modified from the outside.

storage/filesystem/dotgit/dotgit.go
jfontan: It is fun indeed!

storage/filesystem/dotgit/dotgit.go
jfontan: done

kuba--:
Maybe 👍 for this PR speed up merging ;) So we can start extending Storage's Options by adding configurable cache size (https://github.com/src-d/gitbase/issues/440)

smola:
Just a doubt about naming and behaviour: Does this mean that it will not be modified externally or that it won't be modified even by go-git itself while open?

If it's the later, maybe it should be named `Immutable`?

jfontan:
@smola The first case. We don't expect it to be modified externally. We can write with go-git (create new packfiles or objects).

Maybe `Exclusive` or `ExclusiveAccess` is better?

smola:
@jfontan `ExclusiveAccess` looks explicit enough.

storage/filesystem/storage.go
```diff
@@ -11,6 +12,8 @@ import (
 // standard git format (this is, the .git directory). Zero values of this type
 // are not safe to use, see the NewStorage function below.
 type Storage struct {
+	storer.Options
```
mcuadros: Why is an embedded struct? and not a private field?

jfontan: You're right. There's no reason why it should be embedded or public. It's now changed.

PR #942, opened 2018-08-30
storage/dotgit: add KeepDescriptors option

jfontan:
Needs #941 

This option maintains packfile file descriptors opened after reading objects from them. It improves performance as it does not have to be opening packfiles each time an object is needed.

Also adds `Close` to `EncodedObjectStorer` to close all the files manually.

### DIFF
--- a/plumbing/storer/object.go
+++ b/plumbing/storer/object.go
@@ -40,8 +40,6 @@ type EncodedObjectStorer interface {
 	// HasEncodedObject returns ErrObjNotFound if the object doesn't
 	// exist.  If the object does exist, it returns nil.
 	HasEncodedObject(plumbing.Hash) error
-	// Close any opened files.
-	Close() error
 }
 
 // DeltaObjectStorer is an EncodedObjectStorer that can return delta
--- a/plumbing/storer/object.go
+++ b/plumbing/storer/object.go
@@ -40,6 +40,8 @@ type EncodedObjectStorer interface {
 	// HasEncodedObject returns ErrObjNotFound if the object doesn't
 	// exist.  If the object does exist, it returns nil.
 	HasEncodedObject(plumbing.Hash) error
+	// Close any opened files.
+	Close() error
 }
 
 // DeltaObjectStorer is an EncodedObjectStorer that can return delta
--- a/plumbing/storer/object_test.go
+++ b/plumbing/storer/object_test.go
@@ -157,3 +157,7 @@ func (o *MockObjectStorage) IterEncodedObjects(t plumbing.ObjectType) (EncodedOb
 func (o *MockObjectStorage) Begin() Transaction {
 	return nil
 }
+
+func (o *MockObjectStorage) Close() error {
+	return nil
+}
--- a/plumbing/storer/object_test.go
+++ b/plumbing/storer/object_test.go
@@ -157,7 +157,3 @@ func (o *MockObjectStorage) IterEncodedObjects(t plumbing.ObjectType) (EncodedOb
 func (o *MockObjectStorage) Begin() Transaction {
 	return nil
 }
-
-func (o *MockObjectStorage) Close() error {
-	return nil
-}
--- a/storage/filesystem/dotgit/dotgit.go
+++ b/storage/filesystem/dotgit/dotgit.go
@@ -62,6 +62,9 @@ type Options struct {
 	// ExclusiveAccess means that the filesystem is not modified externally
 	// while the repo is open.
 	ExclusiveAccess bool
+	// KeepDescriptors makes the file descriptors to be reused but they will
+	// need to be manually closed calling Close().
+	KeepDescriptors bool
 }
 
 // The DotGit type represents a local git repository on disk. This
@@ -78,6 +81,8 @@ type DotGit struct {
 	objectMap  map[plumbing.Hash]struct{}
 	packList   []plumbing.Hash
 	packMap    map[plumbing.Hash]struct{}
+
+	files map[string]billy.File
 }
 
 // New returns a DotGit value ready to be used. The path argument must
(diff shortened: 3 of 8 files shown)


options.go
```diff
@@ -426,8 +427,12 @@ func (o *GrepOptions) Validate(w *Worktree) error {
 }
 
 // PlainOpenOptions describes how opening a plain repository should be
+
 // performed.
 type PlainOpenOptions struct {
+	// Storage layer options.
```
smola: Maybe we should keep this out of `Plain*` methods and avoid adding `Close` to general interface. Just adding it to the `filesystem` implementation would be enough, and this option would be available only by creating the `filesystem` storage directly.

This way we avoid redundancy in other implementatons or end up exposing a lot of storage-specific options here.

jfontan: Done

plumbing/storer/storer.go
```diff
@@ -13,3 +13,13 @@ type Initializer interface {
 	// any.
 	Init() error
 }
+
+// Options holds configuration for the storage.
+type Options struct {
+	// ExclusiveAccess means that the filesystem is not modified externally
```
mcuadros: Since this is options strictly related to a one implementation I don't believe this is the right place to put this options, this options should be pass and used only to the file system implementation

storage/filesystem/dotgit/dotgit.go
```diff
@@ -60,18 +61,35 @@ var (
 // The DotGit type represents a local git repository on disk. This
 // type is not zero-value-safe, use the New function to initialize it.
 type DotGit struct {
-	fs billy.Filesystem
+	Options storer.Options
```
mcuadros: This should be private, if since can't be changed

storage/filesystem/object.go
```diff
@@ -268,7 +269,9 @@ func (s *ObjectStorage) getFromPackfile(h plumbing.Hash, canBeDelta bool) (
 		return nil, err
 	}
 
-	defer ioutil.CheckClose(f, &err)
+	if !s.dir.Options.KeepDescriptors {
```
mcuadros: Instead of requesting the options to dogGit, should be accessed from a local pointer to the storer.

storage/filesystem/object.go
```diff
@@ -411,6 +414,11 @@ func (s *ObjectStorage) buildPackfileIters(t plumbing.ObjectType, seen map[plumb
 	}, nil
 }
 
+// Close closes all opened files.
+func (s *ObjectStorage) Close() error {
```
mcuadros: This is not covered by a test

storage/filesystem/dotgit/dotgit.go
```diff
@@ -278,6 +394,87 @@ func (d *DotGit) ForEachObjectHash(fun func(plumbing.Hash) error) error {
 	return nil
 }
 
+func (d *DotGit) cleanObjectList() {
+	d.objectMap = nil
+	d.objectList = nil
+}
+
+func (d *DotGit) genObjectList() error {
```
mcuadros: This function is not covered by a test

storage/filesystem/dotgit/dotgit.go
```diff
@@ -241,9 +337,29 @@ func (d *DotGit) Objects() ([]plumbing.Hash, error) {
...
-// Objects returns a slice with the hashes of objects found under the
-// .git/objects/ directory.
+// ForEachObjectHash iterates over the hashes of objects found under the
+// .git/objects/ directory and executes the provided function.
 func (d *DotGit) ForEachObjectHash(fun func(plumbing.Hash) error) error {
+	if !d.Options.ExclusiveAccess {
+		return d.forEachObjectHash(fun)
+	}
+
+	err := d.genObjectList()
```
mcuadros: This code isn't covered either.

storage/filesystem/dotgit/dotgit.go
jfontan: By default go coverage only annotates files from the same package tested. The test that exercise this part belongs to `storage/filesystem`. That's why it does not appear in coverage. The same happens with other functions. I'll add new tests in `storage/filesystem/dotgit` that forces the usage of those functions.

plumbing/storer/object.go
```diff
@@ -40,6 +40,8 @@ type EncodedObjectStorer interface {
 	// HasEncodedObject returns ErrObjNotFound if the object doesn't
 	// exist.  If the object does exist, it returns nil.
 	HasEncodedObject(plumbing.Hash) error
+	// Close any opened files.
+	Close() error
```
smola: What about having `Close()` only on the filesystem implementation, not on the general interface?

jfontan: Somehow I understood that you didn't want to expore `Close` in `git.Repository`. It's not deleted from `EncodedObjectStorer`.

plumbing/storer/object_test.go
```diff
@@ -157,3 +157,7 @@ func (o *MockObjectStorage) IterEncodedObjects(t plumbing.ObjectType) (EncodedOb
 func (o *MockObjectStorage) Begin() Transaction {
 	return nil
 }
+
+func (o *MockObjectStorage) Close() error {
```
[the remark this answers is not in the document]
jfontan: You're right. Deleted that and `Close` in memory storer. Also added a new test for filesystem storer and `KeepDescriptors` option.

smola:
@jfontan AppVeyor is failing with OOM when executing `storage/filesystem` tests. It looks related to the new test, since other builds are ok and this one failed multiple times.

jfontan:
It is indeed the test. The comparison is generating diffs of the packfiles instead of checking if they are the same. Working on a fix.

jfontan:
@smola it seems the bug in windows tests is fixed.

PR #947, opened 2018-09-05
Custom cache

kuba--:
If we want to let apps. adjust cache size (or pass own implementation) we need to expose it somehow.
Recently we added Options to ObjectStorage, so instead of extending ObjectStorage structure and implement new constructors WithCache, I've moved cache under Options structure.

It's related to src-d/gitbase#440

### DIFF
--- a/storage/filesystem/object.go
+++ b/storage/filesystem/object.go
@@ -18,21 +18,29 @@ import (
 )
 
 type ObjectStorage struct {
-	// deltaBaseCache is an object cache uses to cache delta's bases when
-	deltaBaseCache cache.Object
+	options Options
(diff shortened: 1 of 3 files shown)


mcuadros:
The options struct is a container of flags, is not designed to container pointer or state.

mcuadros:
@kuba-- if you need to configure the cache, just send put in the constructor as another arg.

ajnavarro:
@mcuadros Is not just configure the cache, we want to have control about the instance itself, to be able to share it between several repositories or change the implementation to fit our use case, per example.

Should we send it as another constructor param?

PR #949, opened 2018-09-07
Expose Storage cache.

kuba--:

If we want to let apps. adjust cache size (or pass own implementation) we need to expose it somehow.
It's related to src-d/gitbase#440
Also we don't need to return an error from constructors like `NewStorage`.

### DIFF
--- a/storage/filesystem/object.go
+++ b/storage/filesystem/object.go
@@ -27,24 +27,18 @@ type ObjectStorage struct {
 	index map[plumbing.Hash]idxfile.Index
 }
 
-// NewObjectStorage creates a new ObjectStorage with the given .git directory.
-func NewObjectStorage(dir *dotgit.DotGit) (ObjectStorage, error) {
-	return NewObjectStorageWithOptions(dir, Options{})
+// NewObjectStorage creates a new ObjectStorage with the given .git directory and cache.
+func NewObjectStorage(dir *dotgit.DotGit, cache cache.Object) *ObjectStorage {
+	return NewObjectStorageWithOptions(dir, cache, Options{})
 }
 
-// NewObjectStorageWithOptions creates a new ObjectStorage with the given .git
-// directory and sets its options.
-func NewObjectStorageWithOptions(
-	dir *dotgit.DotGit,
-	ops Options,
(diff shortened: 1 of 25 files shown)


storage/filesystem/object.go
```diff
@@ -28,23 +28,22 @@ type ObjectStorage struct {
...
+}
+
+func (s *ObjectStorage) WithOptions(ops Options) *ObjectStorage {
+	s.options = ops
+	return s
+}
 
-	return s, nil
+func (s *ObjectStorage) WithCache(cache cache.Object) *ObjectStorage {
+	s.deltaBaseCache = cache
```
mcuadros: The cache should be passed to the constructor in the same way we do with the dotgit instance since is required, the options are optional, but if in the case of the cache the LRU is being instanced always in the constructor wasting resources.


`func NewObjectStorage(dir *dotgit.DotGit, cache cache.Object) *ObjectStorage {`

kuba--:
@mcuadros - now cache is required param in constructors.

storage/filesystem/object.go
```diff
@@ -182,10 +176,7 @@ func (s *ObjectStorage) EncodedObject(t plumbing.ObjectType, h plumbing.Hash) (p
 			// Create a new object storage with the DotGit(s) and check for the
 			// required hash object. Skip when not found.
 			for _, dg := range dotgits {
-				o, oe := NewObjectStorage(dg)
-				if oe != nil {
-					continue
-				}
+				o := NewObjectStorage(dg, cache.NewObjectLRUDefault())
```
ajnavarro: should we use here `s.deltaBaseCache` ?

kuba--: ohh, good question - I think so.

maguro:
Oy gevalt!  A backward incompatible change passed off in a minor release.  Thanks!

vmarkovtsev:
@maguro This is how you learn to use vendoring and `dep`.

johanbrandhorst:
Not sure why the dismissive tone, this is indeed a breaking change, and it should not have been merged. I've raised https://github.com/src-d/go-git/issues/1262. I think the signature change should be reverted.

PR #962, opened 2018-09-20
storage/filesystem: keep packs open in PackfileIter

jfontan:
PackfileIter was not taking into account the option KeepDescriptors and was always closing the file. This caused "file already closed" errors when iterating packfiles in with KeepDescriptors active.

### DIFF
--- a/storage/filesystem/object.go
+++ b/storage/filesystem/object.go
@@ -396,7 +396,10 @@ func (s *ObjectStorage) IterEncodedObjects(t plumbing.ObjectType) (storer.Encode
 	return storer.NewMultiEncodedObjectIter(iters), nil
 }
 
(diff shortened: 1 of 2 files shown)


storage/filesystem/object.go
```diff
@@ -470,9 +476,10 @@ func (it *lazyPackfilesIter) Close() {
...
-	pack billy.File
-	iter storer.EncodedObjectIter
-	seen map[plumbing.Hash]struct{}
+	pack     billy.File
+	iter     storer.EncodedObjectIter
+	seen     map[plumbing.Hash]struct{}
+	keepPack bool
 }
 
 // NewPackfileIter returns a new EncodedObjectIter for the provided packfile
```
mcuadros: Please add some documentation about `keepPack`

jfontan: Done

Issue #681, opened 2017-12-06
unquoted '\' must be followed by new line


fredcarle:
Has anyone seen this before? I get this `unquoted '\' must be followed by new line` error on some of the repo I try to open with `git.PlainOpen()`.

orirawlings:
Hey @fredcarle,

Can you share any code or links to the repository you are trying to open so we can try and reproduce the issue?

fredcarle:
Unfortunately I can share the link to the repo since it's a private one. But here is the simple code I was using for testing.

```
        r, err := git.PlainOpen("/link/to/my/local/repo")
	if err != nil {
		log.Panic(err)
	}
	remotes, err := r.Remotes()
	if err != nil {
		log.Panic(err)
	}
	for _, r := range remotes {
		fmt.Println("remote", r.String())
	}
```

strib:
I've seen this before.  In one instance, it was caused by trying to read a config written by Git tower: keybase/client#8812

The parser used to read the git config files really doesn't like backslashes.

fredcarle:
That would makes sense since I'm using Tower

orirawlings:
Seems like there is some benefit in extending the git config parser to handle backslashes more gracefully.

japanoise:
This is especially problematic on Windows, because the likelihood of having slashes is much higher (Paths, for example)

scbizu:
Resolved this by removing all cols contain `\` in `.git/config`, which seems like generated by Git Tower.

filipnavara:
It bails out on the following `.git/config`:
```
[core]
	repositoryformatversion = 0
	filemode = false
	bare = true
	symlinks = false
	ignorecase = true
[remote "origin"]
	url = E:\\Git\\emclient\\emclient.git
```

This is perfectly valid and parsable by the original Git parser (https://github.com/git/git/blob/master/config.c#L537-L560).

PR #977, opened 2018-10-05
config: Add test for Windows local paths.

filipnavara:
- Make sure that go-git is able to read unquoted local Windows path with escaped characters (as written by cgit).
- Add test for writing it and round-tripping it, where both cases use quoted string, which is still accepted by cgit.

Issue #681


filipnavara:
I don't think the code coverage metric is calculated correctly if it decreased with added test coverage, or am I missing something?

mcuadros:
@filipnavara don't worry, this small changes in coverages are irrelevant.

PR #979, opened 2018-10-08
plumbing: object, Add support for Log with filenames. Fixes #826

### LINKED ISSUE #826: git log single file
Current version only support query the history of entire project (git log), can support a single file (git log filename)?

gnithin:
Hi,
This is in reference to #826 
This commit basically performs what `git log -- <filename>` should do.

I've added a new class called `commit_walker_file`, which wraps around the iterator of the original `Log` function and returns those commits whose `Difftree` contains the file-path specified. I've added tests for scenarios I could think of.

Kindly let me know if there's something I can do better here.

Signed-off-by: Nithin <nithin.linkin@gmail.com>

### DIFF
--- a/options.go
+++ b/options.go
@@ -330,6 +330,8 @@ type LogOptions struct {
 	// set Order=LogOrderCommitterTime for ordering by committer time (more compatible with `git log`)
 	// set Order=LogOrderBSF for Breadth-first search
 	Order LogOrder
+
+	FileName *string
 }
 
 var (
--- a/options.go
+++ b/options.go
@@ -331,6 +331,8 @@ type LogOptions struct {
 	// set Order=LogOrderBSF for Breadth-first search
 	Order LogOrder
 
+	// Show only those commits in which the specified file was inserted/updated.
+	// It is equivalent to running `git log -- <file-name>`.
 	FileName *string
 }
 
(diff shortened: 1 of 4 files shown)


plumbing/object/commit_walker_file.go
```diff
@@ -0,0 +1,114 @@
...
+func (c *commitFileIter) Next() (*Commit, error) {
+	var err error
+	if c.currentCommit == nil {
+		c.currentCommit, err = c.sourceIter.Next()
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	for {
```
mcuadros: can you move this logic to a function?

options.go
```diff
@@ -330,6 +330,8 @@ type LogOptions struct {
 	// set Order=LogOrderCommitterTime for ordering by committer time (more compatible with `git log`)
 	// set Order=LogOrderBSF for Breadth-first search
 	Order LogOrder
+
```
mcuadros: Can you document this?

plumbing/object/commit_walker_file.go
gnithin: Done

options.go
gnithin: Done

PR #982, opened 2018-10-12
tree: add a Size() method for getting plaintext size

strib:
And add a method to the storer.Object interface to make this possible.

This allows the user of a `Tree` to get the object size without having to read in the entire file.


plumbing/format/packfile/packfile.go
```diff
@@ -90,6 +90,22 @@ func (p *Packfile) GetByOffset(o int64) (plumbing.EncodedObject, error) {
 	return p.nextObject()
 }
 
+func (p *Packfile) GetSizeByOffset(o int64) (size int64, err error) {
```
mcuadros: Can you add the documentation to this method?

Issue #954, opened 2018-09-09
Unable to checkout by an annotated tag


anisimovsergey:
Version: v4.6.0
Revision: 7b6c1266556f59ac436fada3fa6106d4a84f9b56

My repository has several tags and some of them are annotated and some aren't. When I'm trying  to do the following:

```go
	tags, err := repo.Tags()
	if err != nil {
		return err
	}

	w, err := repo.Worktree()
	if err != nil {
		return err
	}

	err = tags.ForEach(func(t *plumbing.Reference) error {
		err = w.Checkout(&git.CheckoutOptions{Hash: t.Hash()})
		if err != nil {
			fmt.Printf("unable to check out commit %s", t.Hash())
		} 
		return nil
	})
```

I'm getting this error: "object not found"

I searched among the issues and found #772, there is a PR for solving it created by @novas0x2a but I've got a feeling that the problem might be deeper.

PR #985, opened 2018-10-15
repository: improve CheckoutOption.Hash doc

mcuadros:
The `CheckoutOption.Hash` can be used only with hash commits, this PR clarifies this on the documentation.

https://github.com/src-d/go-git/issues/954

### DIFF
--- a/options.go
+++ b/options.go
@@ -231,8 +231,9 @@ var (
 
 // CheckoutOptions describes how a checkout 31operation should be performed.
 type CheckoutOptions struct {
-	// Hash to be checked out, if used HEAD will in detached mode. Branch and
-	// Hash are mutually exclusive, if Create is not used.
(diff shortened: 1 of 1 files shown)


options.go
```diff
@@ -231,8 +231,9 @@ var (
 
 // CheckoutOptions describes how a checkout 31operation should be performed.
 type CheckoutOptions struct {
-	// Hash to be checked out, if used HEAD will in detached mode. Branch and
-	// Hash are mutually exclusive, if Create is not used.
+	// Hash, is the hash of the commit to be checked out, if used HEAD will in
```
smola: Maybe rephrase a little bit:

> Hash is the hash of the commit to be checked out. If used, HEAD will be in detached mode. If Create is not used, Branch and Hash are mutually exclusive.

PR #990, opened 2018-10-19
examples: PlainClone with Basic Authentication (Password & Access Token)

mccurdyc:
## Proposed Additions
+ Add a GoDoc example for cloning with `BasicAuth`
+ Add a GoDoc example for cloning with "Token" Auth
+ Add cli example for cloning with `BasicAuth` to `_examples`
+ Add cli example for cloning with "Token" Auth


mccurdyc:
Why does [`TokenAuth`](https://godoc.org/gopkg.in/src-d/go-git.v4/plumbing/transport/http#TokenAuth) not work?

If this is something that would be of value, and there isn't a specific reason why it hasn't been implemented, I would love to deep dive into this. I think it would be a great follow-up, then we could clean up the "Token" auth.

smola:
@mccurdyc Does TokenAuth not work? It was added to support OAuth and it seemed to work, at least in Bitbucket, https://github.com/src-d/go-git/pull/859

If we are talking about GitHub's _personal access tokens_, using them as password is the intended usage AFAIK: https://help.github.com/articles/creating-a-personal-access-token-for-the-command-line/#using-a-token-on-the-command-line

smola:
Looks good to me, thanks!

> I apologize, I was incorrectly using the GitHub personal access token.

Common issue, since it is not documented ;-)

mcuadros:
@mccurdyc can you add this examples to the https://github.com/src-d/go-git/blob/master/_examples/README.md?

mccurdyc:
@mcuadros I have added those to the list of examples. Thanks!

PR #994, opened 2018-10-23
plumbing/format/packfile: Fix broken "thin" packfile support. Fixes #991

### LINKED ISSUE #991: Bug: Parser crashes on thin packs
When the parser is fed a so-called "thin pack", crashes in the `get()` function since it will receive a `nil` in the `Parent` `objInfo` property. A "thin pack" is a pack that references a SHA1 that is not contained in the pack and that the other side knows you must have.
The portion of the code that handles thin packs is not covered.

I have a patch for this that works and I will submit shortly, but I'd need help in writing tests (I need a fixture that produces thin packs)

jpeletier:
We have noticed that thin-pack support by the parser stopped working after upgrading from 4.5.0 to the latest version 4.7.1, but going back, it doesn't work since 4.6.0.
When the parser is given a so-called "thin" pack file, it will panic with a nil-pointer exception in `Parser.get()` as this function receives a nil `o` `*objectInfo` parameter. I traced this back to `indexObjects()` `case plumbing.REFDeltaObject` where the `Parent` property is left uninitialized when a REF delta object base is not found in the current pack. Unit tests did not check for thin-pack eventuality, and thus this case was uncovered.

This is an attempt to fix the issue. The strategy is to initialize `objInfo.Parent` to a "placeholder" parent, so we can detect this during the resolve phase.

I would need your guidance in writing tests for this one, as I am not sure how to produce a thin pack file out of the fixtures: these types of pack files only exsist on the network when git is moving stuff, never stored on disk.

### DIFF
--- a/plumbing/format/packfile/parser_test.go
+++ b/plumbing/format/packfile/parser_test.go
@@ -1,10 +1,13 @@
 package packfile_test
 
 import (
+	"io"
 	"testing"
 
+	git "gopkg.in/src-d/go-git.v4"
 	"gopkg.in/src-d/go-git.v4/plumbing"
 	"gopkg.in/src-d/go-git.v4/plumbing/format/packfile"
+	"gopkg.in/src-d/go-git.v4/plumbing/storer"
 
 	. "gopkg.in/check.v1"
 	"gopkg.in/src-d/go-git-fixtures.v3"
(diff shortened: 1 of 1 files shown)


jpeletier:
@jfontan I have added a test for this, please take a look.
I was not able to use your method above to create a thin packfile. Upon exploring the generated packfile, I detected it actually was embedding the dependencies in.
To create a truly thin packfile I resorted to using `git bundle` to produce a differential bundle, then removing the bundle header. The result is the new `thinpack` fixture.
I had to add that new fixture to the fixtures repo, so this won't work until https://github.com/src-d/go-git-fixtures/pull/11 is merged

Please check it out and let me know. Thanks.

PR #997, opened 2018-10-24
Update COMPATIBILITY.md

dharmendra-lingaiah:
If flags are plural, then usage of aren't is warranted. If it is a single flag, we should use isn't


mcuadros:
Please re-open when the DCO once signed, thanks

PR #1000, opened 2018-10-26
add StackOverflow to support channels

smola:
Since we are now redirecting users to StackOverflow for support
questions, it makes sense to add it to the official support channels.


mcuadros:
Congratulations you won a Furby for being the PR number 1000!!! 

Here is your prize:
91eughbattl _sx425_
/>

PR #1006, opened 2018-10-29
storage: transactional, new storage with transactional capabilities

mcuadros:
This PR introduces a new storage, called transactional allow to do Rollback any kind of operation over the storage, discarding Objects, Config, References, etc

### DIFF
--- a/storage/transactional/config.go
+++ b/storage/transactional/config.go
@@ -2,6 +2,7 @@ package transactional
 
 import "gopkg.in/src-d/go-git.v4/config"
 
+// ConfigStorage implements the storer.ConfigStorage for the transactional package.
 type ConfigStorage struct {
 	config.ConfigStorer
 	temporal config.ConfigStorer
@@ -9,10 +10,13 @@ type ConfigStorage struct {
 	set bool
 }
 
+// NewConfigStorage returns a new ConfigStorer based on a base storer and a
+// temporal storer.
 func NewConfigStorage(s, temporal config.ConfigStorer) *ConfigStorage {
 	return &ConfigStorage{ConfigStorer: s, temporal: temporal}
 }
 
+// SetConfig honors the storer.ConfigStorer interface.
 func (c *ConfigStorage) SetConfig(cfg *config.Config) error {
 	if err := c.temporal.SetConfig(cfg); err != nil {
 		return err
@@ -22,6 +26,7 @@ func (c *ConfigStorage) SetConfig(cfg *config.Config) error {
 	return nil
 }
 
+// Config honors the storer.ConfigStorer interface.
 func (c *ConfigStorage) Config() (*config.Config, error) {
 	if !c.set {
 		return c.ConfigStorer.Config()
@@ -30,6 +35,7 @@ func (c *ConfigStorage) Config() (*config.Config, error) {
 	return c.temporal.Config()
 }
 
+// Commit it copies the config from the temporal storage into the base storage.
 func (c *ConfigStorage) Commit() error {
 	if !c.set {
 		return nil
--- a/storage/transactional/config.go
+++ b/storage/transactional/config.go
@@ -1,0 +1,36 @@
+package transactional
+
+import "gopkg.in/src-d/go-git.v4/config"
+
+type ConfigStorage struct {
+	config.ConfigStorer
+	temporal config.ConfigStorer
+
+	set bool
+}
+
+func NewConfigStorage(s, temporal config.ConfigStorer) *ConfigStorage {
+	return &ConfigStorage{ConfigStorer: s, temporal: temporal}
+}
+
+func (c *ConfigStorage) SetConfig(cfg *config.Config) error {
+	if err := c.temporal.SetConfig(cfg); err != nil {
+		return err
+	}
+
+	c.set = true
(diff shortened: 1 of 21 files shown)


smola:
What would be the relation of this storage with our current `object.Transactioner` interface?

storage/transactional/config.go
```diff
@@ -0,0 +1,36 @@
...
+	if !c.set {
+		return nil
+	}
+
+	cfg, err := c.temporal.Config()
+	if err != nil {
+		return err
+	}
+
+	return c.ConfigStorer.SetConfig(cfg)
```
ajnavarro: This might not be transactional depending on the ConfigStorer implementation under the hood. Per example, using filesystem implementation `SetConfig` can fail on `Write`, causing a partial write.

storage/transactional/object.go
```diff
@@ -0,0 +1,63 @@
+package transactional
```
ajnavarro: No `Commit` method here, when the objects from temporal are copied to the main storer?

storage/transactional/object.go
```diff
@@ -0,0 +1,63 @@
...
+func (o *ObjectStorage) EncodedObject(t plumbing.ObjectType, h plumbing.Hash) (plumbing.EncodedObject, error) {
+	obj, err := o.EncodedObjectStorer.EncodedObject(t, h)
+	if err == plumbing.ErrObjectNotFound {
+		return o.temporal.EncodedObject(t, h)
+	}
+
+	return obj, err
+}
+
+func (o *ObjectStorage) IterEncodedObjects(t plumbing.ObjectType) (storer.EncodedObjectIter, error) {
```
ajnavarro: Mixed concepts here. If we are talking about `transactional` object storage, objects shouldn't be read until `Commit` is called.

Also, depending on the use case, maybe we want to do the opposite, a `Rollback`.

mcuadros:
None, most likely in v5, the old transaction should be deleted.

storage/transactional/config.go
smola: Indeed, this can be a problem. At least for our use case, since we don't do a commit per operation, but after a group of operations. That is, we want a Commit on the full storage, not just part of it.

storage/transactional/reference.go
```diff
@@ -0,0 +1,108 @@
...
+var ErrRefHasChanged = fmt.Errorf("reference has changed concurrently")
+
+type ReferenceStorage struct {
+	storer.ReferenceStorer
+	temporal storer.ReferenceStorer
+
+	// deleted, remaining references at this maps are going to be deleted when
+	// commit is requested, the entries are added when RemoveReference is called
+	// and deleted if SetReference is called.
+	deleted map[plumbing.ReferenceName]struct{}
```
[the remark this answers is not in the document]
smola: Sounds good.

However, I'm still concerned that the whole approach is not really usable by borges. If that was the goal, it would actually need its own implementation of `Commit`...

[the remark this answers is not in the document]
mcuadros: Yes, the implementation of `Commit` should be ad-hoc for each interface, but I don't think is going to be the current one.

mcuadros:
@smola @ajnavarro PTAL

smola:
I'm thinking about how this fits in borges. It looks like it doesn't, since it would produce just additional copies of everything in filesystem for no additional benefit.

mcuadros:
I don't understand several things:
1) Why this should be something that works with the current borges implementation
2) Why you say that produces additional copies

smola:
> Why this should be something that works with the current borges implementation

It does not need to work with _current_ implementation, but I thought this was meant to be used by borges _future_ implementation. So I would expect the result to be equal or better in terms of performance, robustness, etc.

> Why you say that produces additional copies

Well, I have looked better to it and it looks like additional copies would not be necessary when borges provides a Commit implementation.

It currently looks like this:

1. Begin transaction -> a siva file is copied from remote/permanent storage to temporary/local storage.
2. Perform changes -> appends each change directly to the local siva file.
3a. Commit -> Move local siva file to remote storage.
3b or Rollback -> Remove local siva file.

This could be, in theory, optimized as follows:

1. Begin transaction -> no-op
2. Perform changes -> append directly to a new siva block in local/temporary storage, possibly reading directly the previous blocks from remote storage if needed.
3a. Commit -> Concat new block in local storage to remote storage.
3b or Rollback -> Remove new block from local storage.

With this new implementation, if we integrated it with borges, I see it being something like this:

1. Begin transaction -> kind of no-op / lazily creating a new temporary local storage, this can be either a full siva copy (as in our current solution) or a new empty siva block (as in the optimized solution)
2. Perform changes -> perform changes on temporary storage, appended directly to the local siva file or block.
3a. Commit -> Move local siva file to remote storage or Concat local siva block to remote storage.
3b or Rollback -> Remove local siva file or block.

Would it be like this? Looks like it could actually work well with the custom Commit for borges, not sure if I'm missing something.

smola:
@mcuadros What about providing an interface too?

mcuadros:
I marked this package as experimental, is this way we can merged and used in the other projects, and changed it is required. 

//cc @smola @jfontan

PR #1007, opened 2018-10-30
Update LICENSE

vyaspranjal33:
Updated year and organization name in LICENSE file


mcuadros:
We will update the year, the company nade you entered is wrong, thanks

PR #1009, opened 2018-10-31
http: improve TokenAuth documentation

### LINKED ISSUE #999: TokenAuth not working as expected
I expect the following to clone a private repository using a personal access token.

```go
...
	r, err := git.PlainClone(<directory>, false, &git.CloneOptions{
		URL:               <url>,
		RecurseSubmodules: git.DefaultSubmoduleRecursionDepth,
		Auth: &http.TokenAuth{
			Token: <personal_access_token>,
		},
	})
...
```

```bash
git clone https://github.com/mccurdyc/private-repo simple --recursive
error: unexpected client error: unexpected requesting "https://github.com/user/private-repo/info/refs?service=git-upload-pack" status code: 400
```

It looks like GitHub should support it: 
https://blog.github.com/2012-09-21-easier-builds-and-deployments-using-git-over-https-and-oauth/

```bash
git clone https://<personal_access_token>@github.com/user/private-repo.git
```

---
As a side note, looks like the following work:

1.
```bash
curl -H 'Authorization: Bearer <personal_access_token>' https://api.github.com/user/repos?visibility=private
```
2.
```bash
curl -H 'Authorization: token <personal_access_token>' https://api.github.com/user/repos?visibility=private
```

The current implementation of [`TokenAuth` in go-git](https://github.com/src-d/go-git/blob/master/plumbing/transport/http/common.go#L209) does something similar to 2, except would only work for GitHub's API and not for cloning.

@smola added some context to this previously [here](https://github.com/src-d/go-git/pull/990#issuecomment-432540235)

Also, [this goplayground example](https://play.golang.org/p/vcfVaBWV6Q5) aligns with the [GitHub post](https://blog.github.com/2012-09-21-easier-builds-and-deployments-using-git-over-https-and-oauth/) in that you would send the token as username.

> Next, you can simply use an OAuth token for the username and either a blank password or the string x-oauth-basic when cloning a repository.

smola:
Users are often confused with TokenAuth, since it might look that it
should be used with GitHub's OAuth tokens. But that is not the case.

TokenAuth implements HTTP bearer authentication. Most git servers will
use HTTP basic authentication (user+passwords) even for OAuth tokens.

Fixes #999

### DIFF
--- a/plumbing/transport/http/common.go
+++ b/plumbing/transport/http/common.go
@@ -214,7 +214,14 @@ func (a *BasicAuth) String() string {
 	return fmt.Sprintf("%s - %s:%s", a.Name(), a.Username, masked)
 }
 
-// TokenAuth implements the go-git http.AuthMethod and transport.AuthMethod interfaces
+// TokenAuth implements an http.AuthMethod that can be used with http transport
+// to authenticate with HTTP token authentication (also known as bearer
+// authentication).
+//
+// IMPORTANT: If you are looking to use OAuth tokens with popular servers (e.g.
+// GitHub, Bitbucket, GitLab) you should use BasicAuth instead. These servers
+// use basic HTTP authentication, with the OAuth token as user or password.
+// Check the documentation of your git server for details.
 type TokenAuth struct {
 	Token string
 }


smola:
cc @mccurdyc

PR #1019, opened 2018-11-11
storage/filesystem: Added reindex method to  reindex packfiles

jpeletier:
This PR adds a simple method to the Filesystem `ObjectStorage` object so the user can invoke a reindex of all packfiles. This is useful in repos in which go-git does not have exclusive access and git manipulates packfiles, such as when using `git bundle`, or repacking externally.

### DIFF
--- a/storage/filesystem/object.go
+++ b/storage/filesystem/object.go
@@ -61,6 +61,11 @@ func (s *ObjectStorage) requireIndex() error {
 	return nil
 }
 
+// Reindex indexes again all packfiles. Useful if git changed packfiles externally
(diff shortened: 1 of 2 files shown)


mcuadros:
Thanks for the PR, can you signed-off the commit? And provide a test?

jpeletier:
@mcuadros I have added a test for this.
I had to add a new fixture to the fixtures repo, so this won't work until https://github.com/src-d/go-git-fixtures/pull/11 is merged.
Please check and let me know. Thanks

mcuadros:
Looks like is failing in Windows

PR #1031, opened 2018-11-30
git: return better error message when packfile cannot be downloaded

jfontan:
Previously the error message when the connection was closed while fetching was "object not found" and was misleading. Now when the packfile size is 0 the error "unable to fetch packfile" is returned.

Does not come with test as I haven't found a way to correctly break the connection while the packfile is being downloaded. Starting `git daemon` and killing it after some time is not deterministic enough and can be flaky.

Related to src-d/borges#314

### DIFF
--- a/plumbing/format/packfile/common.go
+++ b/plumbing/format/packfile/common.go
@@ -51,7 +51,13 @@ func WritePackfileToObjectStorage(
 	}
 
 	defer ioutil.CheckClose(w, &err)
-	_, err = io.Copy(w, packfile)
+
+	var n int64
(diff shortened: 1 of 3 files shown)


mcuadros:
You can test it, giving to it an empty io.Reader, no?

jfontan:
I've added a small test that check for `ErrEmptyPackfile` error with an empty packfile. Still it does not test that a broken connection returns that error.

smola:
Keep in mind that this needs to trigger a minor, not patch version (new error returned from existing function).

PR #1036, opened 2018-12-10
storage/dotgit: use fs capabilities in setRef

jfontan:
Do not use tags to check if the filesystem is able to open files in read/write mode.

Related to #864


mcuadros:
Can you provide a test?

jfontan:
Now both paths are tested.

Issue #1024, opened 2018-11-19
Expose git log --all equivalent


smola:
Currently, our main way to access all commits is `CommitObjects`, which iterates all commits in a repository, no matter if they are reachable from references or not. This leads to two problems:

1. It is sometimes confusing to iterate over dangling commits. These are very frequent on local repositories that have been subject to local work, rebases, etc. In git, it is pretty hard to iterate over dangling commits. AFAIK there is no porcelain command exposing all of them together (`git reflog` exposes some of them from reflog, git verify-pack exposes them in packfiles, etc).
2. On large repositories, iterating over all commits is significantly slower than getting all reachable commits. The former requires decoding all object headers in every packfile, which on large repositories is slower than just traversing history from all references.

Changing `CommitObjects` behaviour would not be backwards compatible, but we could expose an equivalent to `git log --all` in `Repository.Log` by adding an `All bool` option, which would return an iterator over all commits reachable from any reference.

https://github.com/src-d/go-git/blob/7853ab677e1eb8e5237a4ba2d7caef9f7ac75b28/options.go#L323-L338

We would then add a note to `CommitObjects` godoc, since users will usually want to use `Log` instead.

Issue #1023, opened 2018-11-16
Proposal: Make CommitObjects(), BlobObjects() and so on methods do not return unreachable objects due to performance problems. 


ajnavarro:
Right now, objects interator methods: `TreeObjects()`, `CommitObjects()`, `BlobObjects()`, `TagObjects()` are returning also unreachable objects.

This is because we are iterating the whole packfile, reading object headers, and returning only the ones that match with the requested type. This is causing really slow iterators on really large repositories because we are skipping constantly other objects.

To avoid that, we should just return objects that are referenced by a git reference.

Just as an example, on `CommitObjects()` iterator we should simulate the `git log --all` git command, that means, get all the hashes from references, and start walking the commit history.

WDYT @smola @mcuadros ?

smola:
@ajnavarro I think go-git behavior is correct at the moment, even if it could use better documentation. Even if it weren't, this is not a backwards compatible change and cannot be done in v4.

But the concern is legitimate for a lot of use cases, including gitbase (https://github.com/src-d/gitbase/issues/617). That's why I would like to propose to solve the problem as follows:

1. Change gitbase table implementations to do the right thing.
2. Provide required APIs in go-git when they are missing.
2.1. For example, for commits, we already talked about providing an additional option (`All bool`) to `LogOptions` to mimmick `git log --all`, that should be usable by gitbase for the `commits` table.

smola:
I've filed the git.Log one here: https://github.com/src-d/go-git/issues/1024
I'm still thinking about the right API for other object types.

PR #1044, opened 2019-01-03
Implement git log --all

kuba--:

This PR implements `git log --all`.
It closes https://github.com/src-d/go-git/issues/1024
It's also related to https://github.com/src-d/go-git/issues/1023

Please take a look carefully if it can be done in more optimal way.

### DIFF
--- a/plumbing/object/commit_walker.go
+++ b/plumbing/object/commit_walker.go
@@ -1,6 +1,7 @@
 package object
 
 import (
+	"container/list"
 	"io"
 
 	"gopkg.in/src-d/go-git.v4/plumbing"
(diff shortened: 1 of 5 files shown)


plumbing/object/commit_walker.go
```diff
@@ -181,3 +182,107 @@ func (w *commitPostIterator) ForEach(cb func(*Commit) error) error {
...
+func NewCommitAllRefsIter(
+	storer storer.EncodedObjectStorer,
+	commitIterFunc func(*Commit) CommitIter,
+	commitIter CommitIter,
+	refIter storer.ReferenceIter) CommitIter {
+
+	l := list.New()
+	m := make(map[plumbing.Hash]*list.Element)
+
+	commitIter.ForEach(func(c *Commit) error {
```
kuba--: Maybe we don't need to populate a list by _first iterator_ (in other words for --all we may ignore `From` option and just go through all references)

PR #1045, opened 2019-01-04
Implement git log --all

### LINKED ISSUE #1024: Expose git log --all equivalent

kuba--:

This PR implements git log --all.
It closes #1024
It's also related to #1023

`git log --all` pretends as if all the refs in refs/, along with HEAD, are listed on the command line as <commit>. We start from the HEAD, so `From` hash is ignored for `--all` option.

Please take a look carefully if it can be done in more optimal way.
For _filter by file_ (`git log --all --filename`) we have to double check if the parent commit comes from the real commit (not just the next object returned by iterator).

### DIFF
--- a/plumbing/object/commit_walker.go
+++ b/plumbing/object/commit_walker.go
@@ -1,10 +1,12 @@
 package object
 
 import (
+	"container/list"
 	"io"
 
 	"gopkg.in/src-d/go-git.v4/plumbing"
 	"gopkg.in/src-d/go-git.v4/plumbing/storer"
+	"gopkg.in/src-d/go-git.v4/storage"
 )
 
 type commitPreIterator struct {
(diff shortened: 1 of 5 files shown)


plumbing/object/commit_walker.go
```diff
@@ -181,3 +183,127 @@ func (w *commitPostIterator) ForEach(cb func(*Commit) error) error {
...
+			// we already have the HEAD
+			return nil
+		}
+		c, _ := GetCommit(s, r.Hash())
+		// if it's not a commit - skip it.
+		if c == nil {
+			return nil
+		}
+
+		el, ok := m[c.Hash]
```
jfontan: This check may be done before `GetCommit` as `c.Hash` should be the same as `r.Hash()`.

plumbing/object/commit_walker.go
```diff
@@ -181,3 +183,127 @@ func (w *commitPostIterator) ForEach(cb func(*Commit) error) error {
...
+		}
+
+		var refCommits []*Commit
+		cit := fn(c)
+		for c, e := cit.Next(); e == nil; {
+			el, ok = m[c.Hash]
+			if ok {
+				break
+			}
+			refCommits = append(refCommits, c)
```
jfontan: Is this storing all commits in memory? In big repositories this can be a lot.

kuba--: kind of ... it's storing ref commits till it finds an existing hash in the list, then it inserts all memorised commits into the list and clears the slice, but generally speaking the list `l` is a flatten list of all commits ordered by `commitIterFunc`, so yes it can be a lot. It's a first approach, but maybe we should think how to make it in more optimal way. Actually all our iterators don't retrieve any data on demand but just return an object with some offset already stored in memory, e.g.: `Next() { return stack[offset] }`

jfontan: What mean is that you probably only need to store the hash, not the whole `Commit` data. I believe it's not used at all:

```go
type Commit struct {
    Hash Hash
    Author Signature
    Committer Signature
    PGPSignature string
    Message string
    TreeHash Hash
    ParentHashes []Hash
    s EncodedObjectStorer
}
```

kuba--: I think it's not really just Hash, because at the end we need to return `CommitIter` which satisfies:
```
func Next() (*Commit, error)
```

What I can see as a semi-optimization is to keep in memory just hashes, but in `Next()` just call `GetCommit`.

jfontan: Or move this code (the whole iteration) to `Next` so `GetCommit` is only called once.

kuba--: But we need to have flatten list firs, before we start returning `Next` commit. At the beginning you don't know which ref inserts own commits where (also it depends on order)

plumbing/object/commit_walker.go
```diff
@@ -181,3 +183,127 @@ func (w *commitPostIterator) ForEach(cb func(*Commit) error) error {
...
+	if err != nil {
+		return nil, err
+	}
+
+	refIter, err := s.IterReferences()
+	if err != nil {
+		return nil, err
+	}
+	defer refIter.Close()
+	err = refIter.ForEach(func(r *plumbing.Reference) error {
```
mcuadros: will be great if we can split this function in something more readable

plumbing/object/commit_walker.go
```diff
@@ -181,3 +183,127 @@ func (w *commitPostIterator) ForEach(cb func(*Commit) error) error {
...
+// commitAllIterator stands for commit iterator for all refs.
+type commitAllIterator struct {
+	// el points to the current commit.
+	el *list.Element
+}
+
+// NewCommitAllIter returns a new commit iterator for all refs.
+// s is a repo Storer used to get commits and references.
+// fn is a commit iterator function, used to iterate through ref commits in chosen order
+func NewCommitAllIter(s storage.Storer, fn func(*Commit) CommitIter) (CommitIter, error) {
```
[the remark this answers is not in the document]
mcuadros: Maybe worth to do it now.

plumbing/object/commit_walker_file.go
```diff
@@ -1,23 +1,26 @@
...
 	sourceIter    CommitIter
 	currentCommit *Commit
+	all           bool
 }
 
 // NewCommitFileIterFromIter returns a commit iterator which performs diffTree between
 // successive trees returned from the commit iterator from the argument. The purpose of this is
 // to find the commits that explain how the files that match the path came to be.
-func NewCommitFileIterFromIter(fileName string, commitIter CommitIter) CommitIter {
+func NewCommitFileIterFromIter(fileName string, commitIter CommitIter, all bool) CommitIter {
```
mcuadros: Propagating here the word `all` maybe is not the best option since is not very meaningful

kuba--: yeah, actually it should be something like `checkParents`

plumbing/object/commit_walker_file.go
```diff
@@ -73,8 +76,24 @@ func (c *commitFileIter) getNextFileCommit() (*Commit, error) {
 
 		foundChangeForFile := false
 		for _, change := range changes {
-			if change.name() == c.fileName {
+			if change.name() != c.fileName {
+				continue
+			}
+
+			// filename matches, now check if source iterator contains all commits (from all refs)
+			if c.all {
```
mcuadros: This snippet need a refactor

repository.go
```diff
@@ -1027,41 +1027,64 @@ func (r *Repository) PushContext(ctx context.Context, o *PushOptions) error {
...
+		commitIterFunc = func(c *object.Commit) object.CommitIter {
+			return object.NewCommitIterCTime(c, nil, nil)
+		}
 	default:
 		return nil, fmt.Errorf("invalid Order=%v", o.Order)
 	}
 
-	if o.FileName == nil {
-		return commitIter, nil
+	if o.All {
```
mcuadros: this code can be move to a function

plumbing/object/commit_walker.go
jfontan: After an offline discussion it seems there's not a straightforward way to create this list on the fly. The biggest problem is the order. Also storing all commit structs in memory may not occupy as much as I previously thought.

plumbing/object/commit_walker.go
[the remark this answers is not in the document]
kuba--: Agreed, but it should go in another PR (IMO)

repository.go
kuba--: What exactly do you suggest to extract?
switch-case? else-branch?

mcuadros: I don't know exactly the problem is that I think is very nested and complicated code with a high cyclomatic complexity

kuba--:
@mcuadros - I've refined Log function. PTAL.

PR #1056, opened 2019-01-30
storage/filesystem: check file object before using cache

jfontan:
If the cache is shared between several repositories `getFromUnpacked` can erroneously return an object from other repository.

This decreases performance a little bit as there's an extra fs operation when the object is in the cache but is correct when the cache is shared.


filipnavara:
It makes me wonder why is it a problem that an object from another repository is returned. As long as the hash is the same you should be getting the same object. I also though that FSObject is never ending up in the cache which would pose a problem if the pack file was to disappear from the storage.

PR #1067, opened 2019-02-12
Ignore missing references/objects on log --all

ajnavarro:
To mimic the actual standard git behavior, we must ignore references that are pointing to wrong/unexistent objects.

Signed-off-by: Antonio Jesus Navarro Perez <antnavper@gmail.com>

### DIFF
--- a/plumbing/object/commit_walker.go
+++ b/plumbing/object/commit_walker.go
@@ -197,26 +197,38 @@ func NewCommitAllIter(repoStorer storage.Storer, commitIterFunc func(*Commit) Co
 	commitsPath := list.New()
 	commitsLookup := make(map[plumbing.Hash]*list.Element)
 	head, err := storer.ResolveReference(repoStorer, plumbing.HEAD)
-	if err != nil {
+	if err != nil && err != plumbing.ErrReferenceNotFound {
 		return nil, err
+	} else if err != plumbing.ErrReferenceNotFound {
+		if err = addReference(repoStorer, commitIterFunc, head, commitsPath, commitsLookup); err != nil {
+			return nil, err
+		}
 	}
 
 	// add all references along with the HEAD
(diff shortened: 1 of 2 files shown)


plumbing/object/commit_walker.go
```diff
@@ -197,26 +197,38 @@ func NewCommitAllIter(repoStorer storage.Storer, commitIterFunc func(*Commit) Co
 	commitsPath := list.New()
 	commitsLookup := make(map[plumbing.Hash]*list.Element)
 	head, err := storer.ResolveReference(repoStorer, plumbing.HEAD)
-	if err != nil {
+	if err != nil && err != plumbing.ErrReferenceNotFound {
 		return nil, err
+	} else if err != plumbing.ErrReferenceNotFound {
```
kuba--: Does it make sense?

kuba--: maybe it should be `err == plumbing.ErrReferenceNotFound`

ajnavarro: @kuba-- maybe:
```suggestion
	} else if err == nil {
```

Is more clear?

ajnavarro: @kuba-- I only want to add the head reference if I actually have the reference, so only if the error is not plumbing.ErrReferenceNotFound

Maybe I can check if `err == nil` or `head != nil`

kuba--: So, maybe check if head is not nil (is it possible)?
Because if err is not plumbing.ErrReferenceNotFound does not mean that err != nil.
My question is - shall we consider just plumbing.ErrReferenceNotFound? What do we want to do with other errors?

Maybe what we had at the beginning `err != nil` is good enough?

ajnavarro:
@kuba-- done! and thanks for the tip!

ajnavarro:
@kuba-- I think the logic is ok

PR #1070, opened 2019-02-21
worktree: allow manual ignore patterns when no .gitignore is available

fbsb:
I am using go-git as a library for a personal project and noticed hard-coded excluded patterns are ignored when no `.gitignore` file exists at the root of the repository.

This PR fixes this behavior and allows hard-coded ignore patterns even when no `.gitignore` file is present


mcuadros:
Can you provide a test?

fbsb:
@mcuadros without the patch the test would fail when a .gitignore would not be present in the repository.

PR #1078, opened 2019-02-26
Add merge-base command example

dpordomingo:
Command example implementing `git merge-base` behavior, with `--is-ancestor` and `--independent` modifiers.

usage: 
```shell
$ ./merge_base --help

./merge_base --helpReturns the merge-base between two commits:
./merge_base <path> <baseRev> <headRev>
./merge_base <path> --is-ancestor <baseRev> <headRev>
./merge_base <path> --independent <commitRev>...
```

To offer those features, it was implemented a new walker `filterCommitIter` (like `bfsCommitIterator` one), and three functions `MergeBase`, `IsAncestor` and `Independents`.
- `filterCommitIter` 
  > _it's a `CommitIter` that walks the commit history (in Breadth-first order) of the passed commit. It only returns the commits that validate the passed `isValid CommitFilter`, and only traverse the commits that do not validate `isLimit CommitFilter`._
- `MergeBase(first, second *object.Commit) ([]*object.Commit, error)`
  > _it mimics the behavior of `git merge-base a b`, returning the best common ancestor of the two passed commits; the best common ancestor cannot be reached from other common ancestors._
- `IsAncestor(first, second *object.Commit) (bool, error)` 
  > _it returns `true` if the first commit is an ancestor of the second one, like `git merge --is-ancestor` does_
- `Independents(commits []*object.Commit) ([]*object.Commit, error)` 
  > _it returns a subset of the passed commits, that are not reachable from any other, as `git merge-base --independent commit...` does_

## alternative:

imo the new feature could be part of `go-git` itself, instead of being an `_example`; to do so, I think that `filterCommitIter` could replace `bfsCommitIterator` or be moved into `go-git/plumbing/object` package, and `MergeBase`, `IsAncestor` and `Independents` could be moved into `git.Repository` struct.

If you think it could be feasible as proposed, or with any change, let me know and will work over it during the next OSD :tada:

### DIFF
--- a/_examples/merge_base/main.go
+++ b/_examples/merge_base/main.go
@@ -1,0 +1,133 @@
+package main
+
+import (
+	"fmt"
+	"os"
+	"strings"
+
+	"gopkg.in/src-d/go-git.v4"
+	"gopkg.in/src-d/go-git.v4/_examples/merge_base/repository"
+	"gopkg.in/src-d/go-git.v4/plumbing"
+	"gopkg.in/src-d/go-git.v4/plumbing/object"
+
+	. "gopkg.in/src-d/go-git.v4/_examples"
+)
+
+func helpAndExit(s string, exitCode int) {
+	if s != "" {
+		Warning("%s", s)
+	}
+
+	Warning("%s %s", os.Args[0], "<path> <baseRev> <headRev>")
+	Warning("%s %s", os.Args[0], "<path> --is-ancestor <baseRev> <headRev>")
+	Warning("%s %s", os.Args[0], "<path> --independent <commitRev>...")
+
+	os.Exit(exitCode)
+}
+
+// Command that mimics `git merge-base <baseRev> <headRev>`
+// Command that mimics `git merge-base --is-ancestor <baseRev> <headRev>`
+// Command that mimics `git merge-base --independent <commitRev>...`
+func main() {
+
+	if len(os.Args) < 2 || os.Args[1] == "--help" || os.Args[1] == "-h" {
+		helpAndExit("Returns the merge-base between two commits:", 0)
+	}
+
+	if len(os.Args) < 4 {
+		helpAndExit("Wrong syntax, Usage:", 1)
+	}
+
+	path := os.Args[1]
+
+	var modeIndependent, modeAncestor bool
+	var commitRevs []string
+	var res []*object.Commit
+
+	switch os.Args[2] {
+	case "--independent":
+		modeIndependent = true
+		commitRevs = os.Args[3:]
+	case "--is-ancestor":
+		modeAncestor = true
+		commitRevs = os.Args[3:]
+		if len(os.Args) != 5 {
+			helpAndExit("Wrong syntax, Usage:", 1)
+		}
+	default:
+		commitRevs = os.Args[2:]
+		if len(os.Args) != 4 {
+			helpAndExit("Wrong syntax, Usage:", 1)
+		}
+	}
+
+	// Open a git repository from the given path
+	repo, err := git.PlainOpen(path)
+	CheckIfError(err)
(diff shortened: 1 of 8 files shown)


mcuadros:
Will be nice to add a README to this example to understand how works

dpordomingo:
Would it be enough adding that info to the command itself?
Something like the following, that I addapted from `git merge-base` command.

Running it wothout no options:
```
$ ./merge_base

usage: merge_base <commitRev> <commitRev>
   or: merge_base --independent <commitRev>...
   or: merge_base --is-ancestor <commitRev> <commitRev>

    (no option)       lists the best common ancestors of the two passed commits
    --independent     list commits not reachable from the others
    --is-ancestor     is the first one ancestor of the other?
```

with `--help`
```
$ ./merge_base --help

NAME
   merge_base - Lists the best common ancestors of the two passed commit revisions

SYNOPSIS
 usage: merge_base <commitRev> <commitRev>
    or: merge_base --independent <commitRev>...
    or: merge_base --is-ancestor <commitRev> <commitRev>

DESCRIPTION:
    merge_base finds best common ancestor(s) between two commits. One common
    ancestor is better than another common ancestor if the latter is an ancestor
    of the former.
    A common ancestor that does not have any better common ancestor is a best
    common ancestor, i.e. a merge base. Note that there can be more than one
    merge base for a pair of commits.
    Commits that does not share a common history has no common ancestors.

OPTIONS:
    As the most common special case, specifying only two commits on the command
    line means computing the merge base between the given two commits.

--independent
    List the sub group from the passed commits, that cannot be reached from any
    other of the passed ones. In other words, it prints a minimal subset of the
    supplied commits with the same ancestors.

--is-ancestor
    Check if the first commit is an ancestor of the second one, and exit with
    status 0 if true, or with status 1 if not. Errors are signaled by a non-zero
    status that is not 1.

DISCUSSION
    Given two commits A and B, merge_base A B will output a commit which is the
    best common ancestor of both, what means that is reachable from both A and B
    through the parent relationship.

    For example, with this topology:

             o---o---o---o---B
            /       /
    ---3---2---o---1---o---A

    the merge base between A and B is 1.

    With the given topology 2 and 3 are also common ancestors of A and B, but
    they are not the best ones because they can be also reached from 1.

    When the history involves cross-cross merges, there can be more than one best
    common ancestor for two commits. For example, with this topology:

    ---1---o---A
        \ /
         X
        / \
    ---2---o---o---B

    When the history involves feature branches depending from another feature
    branches there can be also more than one common anvestor. For example:


           o---o---o
          /         \
         1---o---A   \
        /       /     \
    ---o---o---2---o---o---B

    In both examples, both 1 and 2 are merge-bases of A and B for each situation.
    Neither one is better than the other (both are best merge bases) because 1
    can not be reached from 2, nor the opposite.
```

dpordomingo:
by the way, wdyt @mcuadros about adding the core of this command (`filterCommitIter`, `MergeBase`, `IsAncestor` and `Independents`) into `git-base` itself?
That way, `MergeBase` could be also used by other apps.

dpordomingo:
## TODO:
- before merging this, I should review exit codes, because in `git merge-base`, exit code `1` is for `nothing found`, and other error codes are for real errors; My proposal does not respect that currently, and imo it should be done.

smola:
@mcuadros  I think the backing iterators and so on would belong to go-git core, maybe a `GetMergeBase` function too?

dpordomingo:
If you want to see how would it look like I can do it in my next OSD (that hopefully would be next Monday)

dpordomingo:
superseded by:
- https://github.com/src-d/go-git/pull/1097 with the core of the functionality
- https://github.com/src-d/go-git/pull/1096 with the command itself

Issue #1083, opened 2019-03-09
Flaky diff behavior


vmarkovtsev:
`diffmatchpatch.DiffMatchPatch` has a member named `DiffTimeout`, which is `time.Second` by default. If the diff calculation takes more than that limit, it is aborted and the current, half-baked results are returned immediately. This is really, really evil for all the users. Imagine that you are running some actions which involve go-git diffs under heavy load, e.g. in gitbase, or you run them on a retarded machine, such as TravisCI. You are going to have flaky behavior, which is extremely hard to debug.

I have recently hit this problem in Hercules actually (https://github.com/src-d/hercules/pull/234#issuecomment-471200837), and it was quite painful.

I suggest to at least make the timeout limit configurable and documented in `diff.go`, at maximum set it to a safely big value, like one hour.

cc @ajnavarro

PR #1084, opened 2019-03-10
Increase diffmatchcpatch timeout

vmarkovtsev:
Fixes https://github.com/src-d/go-git/issues/1083

Signed-off-by: Vadim Markovtsev <vadim@sourced.tech>

I can add a test, but it requires running 100 goroutines under `-race` and increases Travis run time by 10 minutes.

### DIFF
--- a/utils/diff/diff.go
+++ b/utils/diff/diff.go
@@ -14,10 +14,24 @@ import (
 )
 
 // Do computes the (line oriented) modifications needed to turn the src
-// string into the dst string.
+// string into the dst string. The underlying algorithm is Meyers,
+// its complexity is O(N*d) where N is min(lines(src), lines(dst)) and d
+// is the size of the diff.
 func Do(src, dst string) (diffs []diffmatchpatch.Diff) {
+	// the default timeout is time.Second which may be too small under heavy load
+	return DoWithTimeout(src, dst, time.Hour)
+}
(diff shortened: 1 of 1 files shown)


utils/diff/diff.go
```diff
@@ -16,6 +17,7 @@ import (
 // string into the dst string.
 func Do(src, dst string) (diffs []diffmatchpatch.Diff) {
 	dmp := diffmatchpatch.New()
+	dmp.DiffTimeout = time.Hour  // the default is time.Second which may be too little under heavy load
```
mcuadros: Well, maybe we can do this an argument?

Maybe `DoWithTimeout`?

vmarkovtsev: Then we should remove the current `Do` function, because the root of the problem is a too small default timeout. wdyt?

mcuadros: Nope, first we need to preserve the API, and second, I am not saying that we are going to keep the second, just that should be configurable.

vmarkovtsev: So I should add `DoWithTimeout` and also preserve the changed default timeout, correct?

mcuadros: Yup

vmarkovtsev:
@mcuadros Done

PR #1086, opened 2019-03-12
fix "reference delta not found" when cloning from aws codecommit

kcburge:
The existing code does not handle the case where a ref delta refers to another
deltified object.  Once the "parent" was undeltified, not attempt was made to
resolve references to that object's SHA1.

This change simply checks to see if there is an ExternalRef to the object just
resolved, and if so, replaces it and adopts its children.

I'm not an expert in the bowels of git, but this fixed my inability to clone
from codecommit.

### DIFF
--- a/plumbing/format/packfile/parser.go
+++ b/plumbing/format/packfile/parser.go
@@ -286,6 +286,18 @@ func (p *Parser) resolveDeltas() error {
 				if err := p.resolveObject(ioutil.Discard, child, content); err != nil {
 					return err
 				}
+
+				extRef, ok := p.oiByHash[child.SHA1]
+				if ok && extRef.ExternalRef {
+					// replace parent placeholder
+					p.oiByHash[child.SHA1] = child
+
+					// adopt children
+					child.Children = extRef.Children
+					for _, c := range child.Children {
+						c.Parent = child
+					}
+				}
 			}
 
 			// Remove the delta from the cache.


jfontan:
@kcburge, are you able to iterate over all the objects after that change? I believe this change fixes index creation but there may be more problems.

jfontan (approved):
I believe we don't have fixtures to test this case that would be good to have. Still this change looks good.

kcburge:
I resolved the DCO issue. I apologize for not having time to resolve the codecov issue.

mcuadros:
@kcburge can you provide a test?

kcburge:
I don't have anything to test with - I ran into this issue pulling from AWS codecommit (and not all repositories had this issue -- which is not invalid behavior of Git).  It has been quite a while since I thought about this, but, from what I recall, the packfile objects parsed in order like this:

Object 1 OBJ_BLOB
Object 2 OBJ_REF_DELTA (based on Object 1)
Object 3 OBJ_REF_DELTA (based on Object 2) <-- this is where it would blow up

Basically the second (and more) subsequent ref delta's referred to SHA1's that didn't exist until the prior ref deltas were resolved.  I don't know how hard it would be to mock this data structure.

Sorry I can't be more helpful at this time.

mcuadros:
What you can do upload to go-git-fixture a copy of the packfile or the capture the HTTP response to create a fixture.

PR #1088, opened 2019-03-14
plumbing: object, Count stats properly when no new line added at the …

### LINKED ISSUE #1074: Chunk stats calculation has wrong line count for additions and deletions
Current code in `patch.go` has:
```go
		for _, chunk := range fp.Chunks() {
			switch chunk.Type() {
			case fdiff.Add:
				cs.Addition += strings.Count(chunk.Content(), "\n")
			case fdiff.Delete:
				cs.Deletion += strings.Count(chunk.Content(), "\n")
			}
		}
```
That will have wrong count as if chunk content is `line1\nline2` that is 2 additions/deletions not one.

Looks like this calculation is wrong only if there is no newline at the end of file

oleksii-shnyra:
…end. Fixes #1074

Signed-off-by: Oleksii Shnyra <oleksii@global>

### DIFF
--- a/plumbing/object/commit_stats_test.go
+++ b/plumbing/object/commit_stats_test.go
@@ -1,0 +1,78 @@
+package object_test
+
+import (
+	"time"
+
+	"gopkg.in/src-d/go-git.v4"
+	"gopkg.in/src-d/go-git.v4/plumbing"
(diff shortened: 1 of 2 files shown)


jfontan:
Can you please add some tests for this change? Otherwise LGTM. Thanks!

PR #1090, opened 2019-03-20
ssh: leverage proxy from environment

dweomer:
I was working on [a little git-cli drop-in](/dweomer/git-semver) that creates an orphan branch for tracking (per-branch) versioning information when I realized that it did not work for clone/pull/push with SSH URLs in a proxied environment (good ole SOCKS5). I made a small change in `ssh.command#connect()` to replace the invocation of `ssh.Dial()` to a copy of that function that uses `golang.org/x/net/proxy.FromEnvironment()` to get a dialer and then hands off the dialed connection to `ssh.NewClientConn`. This breaks the connection timeout facility because `proxy.FromEnvironment()` returns a `proxy.Dialer` which is a single-method interface type (that doesn't support a timeout or context arg).

The cool thing about `proxy.FromEnvironment()` is that it returns a working `Dialer` (i.e. directly connecting) if no proxy details are found or the setup for recognized proxy details fails for any reason.

There are some existing issues tracking some sort of `proxy.DialContext` implementation:

* golang/go#17759
* golang/go#19688

I am not confident that the Go folks will be tackling this lack of a timeout-capable dialer in `x/net/proxy` anytime soon so I understand if the behavioral breakage in this proposed change is not acceptable.

Signed-off-by: Jacob Blain Christen <dweomer5@gmail.com>

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,6 +3,7 @@ module gopkg.in/src-d/go-git.v4
 require (
 	github.com/alcortesm/tgz v0.0.0-20161220082320-9c5fe88206d7 // indirect
 	github.com/anmitsu/go-shlex v0.0.0-20161002113705-648efa622239 // indirect
+	github.com/armon/go-socks5 v0.0.0-20160902184237-e75332964ef5
 	github.com/davecgh/go-spew v1.1.1 // indirect
 	github.com/emirpasic/gods v1.12.0
 	github.com/flynn/go-shlex v0.0.0-20150515145356-3f9db97f8568 // indirect
@@ -19,7 +20,7 @@ require (
 	github.com/stretchr/testify v1.3.0 // indirect
 	github.com/xanzy/ssh-agent v0.2.1
 	golang.org/x/crypto v0.0.0-20190422183909-d864b10871cd
-	golang.org/x/net v0.0.0-20190420063019-afa5a82059c6 // indirect
+	golang.org/x/net v0.0.0-20190502183928-7f726cade0ab
 	golang.org/x/sys v0.0.0-20190422165155-953cdadca894 // indirect
 	golang.org/x/text v0.3.0
 	gopkg.in/check.v1 v1.0.0-20180628173108-788fd7840127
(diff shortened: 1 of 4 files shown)


mcuadros:
This looks awesome! Can you create test for this?

dweomer:
@mcuadros added a `ssh.ProxySuite` that shims `ssh.UploadPackSuite` tests with a socks5 proxy, on localhost, in front of the git server.

dweomer:
Please keep in mind, this change does not honor the `config.Timeout` that is passed into `dial` and so I will understand if this is considered a regression.

I thought of writing a simple select wrapper (with the timeout) around the `proxy.Dialer#Dial()` invocation but that would leak go routines.

Let me know if you would prefer that the timeout still be honored and I will have another at it.

dweomer:
I am looking into submitting a couple of upstream fixes (x/net/proxy, and x/crypto/ssh) that would result in this change being simpler.

dweomer:
@mcuadros I've updated to use my `proxy.Dial` change from golang/net#38.

I have added a test suite that is just a repeat of the [plumbing/transport/ssh.UploadPackSuite](https://github.com/src-d/go-git/blob/master/plumbing/transport/ssh/upload_pack_test.go#L24) but communicating to a SOCKSv5 server via [github.com/armon/go-socks5](https://github.com/armon/go-socks5).

dweomer:
@mcuadros Additionally, I checked the coverage of the [plumbing/transport/ssh](https://github.com/src-d/go-git/tree/master/plumbing/transport/ssh) package and it was at 74.2% before my change and at 74.3% after (according to `go test -cover ./plumbing/transport/ssh/.`)

dweomer:
So:
Screen Shot 2019-05-02 at 1 21 03 PM

But:
Screen Shot 2019-05-02 at 1 21 33 PM

🤔

dweomer:
@mcuadros thank you for the merge.

When should I look for a release/tag? I can add a `replace` directive to my downstream [`go.mod`](https://github.com/edgexfoundry/git-semver/blob/master/go.mod) but then we will stop getting updates to the v4 stream.

PR #1096, opened 2019-03-25
Add merge base command

dpordomingo:
Uses the code from https://github.com/src-d/go-git/pull/1097 to implement a command, mimicking `git merge-base`.

Help with a full description of the implemented functionalities is shown running `merge_base --help`.

usage: 
```shell
$ ./merge_base

Returns the merge-base between two commits:

usage: merge_base <path> <commitRev> <commitRev>
   or: merge_base <path> --independent <commitRev>...
   or: merge_base <path> --is-ancestor <commitRev> <commitRev>
   or: merge_base --help

    (no option)       lists the best common ancestors of the two passed commits
    --independent     list commits not reachable from the others
    --is-ancestor     is the first one ancestor of the other?
    --help            show the full help message of merge_base
```

PR #1097, opened 2019-03-25
Create merge-base feature

dpordomingo:
suggested by https://github.com/src-d/go-git/pull/1078

~blocked by https://github.com/src-d/go-git-fixtures/pull/13~
blocks https://github.com/src-d/lookout/pull/530

This PR adds some features from `git merge-base` command.

_**note:**_
_you can see more details about this PR, alternatives and performance concerns in the comments below_

Iit was implemented a new walker `filterCommitIter` (like `bfsCommitIterator` one), and three functions `MergeBase`, `IsAncestor` and `Independents`.
- `plumbing/object.filterCommitIter` 
  > _it's a `CommitIter` that walks the commit history (in Breadth-first order) of the passed commit. It only returns the commits that validate the passed `isValid CommitFilter`, and only traverse the commits that do not validate `isLimit CommitFilter`._
- `git.MergeBase(first, second *object.Commit) ([]*object.Commit, error)`
  > _it mimics the behavior of `git merge-base a b`, returning the best common ancestor of the two passed commits; the best common ancestors cannot be reached from other common ancestors._
- `git.IsAncestor(first, second *object.Commit) (bool, error)` 
  > _it returns `true` if the first commit is an ancestor of the second one, like `git merge --is-ancestor` does_
- `git.Independents(commits []*object.Commit) ([]*object.Commit, error)` 
  > _it returns a subset of the passed commits, that are not reachable from any other, as `git merge-base --independent commit...` does_

### DIFF
--- a/plumbing/object/commit_walker_bfs_filtered.go
+++ b/plumbing/object/commit_walker_bfs_filtered.go
@@ -1,0 +1,176 @@
+package object
+
+import (
+	"io"
+
+	"gopkg.in/src-d/go-git.v4/plumbing"
+	"gopkg.in/src-d/go-git.v4/plumbing/storer"
+)
+
+// NewFilterCommitIter returns a CommitIter that walks the commit history,
+// starting at the passed commit and visiting its parents in Breadth-first order.
+// The commits returned by the CommitIter will validate the passed CommitFilter.
+// The history won't be transversed beyond a commit if isLimit is true for it.
+// Each commit will be visited only once.
+// If the commit history can not be traversed, or the Close() method is called,
+// the CommitIter won't return more commits.
+// If no isValid is passed, all ancestors of from commit will be valid.
+// If no isLimit is limmit, all ancestors of all commits will be visited.
+func NewFilterCommitIter(
+	from *Commit,
+	isValid *CommitFilter,
+	isLimit *CommitFilter,
+) CommitIter {
+	var validFilter CommitFilter
+	if isValid == nil {
+		validFilter = func(_ *Commit) bool {
+			return true
+		}
+	} else {
+		validFilter = *isValid
+	}
+
+	var limitFilter CommitFilter
+	if isLimit == nil {
+		limitFilter = func(_ *Commit) bool {
+			return false
+		}
+	} else {
+		limitFilter = *isLimit
+	}
+
+	return &filterCommitIter{
+		isValid: validFilter,
+		isLimit: limitFilter,
+		visited: map[plumbing.Hash]struct{}{},
+		queue:   []*Commit{from},
+	}
+}
+
+// CommitFilter returns a boolean for the passed Commit
+type CommitFilter func(*Commit) bool
+
+// filterCommitIter implments CommitIter
+type filterCommitIter struct {
+	isValid CommitFilter
+	isLimit CommitFilter
+	visited map[plumbing.Hash]struct{}
+	queue   []*Commit
+	lastErr error
+}
+
+// Next returns the next commit of the CommitIter.
+// It will return io.EOF if there are no more commits to visit,
+// or an error if the history could not be traversed.
+func (w *filterCommitIter) Next() (*Commit, error) {
+	var commit *Commit
+	var err error
+	for {
+		commit, err = w.popNewFromQueue()
+		if err != nil {
+			return nil, w.close(err)
+		}
+
+		w.visited[commit.Hash] = struct{}{}
+
+		if !w.isLimit(commit) {
+			err = w.addToQueue(commit.s, commit.ParentHashes...)
+			if err != nil {
+				return nil, w.close(err)
+			}
+		}
+
+		if w.isValid(commit) {
+			return commit, nil
+		}
+	}
+}
+
+// ForEach runs the passed callback over each Commit returned by the CommitIter
+// until the callback returns an error or there is no more commits to traverse.
+func (w *filterCommitIter) ForEach(cb func(*Commit) error) error {
+	for {
+		commit, err := w.Next()
+		if err == io.EOF {
+			break
+		}
+
+		if err != nil {
+			return err
+		}
+
+		if err := cb(commit); err == storer.ErrStop {
+			break
+		} else if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+// Error returns the error that caused that the CommitIter is no longer returning commits
+func (w *filterCommitIter) Error() error {
+	return w.lastErr
+}
+
+// Close closes the CommitIter
+func (w *filterCommitIter) Close() {
+	w.visited = map[plumbing.Hash]struct{}{}
+	w.queue = []*Commit{}
+	w.isLimit = nil
+	w.isValid = nil
+}
+
+// close closes the CommitIter with an error
+func (w *filterCommitIter) close(err error) error {
+	w.Close()
+	w.lastErr = err
+	return err
+}
+
+// popNewFromQueue returns the first new commit from the internal fifo queue,
+// or an io.EOF error if the queue is empty
+func (w *filterCommitIter) popNewFromQueue() (*Commit, error) {
+	var first *Commit
+	for {
+		if len(w.queue) == 0 {
+			if w.lastErr != nil {
+				return nil, w.lastErr
+			}
+
+			return nil, io.EOF
+		}
+
+		first = w.queue[0]
+		w.queue = w.queue[1:]
+		if _, ok := w.visited[first.Hash]; ok {
+			continue
+		}
+
+		return first, nil
+	}
+}
+
+// addToQueue adds the passed commits to the internal fifo queue if they weren't seen
+// or returns an error if the passed hashes could not be used to get valid commits
+func (w *filterCommitIter) addToQueue(
+	store storer.EncodedObjectStorer,
+	hashes ...plumbing.Hash,
+) error {
+	for _, hash := range hashes {
+		if _, ok := w.visited[hash]; ok {
+			continue
+		}
+
+		commit, err := GetCommit(store, hash)
+		if err != nil {
+			return err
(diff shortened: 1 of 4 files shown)


dpordomingo:
# alternatives

**(A)** `plumbing/object.filterCommitIter` could be replaced by the already existent `plumbing/object.bfsCommitIterator`:
- pros: less new code,
- cons:
  - the functionality of `isLimit` and `isValid` would be lost, and it should be implemented again and again for those who need it (like in `git.MergeBase`, but also in any other cases where it could be needed)
  - considering the performance analysis done below, imo the proposed improvement could not be done (because current `plumbing/object.bfsCommitIterator` does not let you remove items from the iter queue once they have been added, as `isLimit` functionality could do).

**(B)** since `plumbing/object.bfsCommitIterator` is a special subcase of `plumbing/object.filterCommitIter`, it could be (safely) deleted, and `plumbing/object.NewCommitIterBSF` could be rewritten using the new iter, as done here https://github.com/dpordomingo/go-git/commit/ab3f18f640778a2b5a40568d6cf8e491809bae8a
- pros: less code,
- cons: idk if there could be some tricky use case of current `plumbing/object.bfsCommitIterator` not being documented nor tested in current tests; example: the async usage of "seenExternal", as it was added as a test-case in https://github.com/dpordomingo/go-git/commit/ab3f18f640778a2b5a40568d6cf8e491809bae8a, and it would be also supported with that new implementation.

I'd vote for  **(B)**, but it could be done in a subsequent PR, or if required during the revision.

dpordomingo:
# Explaining recursive behavior

given `git.Independents([]*object.Commit) ([]*object.Commit, error)` relies on a recursive function, @jfontan exposed his concerns about its performance at https://github.com/src-d/lookout/pull/530#discussion_r262083898.

Here is the explanation about how it works :point_down: 

The history can potentially be walked as many times as candidates provided for `Independents`.
But every time a candidate is found as an ancestor of any other, it is removed from the candidates' list, so it won't be considered anymore.
The order in which you request the candidates no longer affects how many times the history is traversed because they're previously sorted by `Committer.When desc`, assuming that older commits most probably won't be descendant of newer ones.

In the worst scenario, the history of each candidate will be fully traversed as many times as orphan branches are included in the candidates to be analyzed.

given this history:
```
          o----o----B~2---B^---B
         /         /
--A~4---o---A~2---o---A

---o---C^---C
```

then (assuming that topological order matches the historical order):
- **(A)** `Independents(B, B^)`; `1step`; walks the history only once, because starting by `B`, it reaches in one step `B^`; once `B^` is reached, it is removed from the candidates list, and since there only one candidate left (`B`), it is returned with no more steps. 
- **(B)**`Independents(B^, B)`; `1step`, because ordering the candidates by `Committer.When desc`, will convert them as being **(A)**;
- **(C)**`Independents(B, B^, B~2)`; `2steps`; walks only once (as in the previous example), but it requires two steps to reach `B^` and then `B~2`
- **(D)**`Independents(B, A~4)`; `7steps`; walks the history only once (as in the previous example), but it will require more steps to reach `A~4`
- **(E)** `Independents(B, A)`; `1full+2steps`; during first walk from `B` it won't be reached `A`, and during the second walk from `A`, it will be aborted once `A~2` is reached, because, beyond that commit, all the history is already known.
- **(F)**  `Independents(B, B^, B~2, A, A~2, A~4, C, C^)`; `2full+2steps`; during the first history walk from `B` it will be removed `B^`, `B~2`, `A~2`, `A~4` from the candidates list, so the second iteration will consider only `B`, `A`, `C`, `C^`; then from `A` the walk will be stopped once reached `A~2` as in **(E)**; the third iteration, starting from `C` will find (and remove) `C^` from the list, so the result will be only `A`, `B` and `C`.

assuming that topological order **DOES NOT** match the historical order for `B` and `B^`:
-  **(G)**,  `1full + 1step`; walks the history twice, because first iteration walking from `B^` does not reach `B` after a full walk from `B^`; then in the second iteration walking from `B` it is the same case as **(A)**, so it is returned `A` with only one extra step

dpordomingo:
# Performance ~"benchmarks"~

As proposed by @jfontan https://github.com/src-d/lookout/pull/530#discussion_r262083898, I did some tests:

### Running over `kubernetes/kubernetes` and `torvalds/linux`

The time needed to calculate the common ancestor with `go-git` stays as expected and explained above, and in edge-cases, it is almost the same than the time required to do a `log` between the involved commits:
- when one of the passed commits is an ancestor of the other, the time needed depends on its distance;
- when the passed commits are Independent, it is needed one full history scan from the newer, and the process takes as much as `go-git` needs to iterate over all its ancestors; 

On `kubernetes/kubernetes` and `torvalds/linux`, doing a full traverse of its histories from `master`:

|repo|`master` ancestors|list all|list only one|
|--|--|--|--|
|kubernetes/kubernetes|76018 commits|`~5s`|`~0.03s`|
|torvalds/linux|812644 commits|`~60s`|`~5s`|

`git log master` tooked `~10s` over `torvalds/linux`, and almost `~1s` over `kubernetes/kubernetes`.


### Testing real PRs from `kubernetes/kubernetes`:

I tried `go-git/_examples/merge_base` command over the last 245 merged PR from `kubernetes/kubernetes` (there is no valid PRs in `linux` repo), to calculate the `merge-base` between their `base` and `head`:
  - 140 PRs required a full traverse, taking `~5s`,
  - 105 PRs were analyzed almost instantaneously in `~.3s`, without a full history traverse.

## Conclusions and possible improvements:

When the passed commits are _Independents_ it is needed to fully traverse the history at least once.

There is a possible improvement to avoid it:

- Stop walking the history of the ancestors of the common ancestors already found.

If you want, it could be done in this or in a subsequent PR.

dpordomingo:
Travis is failing https://travis-ci.org/src-d/go-git/jobs/511125046#L668 because the new test cases depend on new fixtures, to be merged by https://github.com/src-d/go-git-fixtures/pull/13

dpordomingo:
[AppVeyor failed](https://ci.appveyor.com/project/mcuadros/go-git/builds/23345917) for the same reason; now that src-d/go-git-fixtures#13 was merged it should pass, but I can not restart the build.

mcuadros:
I have some doubts about the API, since go-git, is getting being some times is hard to find if X functionality is implemented or not. How this PR is done, the functions are spare, in a big package, some my I want to consider 2 different options:

- Move the functions to be methods of `Commit`, fits perfect for `MergeBase`, `IsAncestor`, not for `Independents`
- Move the functions to a new package, called `merge` or something similar

/cc @jfontan @smola

dpordomingo:
I agree with your @mcuadros and both proposals would fit me, so just let me know.
About keeping `Independents` in a different package than the two others... I don't like it; I'd move all at once.
Would it be ok if moved to `plumbing/mergebase` or similar?

merge_base.go
```diff
@@ -0,0 +1,211 @@
...
+	if err == errIsReachable {
+		return true, nil
+	}
+
+	return false, nil
+}
+
+// ancestorsIndex returns a map with the ancestors of the starting commit if the
+// excluded one is not one of them. It returns errIsReachable if the excluded commit
+// is ancestor of the starting, or another error if the history is not transversable.
```
smola:
```suggestion
// is ancestor of the starting, or another error if the history is not traversable.
```

merge_base.go
```diff
@@ -0,0 +1,211 @@
...
+			excluded[fromAncestor.Hash] = true
+			return nil
+		})
+
+		if err != nil {
+			return nil, err
+		}
+
+		if len(res) < len(candidates) {
+			return independents(res, excluded, indexOf(res, from)+1)
```
smola: Let's avoid recursion? Git histories can get to >1M commits.

smola:
`merge` or `mergebase` package names look too specific to hold all `MergeBase`, `IsAncestor` and `Independents`, since the later two are more general.

merge_base.go
dpordomingo: I'm not sure about understanding your suggestion.
This recursive call precisely reduces the iterations and steps needed to find the independent commits, because once one is found as reachable from the others, it removes it, and jumps to the next one.

dpordomingo: ...or I could try with a `continue`... to make it more explicit how the filtering works :dancer:
(I'll try later)

mcuadros:
@smola what about the other option?

dpordomingo:
@smola I removed the recursive call at 5e0de4e

yelirekim:
Is this work abandoned, or likely to move forward?  I'm trying to decide whether to fork and merge this, or if it's likely to end up in the upstream sometime soon.

Is all that's left here to decide on where the package lines are drawn?

dpordomingo:
Not dead on my side ;)
I'll resume the work now that my vacations finished. If the maintainers agree :)

dpordomingo:
Thanks for your suggestions!

Done as you recommended during your code review:
All the new functionality was moved ef3c257e6cf0c7fc307e3f0ab132e92071b25c65 into `plumbing/object`:
- `MergeBase` and `IsAncestor` will belong to `plumbing/object.Commit`
- `Independents` will be an independent function in `plumbing/object`

Also, I refactored `IsAncestor` to be more performant 52ede6ca9475dd5e7c7b30a6ad05b0316d848ff9

should I address any other issue?

I should squash some commits in order to have a cleaner history, so let me know before you merge.

mcuadros:
Yes please clean the history

dpordomingo:
@mcuadros squashed in one commit f62852b

ces131:
Is there going to be a new release soon containing this merge?? (If not, there definitely should be!)

dpordomingo:
I'll also rebase https://github.com/src-d/go-git/pull/1096 on top of current go-git, now that this has been merged.
Imo it could be useful if it would be released at the same time that this change.

PR #1103, opened 2019-04-04
Fix tag message encoding/decoding

scjudd:
The Decode method was adding one too many newlines to the tag message, causing signature verification to fail.


scjudd:
Actually, hold on, this breaks in the case that the tag message has newlines in it.

scjudd:
I should note that the error occurred with a signed tag made with `git tag -s -m 'This is a signed tag!'` It appears that, although this fixes the issue for tags made with `git`, this is failing unit tests. I'm digging into it.

scjudd:
Closing until I come up with something that fixes my use case and also passes unit tests.

PR #1104, opened 2019-04-04
plumbing: object, Fix tag message decoding

scjudd:
The `Decode` method was adding one too many newlines to the tag message, causing signature verification to fail. This is because in signed tags produced by `git`, there is a newline after the PGP signature block, resulting in `messageAndSig` having one extra (empty) `[]byte` element. This caused `t.Message` to receive one extra newline.


mcuadros:
Can you provide a test? Thanks!

mcuadros:
Included in https://github.com/src-d/go-git/pull/1118

PR #1115, opened 2019-04-18
plumbing: commit.StatsContext and fix for orphan commit

mcuadros:
This introduce the missing `commit.StatsContext` and also fix a bug when `commit.Stats` its executed over a orphan commit.

### DIFF
--- a/plumbing/object/commit.go
+++ b/plumbing/object/commit.go
@@ -297,7 +297,7 @@ func (c *Commit) Stats() (FileStats, error) {
 }
 
 // StatsContext returns the stats of a commit. Error will be return if context
-// expires. Provided context must be non-nil
+// expires. Provided context must be non-nil.
 func (c *Commit) StatsContext(ctx context.Context) (FileStats, error) {
 	fromTree, err := c.Tree()
 	if err != nil {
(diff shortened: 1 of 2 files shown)


plumbing/object/commit.go
```diff
@@ -76,7 +76,7 @@ func (c *Commit) Tree() (*Tree, error) {
 	return GetTree(c.s, c.TreeHash)
 }
 
-// Patch returns the Patch between the actual commit and the provided one.
+// PatchContext returns the Patch between the actual commit and the provided one.
 // Error will be return if context expires. Provided context must be non-nil
```
jfontan:
```suggestion
// Error will be return if context expires. Provided context must be non-nil.
```

plumbing/object/commit.go
```diff
@@ -291,25 +291,33 @@ func (b *Commit) encode(o plumbing.EncodedObject, includeSig bool) (err error) {
...
+// Stats returns the stats of a commit.
 func (c *Commit) Stats() (FileStats, error) {
-	// Get the previous commit.
-	ci := c.Parents()
-	parentCommit, err := ci.Next()
+	return c.StatsContext(context.Background())
+}
+
+// StatsContext returns the stats of a commit. Error will be return if context
+// expires. Provided context must be non-nil
```
jfontan:
```suggestion
// expires. Provided context must be non-nil.
```

PR #1120, opened 2019-04-20
filesystem: ObjectStorage, keep packfiles open if KeepDescriptors is enabled

filipnavara:
This gives a mild performance improvement if one happens to hit objects that are sequential and still cached by the `Scanner` in the `Packfile`. It also avoids recreating a lot of objects.


filipnavara:
Closing in favor of #1123.

PR #1121, opened 2019-04-20
plumbing: packfile, apply small object reading optimization also for delta objects

filipnavara:
In #963 I addressed issue with reading many small objects from a pack file. I originally opted not to apply the optimization to delta objects. It turns out that when traversing a commit history the tree objects are often small and delta encoded. This extends the original optimization to handle this case too. Tested on https://github.com/rails/rails repository and it improves the complete history traversal time by ~20%.

### DIFF
--- a/plumbing/format/packfile/packfile.go
+++ b/plumbing/format/packfile/packfile.go
@@ -478,6 +478,7 @@ func (i *objectIter) Next() (plumbing.EncodedObject, error) {
 				}
 			} else if obj, ok := i.p.cacheGet(e.Hash); ok {
 				if obj.Type() != i.typ {
+					i.p.offsetToType[int64(e.Offset)] = obj.Type()
 					continue
 				}
 				return obj, nil
@@ -493,12 +494,14 @@ func (i *objectIter) Next() (plumbing.EncodedObject, error) {
 						return nil, err
 					}
 					if typ != i.typ {
+						i.p.offsetToType[int64(e.Offset)] = typ
 						continue
 					}
 					// getObjectType will seek in the file so we cannot use getNextObject safely
 					return i.p.objectAtOffset(int64(e.Offset), e.Hash)
 				} else {
 					if h.Type != i.typ {
+						i.p.offsetToType[int64(e.Offset)] = h.Type
 						continue
 					}
 					return i.p.getNextObject(h, e.Hash)
--- a/plumbing/format/packfile/packfile.go
+++ b/plumbing/format/packfile/packfile.go
@@ -83,10 +83,8 @@ func (p *Packfile) Get(h plumbing.Hash) (plumbing.EncodedObject, error) {
 // offset.
 func (p *Packfile) GetByOffset(o int64) (plumbing.EncodedObject, error) {
 	hash, err := p.FindHash(o)
-	if err == nil {
-		if obj, ok := p.deltaBaseCache.Get(hash); ok {
-			return obj, nil
-		}
+	if err != nil {
+		return nil, err
 	}
 
 	return p.objectAtOffset(o, hash)
@@ -180,10 +178,16 @@ func (p *Packfile) getObjectType(h *ObjectHeader) (typ plumbing.ObjectType, err 
 		err = ErrInvalidObject.AddDetails("type %q", h.Type)
 	}
 
+	p.offsetToType[h.Offset] = typ
+
 	return
 }
 
 func (p *Packfile) objectAtOffset(offset int64, hash plumbing.Hash) (plumbing.EncodedObject, error) {
+	if obj, ok := p.deltaBaseCache.Get(hash); ok {
+		return obj, nil
+	}
+
 	h, err := p.objectHeaderAtOffset(offset)
 	if err != nil {
 		if err == io.EOF || isInvalid(err) {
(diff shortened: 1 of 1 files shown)


plumbing/format/packfile/packfile.go
```diff
@@ -194,21 +198,42 @@ func (p *Packfile) objectAtOffset(offset int64) (plumbing.EncodedObject, error)
...
+			return p.getNextObject(h)
+		}
 
-	hash, err := p.FindHash(h.Offset)
-	if err != nil {
-		return nil, err
-	}
+		// For delta objects we read the delta data and create a special object
+		// that will hold them in memory and resolve them lazily to the referenced
+		// object.
```
filipnavara: I forgot to update this commit. It's from my previous approach and it's no longer relevant.

plumbing/format/packfile/packfile.go
```diff
@@ -312,18 +344,24 @@ func (p *Packfile) fillREFDeltaObjectContent(obj plumbing.EncodedObject, ref plu
 	obj.SetType(base.Type())
 	err = ApplyDelta(obj, base, buf.Bytes())
 	p.cachePut(obj)
-	bufPool.Put(buf)
 
 	return err
 }
 
 func (p *Packfile) fillOFSDeltaObjectContent(obj plumbing.EncodedObject, offset int64) error {
 	buf := bytes.NewBuffer(nil)
```
jfontan: Shouldn't we get the buffer from the pool instead of creating a new one?

```go
buf := bufPool.Get().(*bytes.Buffer)
```

filipnavara: Yes, I missed that one. I'll fix it in a minute.

jfontan:
I'm investigating a decrease in speed when using these changes it [gitbase](https://github.com/src-d/gitbase).

When iterating all trees with a cold cache it's faster with these changes (35s vs 25s). We keep the cache and in a second run the previous go-git code is faster (16s vs 25s). It seems the cache is not being filled correctly and in one profile I can see that `Packfile.objectAtOffset` uses 3.13s an with these changes it uses 11.08s, most of it in recreating again the objects that should be in cache.

filipnavara:
There's couple of flaws I didn't realize when coding this.

Firstly, it slightly regresses scenario where someone was accessing the small delta objects only to check the header. Now it falls back to decoding the content which can be comparatively expensive. This notably regresses `packfile.GetByType` iterations and by proxy the high level methods like `CommitObjects`. Not only it will read more data from the file it can end up loading enough to cause cache spills.

Secondly, I assumed that if the delta size is small and the expanded size is small then it should not be a big issue to load the base object and just apply the delta. The flaw is that the delta object can reference huge base object but only reuse a small part of it.

I have ideas on how to fix both of the issues without introducing new regressions. It will require significant reworking of the logic and a lot of testing though.

plumbing/format/packfile/packfile.go
```diff
@@ -176,10 +178,16 @@ func (p *Packfile) getObjectType(h *ObjectHeader) (typ plumbing.ObjectType, err
...
 	}
 
+	p.offsetToType[h.Offset] = typ
+
 	return
 }
 
-func (p *Packfile) objectAtOffset(offset int64) (plumbing.EncodedObject, error) {
+func (p *Packfile) objectAtOffset(offset int64, hash plumbing.Hash) (plumbing.EncodedObject, error) {
+	if obj, ok := p.deltaBaseCache.Get(hash); ok {
```
jfontan: Use `p.cacheGet(hash)` to check for nil cache.

plumbing/format/packfile/packfile.go
```diff
@@ -188,27 +196,54 @@ func (p *Packfile) objectAtOffset(offset int64) (plumbing.EncodedObject, error)
...
+		size = p.getDeltaObjectSize(buf)
+		if size <= smallObjectThreshold {
+			var obj = new(plumbing.MemoryObject)
+			obj.SetSize(size)
+			if h.Type == plumbing.REFDeltaObject {
+				err = p.fillREFDeltaObjectContentWithBuffer(obj, h.Reference, buf)
+			} else {
+				err = p.fillOFSDeltaObjectContentWithBuffer(obj, h.OffsetReference, buf)
+			}
+			return obj, nil
```
jfontan: `err` is lost. There should be an error check or `return obj, err`.

jfontan:
After more investigation I've found that the iterator is not using the cache to retrieve objects so delta objects are regenerated even when they are already in the cache. After enabling it the performance increases but triggers a bug I am investigating.

filipnavara:
> After more investigation I've found that the iterator is not using the cache to retrieve objects so delta objects are regenerated even when they are already in the cache.

I forgot to fix the "cold" code path in the iterator to try the cache.

Note that if `offsetToType` is kept around (ie. packfiles are reused) it will hit the cache properly because it will always find the offset in this map and then either skip over the object or load it through `GetByOffset` which is checking the cache.

filipnavara:
> After enabling it the performance increases but triggers a bug I am investigating.

I see... I will look into it tomorrow :-)

filipnavara:
I think I figured out what's wrong... `getObjectType` in the iterator code can seek and thus the next `getNextObject` call reads wrong data.

PR #1122, opened 2019-04-20
plumbing: packfile/scanner, add option to skip calculating checksum

filipnavara:
…and use the option at places where the checksum is ignored anyway.

### DIFF
--- a/plumbing/format/packfile/packfile.go
+++ b/plumbing/format/packfile/packfile.go
@@ -50,7 +50,7 @@ func NewPackfileWithCache(
 	file billy.File,
 	cache cache.Object,
 ) *Packfile {
-	s := NewScanner(file)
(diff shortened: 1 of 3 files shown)


filipnavara:
Seems that #1124 addresses my performance concerns in a better way where the CRC calculation is fast enough to be ignored. Feel free to close this PR in favor of that one.

jfontan:
Even if #1124 alleviates the problem I also think that it's good to have a switch to disable generation of CRC. If I remember correctly CRC is only useful for index generation or verification. Probably it's better to wait until the changes from #1124 are merged and add the switch on top of that.

filipnavara:
The overhead of calculating the CRC may be small enough to be negligable now. I'll rework this PR and test it on my use case to see whether there's still any point in pursuing it.

PR #1123, opened 2019-04-21
filesystem: ObjectStorage, MaxOpenDescriptors option

saracen:
The `MaxOpenDescriptors` option provides a middle ground solution between keeping all pack files open (as offered by the `KeepDescriptors` option) and keeping none open.

This PR is a modification of #1109. This version moves to keeping packfile objects around, rather than just the open file. Whilst updating this, PR #1120 was opened that offers effectively the same solution to that problem. But there's a few other changes here in relation to that, too:
- Using `KeepDescriptors` when calling `NewStorageWithOptions` no longer passes the option to dotgit, as there's little point in keeping a map of open files in both locations.
- Packfile's scanner can be accessed (packfile.Scanner()), this avoids creating a new scanner around the file when the packfile is already open (as done here https://github.com/src-d/go-git/pull/1123/files#diff-e927e977f53290ca7a0e091f6ffd9081R456).
- The dotgit version now uses the packfile's hash as the map key, instead of the path name. This results in a nice performance boost when iterating over many files, as `objectPackPath` (using path.Join) isn't as fast as you'd imagine.

### DIFF
--- a/storage/filesystem/object.go
+++ b/storage/filesystem/object.go
@@ -26,6 +26,10 @@ type ObjectStorage struct {
 
 	dir   *dotgit.DotGit
 	index map[plumbing.Hash]idxfile.Index
+
+	packList    []plumbing.Hash
+	packListIdx int
+	packfiles   map[plumbing.Hash]*packfile.Packfile
 }
 
 // NewObjectStorage creates a new ObjectStorage with the given .git directory and cache.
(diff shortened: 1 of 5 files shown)


jfontan:
LGTM. Thanks!

storage/filesystem/object.go
```diff
@@ -187,6 +191,57 @@ func (s *ObjectStorage) encodedObjectSizeFromUnpacked(h plumbing.Hash) (
 	return size, err
 }
 
+func (s *ObjectStorage) packfile(idx idxfile.Index, pack plumbing.Hash) (*packfile.Packfile, error) {
+	if s.packfiles == nil {
+		if s.options.MaxOpenDescriptors > 0 {
```
mcuadros: can you move this logic to a function?

saracen: I've now split it up so the code that handles what is more clearly defined. At the start we check to see if the packfile is cached with `packfileFromCache`, at the end we do `storePackfileInCache` if a new packfile was opened.

PR #1124, opened 2019-04-21
plumbing: packfile/scanner, readability/performance improvements, zlib pooling

saracen:
This migrates scanner's `trackableReader`, `byteReadSeeker` and `teeReader`'s functionality into one `scannerReader`. I wanted the ability to have a pool of Scanners if need be, and offering a `Reset()` and dealing with all 3 readers was proving difficult. I believe this version is also easier to read.

In addition:
- There's a zlib pool. Pooling zlib isn't obvious, as although it has a Reset() function, `zlib.NewReader` doesn't accept nil. This limitation is by-passed by priming the reader with a zero byte compressed payload. I couldn't find a better way of doing this.
- Previously, writing to the crc hash still often performed small writes (via Read(), not just ReadByte()). Now all small reads are buffered when they need to be. The buffer size is now smaller and sized to take advantage of the CLMUL instruction crc32 uses.
- Deprecated scanner's `Flush()`. There's no need for this to be public.

### DIFF
--- a/plumbing/format/packfile/scanner.go
+++ b/plumbing/format/packfile/scanner.go
@@ -39,8 +39,7 @@ type ObjectHeader struct {
 }
 
 type Scanner struct {
-	r   reader
-	zr  readerResetter
+	r   *scannerReader
 	crc hash.Hash32
 
 	// pendingObject is used to detect if an object has been read, or still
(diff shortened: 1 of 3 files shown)


filipnavara:
I have seen some woefully unoptimal code paths in that code when I was recently profiling Gitea. I'd gladly give it a try once it passes the CI tests. It may be a better alternative to #1122.

plumbing/format/packfile/scanner.go
```diff
@@ -378,110 +378,82 @@ func (s *Scanner) Close() error {
...
+// then only whence=io.SeekCurrent is supported, any other operation fails.
+func (r *scannerReader) Seek(offset int64, whence int) (int64, error) {
+	var err error
 
-	n, err = r.reader.Read(p)
-	if n > 0 {
-		if n, err := r.w.Write(p[:n]); err != nil {
-			return n, err
+	if seeker, ok := r.reader.(io.ReadSeeker); !ok {
+		if whence != io.SeekCurrent && offset != 0 {
```
filipnavara: should be `||`

saracen: Thanks! I've just noticed the last second modification I made that's breaking the tests too. Will push shortly :)

saracen: Got there in the end :)

plumbing/format/packfile/scanner.go
```diff
@@ -56,19 +55,27 @@ type Scanner struct {
...
 	crc := crc32.NewIEEE()
 	return &Scanner{
-		r:          newTeeReader(newByteReadSeeker(seeker), crc),
+		r:          newScannerReader(r, crc),
 		crc:        crc,
 		IsSeekable: ok,
 	}
 }
 
+func (s *Scanner) Reset(r io.Reader) {
```
jfontan: It would be nice to hace a test that reuses scanner with `Reset`. Other than that LGTM! Thanks!

saracen: Good shout! I noticed a problem whilst figuring out how/what to test. Even when the reader is seekable, I'm returning the `r.offset` when `whence == io.SeekCurrent && offset == 0`, which incorrect assumes that the reader provided to start with was set at the beginning of the file.

I'll fix that and add a test shortly!

saracen: I've added two tests. `TestReaderReset` checks basics usage and that fields are correctly reset. `TestReaderResetSeeks` checks that the seek behavior is correct.

PR #1125, opened 2019-04-21
plumbing: TreeWalker performance improvement, bufio pool for objects

saracen:
Removed `path.Clean` and `path.Join`, as they're expensive in comparison to basic string manipulation that can be used here.

Added bufio.Buffer pool to be used by tag, tree and commit objects.

### DIFF
--- a/plumbing/object/tree.go
+++ b/plumbing/object/tree.go
@@ -230,7 +230,9 @@ func (t *Tree) Decode(o plumbing.EncodedObject) (err error) {
 	}
 	defer ioutil.CheckClose(reader, &err)
 
-	r := bufio.NewReader(reader)
+	r := bufPool.Get().(*bufio.Reader)
+	defer bufPool.Put(r)
+	r.Reset(reader)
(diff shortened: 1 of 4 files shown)


filipnavara:
Are you reading my mind? Seriously, I just hit this performance bottleneck on my code for generating bloom filters :)

plumbing/object/tree.go
```diff
@@ -433,9 +439,13 @@ func (w *TreeWalker) Next() (name string, entry TreeEntry, err error) {
 		return
 	}
 
-	if t, ok := obj.(*Tree); ok {
-		w.stack = append(w.stack, &treeEntryIter{t, 0})
-		w.base = path.Join(w.base, entry.Name)
+	if obj != nil {
+		w.stack = append(w.stack, &treeEntryIter{obj, 0})
+		if len(w.base) > 0 {
```
mcuadros: Can you move this code to a function instead of duplicate it?

PR #1126, opened 2019-04-21
plumbing: format/index perf, buffered reads, reflection removal

saracen:
Large performance increase by buffering reads.

There were a few instances where binary.Read() would end up using reflection on &plumbing.Hash, rather than treating it as a byte slice. This has now been resolved.

### DIFF
--- a/utils/binary/read.go
+++ b/utils/binary/read.go
@@ -25,6 +25,15 @@ func Read(r io.Reader, data ...interface{}) error {
 
 // ReadUntil reads from r untin delim is found
 func ReadUntil(r io.Reader, delim byte) ([]byte, error) {
+	if bufr, ok := r.(*bufio.Reader); ok {
+		value, err := bufr.ReadBytes(delim)
+		if err != nil || len(value) == 0 {
+			return nil, err
+		}
+
+		return value[:len(value)-1], nil
+	}
+
 	var buf [1]byte
 	value := make([]byte, 0, 16)
 	for {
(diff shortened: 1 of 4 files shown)


utils/binary/read.go
```diff
@@ -25,6 +25,15 @@ func Read(r io.Reader, data ...interface{}) error {
 
 // ReadUntil reads from r untin delim is found
 func ReadUntil(r io.Reader, delim byte) ([]byte, error) {
+	if bufr, ok := r.(*bufio.Reader); ok {
```
mcuadros: Can you move this to a function called `ReadUntilFromBufioReader` or something like that, and call it from here?

saracen: Yeah, will push shortly with a test for it.

The only problem with this approach is that bufio.Reader satisfies io.Reader too, so if you pass a bufio.Reader to it, it's not going to be immediately obvious you're getting slower behaviour.

There's very few places using ReadUntil, so it's not an issue now. But it's something to look out for in the future and to ensure that if the reader is buffered, they're using the correct variant.

mcuadros: I mean call one from the other, while the new behavior is public

mcuadros:
```
if bufr, ok := r.(*bufio.Reader); ok {
	return ReadUntilFromBufioReader(bufr)
}
```

saracen: Done! Sorry about that, re-reading your comment, it was clear enough. My bad.

Issue #1116, opened 2019-04-18
[FR] Add method PGPPayload() for tag and commit


sapk:
Before doing a PR I would like to know your if it would be accepted and what it should look like to be accepted.

Currently only the full Encode method of a commit/tag is exposed publicly. In the goal of being able to verify the payload directly from a keyring and not a armored key string like the Verify method of tag and commit, I would suggest to add another method like `PGPPayload() string` that export the result of `t.encode(o, false)`. This method could also be re-use in the Verify of commit and tag.

I mainly ask because it doesn't follow the contributing rule: `You should be able to run the same query using git. We don't accept features that are not implemented in the official git implementation.` as I don't think we can export without the sig with the git cli. But this method would be very useful and limit the duplication of code in and outside go-git to do the same.

filipnavara:
Notably it's something we run into in Gitea. The GitHub API exposes the PGP payload to the users (https://developer.github.com/v3/git/commits/) and libgit2 has an API to get it (https://libgit2.org/libgit2/#HEAD/group/commit/git_commit_extract_signature).

sapk:
Close by  #1127

PR #1127, opened 2019-04-21
plumbing: object/{commit,tag} add EncodeWithoutSignature, Implement #1116

sapk:
Expose the PGP payload of commits and tags like discussed in #1116.

Signed-off-by: Antoine GIRARD <sapk@sapk.fr>

### DIFF
--- a/plumbing/object/commit.go
+++ b/plumbing/object/commit.go
@@ -235,6 +235,11 @@ func (b *Commit) Encode(o plumbing.EncodedObject) error {
 	return b.encode(o, true)
 }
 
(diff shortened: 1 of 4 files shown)


plumbing/object/commit.go
```diff
@@ -233,6 +233,16 @@ func (b *Commit) Encode(o plumbing.EncodedObject) error {
 	return b.encode(o, true)
 }
 
+// PGPPayload export to a Reader the commit payload that is signed
+func (b *Commit) PGPPayload() (io.Reader, error) {
```
mcuadros: I rather keep the pattern of `Encode` and just expose `EncodeWithoutSignature` for example.

sapk: done

PR #1130, opened 2019-04-23
plumbing: format/gitattributes support

saracen:
Implements gitattributes parsing, matching and attribute extraction.

I've tried to mirror the gitignore API as much as possible. 

## go doc
```
VARIABLES

var (
        ErrMacroNotAllowed      = errors.New("macro not allowed")
        ErrInvalidAttributeName = errors.New("Invalid attribute name")
)

TYPES

type Attribute interface {
        Name() string
        IsSet() bool
        IsUnset() bool
        IsUnspecified() bool
        IsValueSet() bool
        Value() string
        String() string
}

type MatchAttribute struct {
        Name       string
        Pattern    Pattern
        Attributes []Attribute
}

func LoadGlobalPatterns(fs billy.Filesystem) (attributes []MatchAttribute, err error)
    LoadGlobalPatterns loads gitattributes patterns and attributes from the
    gitattributes file declared in a user's ~/.gitconfig file. If the
    ~/.gitconfig file does not exist the function will return nil. If the
    core.attributesFile property is not declared, the function will return nil.
    If the file pointed to by the core.attributesfile property does not exist,
    the function will return nil. The function assumes fs is rooted at the root
    filesystem.

func LoadSystemPatterns(fs billy.Filesystem) (attributes []MatchAttribute, err error)
    LoadSystemPatterns loads gitattributes patterns and attributes from the
    gitattributes file declared in a system's /etc/gitconfig file. If the
    /etc/gitconfig file does not exist the function will return nil. If the
    core.attributesfile property is not declared, the function will return nil.
    If the file pointed to by the core.attributesfile property does not exist,
    the function will return nil. The function assumes fs is rooted at the root
    filesystem.

func ParseAttributesLine(line string, domain []string, allowMacro bool) (m MatchAttribute, err error)
    ParseAttributesLine parses a gitattribute line, extracting path pattern and
    attributes.

func ReadAttributes(r io.Reader, domain []string, allowMacro bool) (attributes []MatchAttribute, err error)
    ReadAttributes reads patterns and attributes from the gitattributes format.

func ReadAttributesFile(fs billy.Filesystem, path []string, attributesFile string, allowMacro bool) ([]MatchAttribute, error)
func ReadPatterns(fs billy.Filesystem, path []string) (attributes []MatchAttribute, err error)
    ReadPatterns reads gitattributes patterns recursively through the directory
    structure. The result is in ascending order of priority (last higher).

    The .gitattribute file in the root directory will allow custom macro
    definitions. Custom macro definitions in other directories .gitattributes
    will return an error.

type Matcher interface {
        // Match matches patterns in the order of priorities.
        Match(path []string, attributes []string) (map[string]Attribute, bool)
}
    Matcher defines a global multi-pattern matcher for gitattributes patterns

func NewMatcher(stack []MatchAttribute) Matcher
    NewMatcher constructs a new matcher. Patterns must be given in the order of
    increasing priority. That is the most generic settings files first, then the
    content of the repo .gitattributes, then content of .gitattributes down the
    path.

type MatcherOptions struct{}

type Pattern interface {
        // Match matches the given path to the pattern.
        Match(path []string) bool
}
    Pattern defines a gitattributes pattern.

func ParsePattern(p string, domain []string) Pattern
    ParsePattern parses a gitattributes pattern string into the Pattern
    structure.

```

### DIFF
--- a/plumbing/format/gitattributes/attributes.go
+++ b/plumbing/format/gitattributes/attributes.go
@@ -1,0 +1,214 @@
+package gitattributes
+
+import (
+	"errors"
+	"io"
+	"io/ioutil"
+	"strings"
+)
+
+const (
+	commentPrefix = "#"
+	eol           = "\n"
+	macroPrefix   = "[attr]"
+)
+
+var (
+	ErrMacroNotAllowed      = errors.New("macro not allowed")
+	ErrInvalidAttributeName = errors.New("Invalid attribute name")
+)
+
+type MatchAttribute struct {
+	Name       string
+	Pattern    Pattern
+	Attributes []Attribute
+}
+
+type attributeState byte
+
+const (
+	attributeUnknown     attributeState = 0
+	attributeSet         attributeState = 1
+	attributeUnspecified attributeState = '!'
+	attributeUnset       attributeState = '-'
+	attributeSetValue    attributeState = '='
+)
+
+type Attribute interface {
+	Name() string
+	IsSet() bool
+	IsUnset() bool
+	IsUnspecified() bool
+	IsValueSet() bool
+	Value() string
+	String() string
+}
+
+type attribute struct {
+	name  string
+	state attributeState
+	value string
+}
+
+func (a attribute) Name() string {
+	return a.name
+}
+
+func (a attribute) IsSet() bool {
+	return a.state == attributeSet
+}
+
+func (a attribute) IsUnset() bool {
+	return a.state == attributeUnset
+}
+
+func (a attribute) IsUnspecified() bool {
+	return a.state == attributeUnspecified
+}
+
+func (a attribute) IsValueSet() bool {
+	return a.state == attributeSetValue
+}
+
+func (a attribute) Value() string {
+	return a.value
+}
+
+func (a attribute) String() string {
+	switch a.state {
+	case attributeSet:
+		return a.name + ": set"
+	case attributeUnset:
+		return a.name + ": unset"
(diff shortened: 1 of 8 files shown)


plumbing/format/gitattributes/attributes.go
```diff
@@ -0,0 +1,214 @@
+package gitattributes
```
mcuadros: Maybe it better simply call the package `attributes`

plumbing/format/gitattributes/dir_test.go
```diff
@@ -0,0 +1,199 @@
+package gitattributes
+
+import (
+	"os"
+	"os/user"
+	"strconv"
+
+	. "gopkg.in/check.v1"
+	billy "gopkg.in/src-d/go-billy.v4"
```
mcuadros: remove the alias

plumbing/format/gitattributes/attributes.go
saracen: I did wonder that, but stuck with `gitattributes` as go-git already has `gitignore`. Git's own documentation also refers to it as "gitattributes" (https://git-scm.com/docs/gitattributes).

I don't mind either way though

plumbing/format/gitattributes/dir_test.go
mcuadros: can you make this change please?

saracen: Done!

PR #1131, opened 2019-04-23
go modules update

mcuadros:


jfontan:
I believe you should execute `go mod tidy` to delete old versions from `go.sum`.

PR #1136, opened 2019-04-25
plumbing: format/idxfile, avoid creating temporary buffers to decode integers

filipnavara:
On Linux repository (210 MB pack index; 6,596,227 objects) this saves around 2 seconds in `genOffsetHash` function.

### DIFF
--- a/plumbing/format/idxfile/idxfile.go
+++ b/plumbing/format/idxfile/idxfile.go
@@ -5,8 +5,9 @@ import (
 	"io"
 	"sort"
 
+	encbin "encoding/binary"
+
 	"gopkg.in/src-d/go-git.v4/plumbing"
-	"gopkg.in/src-d/go-git.v4/utils/binary"
 )
 
 const (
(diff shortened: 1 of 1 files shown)


filipnavara:
The profile shows there's still some room for optimization here:

image

plumbing/format/idxfile/idxfile.go
```diff
@@ -213,22 +204,19 @@ func (idx *MemoryIndex) genOffsetHash() error {
...
-			return err
+	var hash plumbing.Hash
+	i := uint32(0)
+	for firstLevel, fanoutValue := range idx.Fanout {
+		mappedFirstLevel := idx.FanoutMapping[firstLevel]
+		for secondLevel := uint32(0); i < fanoutValue; i++ {
+			copy(hash[:], idx.Names[mappedFirstLevel][secondLevel*objectIDLength:])
+			offset := int64(idx.getOffset(mappedFirstLevel, int(secondLevel)))
+			idx.offsetHash[offset] = hash
+			secondLevel++
```
jfontan: Is this a leftover?

filipnavara: Nope. The loop is a bit weird since it's iterating over two variables. Maybe there's some better Go syntax for it?

`secondLevel` goes from 0 to maximum value on that fanout level (not checked directly). `i` goes from 0 over both the inner and outer loop and bounds the inner loop.

PR #1142, opened 2019-05-03
git : allows to create a Remote without a Repository

EmrysMyrddin:
Git has the `ls-remote` command which allows to remotely retrieve the list of references.

We want to use it to list remote tags without cloning the entire repository for example.

To implement the same behaviour with `go-git` we need to create a `Remote` structure by hand and use the `List` function.

I have added a runnable example which prints out the list of tags of a sample repository.

### DIFF
--- a/_examples/ls-remote/main.go
+++ b/_examples/ls-remote/main.go
@@ -1,0 +1,42 @@
+package main
+
+import (
+	"log"
+
+	"gopkg.in/src-d/go-git.v4"
+	"gopkg.in/src-d/go-git.v4/config"
+	"gopkg.in/src-d/go-git.v4/storage/memory"
+)
+
+// Retrieve remote tags without cloning repository
+func main() {
+
+	// Create the remote with repository URL
+	rem := git.NewRemote(memory.NewStorage(), &config.RemoteConfig{
+		Name: "origin",
+		URLs: []string{"https://github.com/Zenika/MARCEL"},
+	})
+
+	log.Print("Fetching tags...")
+
+	// We can then use every Remote functions to retrieve wanted informations
+	refs, err := rem.List(&git.ListOptions{})
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	// Filters the references list and only keeps tags
+	var tags []string
+	for _, ref := range refs {
+		if ref.Name().IsTag() {
+			tags = append(tags, ref.Name().Short())
+		}
+	}
+
+	if len(tags) == 0 {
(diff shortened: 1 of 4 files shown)


_examples/ls-remote/main.go
```diff
@@ -0,0 +1,42 @@
...
+
+import (
+	"log"
+
+	"gopkg.in/src-d/go-git.v4"
+	"gopkg.in/src-d/go-git.v4/config"
+	"gopkg.in/src-d/go-git.v4/storage/memory"
+)
+
+// Retreive remote tags without cloning repository
```
nlepage:
```suggestion
// Retrieve remote tags without cloning repository
```

_examples/ls-remote/main.go
```diff
@@ -0,0 +1,42 @@
...
+
+	// Create the remote with repository URL
+	rem := git.NewRemote(memory.NewStorage(), &config.RemoteConfig{
+		Name: "origin",
+		URLs: []string{"https://github.com/Zenika/MARCEL"},
+	})
+
+	log.Print("Fetching tags...")
+
+	// We can then use every Remote functions to retreive wanted informations
```
nlepage:
```suggestion
	// We can then use every Remote functions to retrieve wanted informations
```

_examples/ls-remote/main.go
```diff
@@ -0,0 +1,42 @@
...
+	refs, err := rem.List(&git.ListOptions{})
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	// Filters the reference list and only keep tags
```
nlepage:
```suggestion
	// Filters the references list and only keeps tags
```

remote.go
```diff
@@ -45,7 +45,7 @@ type Remote struct {
 	s storage.Storer
 }
 
-func newRemote(s storage.Storer, c *config.RemoteConfig) *Remote {
+func NewRemote(s storage.Storer, c *config.RemoteConfig) *Remote {
```
nlepage:
```suggestion
// NewRemote creates a new Remote.
// The intended purpose is to use the Remote for tasks such as listing remote references (like using git ls-remote).
// Otherwise Remotes should be created via the use of a Repository.
func NewRemote(s storage.Storer, c *config.RemoteConfig) *Remote {
```

mcuadros: can you document the new exported function?

EmrysMyrddin: sorry for the late response, I have added the documentation

PR #1152, opened 2019-05-17
Fix an edge case when fetching remotes with unrelated incomplete history. Fixes #1151

### LINKED ISSUE #1151: NewCommitPreorderIter Fails with `object not found` When iterating over incomplete history.
And this causes `Fetch` to fail under certain circumstances, such as (at least) when fetching remotes with unrelated histories.

NonLogicalDev:
In getHaves, allow the getHavesRef to fail with `object not found` when
iterating over incomplete history, and don't treat it as a fatal error.

It might also worth to revise the `object.NewCommitPreorderIter` logic,
to perhaps not fail once reachable commits are exhausted.


mcuadros:
How behaves git?

NonLogicalDev:
It does not error out when attempting to fetch on a shallowly cloned repository. Just checked:
`git clone <repo> --depth 1`
At this point the history of the branch only has one commit.
`git fetch --depth 10`
No errors and the history now contains 10 commits, as one would expect.

In the same situation `go-git` catches an error `object not found` when enumerating `wants`

NonLogicalDev:
It does not error out when attempting to fetch on a shallowly cloned repository. Just checked:

```
git init
git remote add repo1 <url1>
git fetch repo1 --depth 1
git remote add repo2 <url2>
git fetch repo2 --depth 1 // this works in git, but equivalent fails in `go-git`
```

In the same situation `go-git` catches an error `object not found` when enumerating `wants`

NonLogicalDev:
@mcuadros Any update on this?

Issue #1149, opened 2019-05-16
storage/filesystem: dotgit, could not get blob from packfiles


vvatanabe:
Could not get blob from packfiles, occurs a error `packfile not found`.

The cause is an unnecessary temporary file parsing error generated by `git gc`.

We should correctly check the packfile name format and skip over not loading those temporary files,

name format of temporary file:
`.tmp-12345-pack-{hash}.pack`

I have a PR.

filipnavara:
I have run into multiple instances where something like that happens. It seems that regular GIT sometimes produces and consumes a `objects/info/packs` file with a list of actively used packfiles. Often I have some more pack files in `objects/pack` that were not cleaned during previous repacking.

PR #1153, opened 2019-05-17
could not get blob from packfiles

vvatanabe:
fixes by #1149

First, I updated the test case to reproduce the problem, and I will add a patch to fix this problem.


vvatanabe:
This problem was reproduced a build below.
https://travis-ci.org/src-d/go-git/builds/533637111

vvatanabe:
Fixed, plz chk it.

mcuadros:
If I am not wrong the .tmp files are files to ignore, are you sure that git is reading this files?

vvatanabe:
Yes. We provide Git hosting service and tried go-git experimentally.

When `git gc` did not complete successfully on bare repository, `.tmp_xxx.pack` remained.  And when using `func (r * Repository) BlobObject (h plumbing.Hash) (* object.Blob, error)` to get a packed blob read `.tmp_xxx.pack` internally and occured a error by to get a wrong hash.

filipnavara:
Related: https://github.com/src-d/go-git/issues/1223. As I mentioned in https://github.com/src-d/go-git/issues/1149#issuecomment-492969265 the logic that GIT uses is actually different. I don't think the .tmp files should be ignored but the `objects/info/packs` file should be consulted.

mcuadros:
Merged in https://github.com/go-git/go-git/pull/6

saracen:
@mcuadros Super happy to see go-git continuing! I only found out it'd moved from this PR. The new repository mentions there was some legal issues - are you able to post a little about that somewhere and maybe post to r/golang about where the project has now landed?

PR #1160, opened 2019-06-04
fix wildcard handling in RefSpec matching

novas0x2a:
1) The guard logic here was inverted, resulting in an always-false
   branch, which meant that the suffix after the wildcard was
   incorrectly ignored.
2) Wildcards were treated as 1-or-more matches, but git treats them as
   0-or-more. This change aligns go-git with git, but represents a bit
   of a breaking change for go-git.


jfontan:
I believe that windows tests are failing because a network problem. @mcuadros can you restart the test?

Issue #457, opened 2017-07-01
Help diagnosing Blame concurrency problems


TheDahv:
**Go Version**: `go version go1.8 darwin/amd64`
**go-git Version**: fresh install with `go get gopkg.in/src-d/go-git.v4/...`

First off, thank you for this project! I was very excited when I saw an alternative to `git2go` that didn't have a dependency on `libgit2` and that had a blame implementation. For context, I'm working on a problem that involves scanning files that were changed in a branch, blaming each of them, and then determining experience by committer via git-blame to recommend a reviewer for a pull request.

This one will be odd to write because I don't know what exactly is triggering this problem. For the example code, I have 1 simple repo in which this works and another more complex one where it doesn't. So I apologize in advance; I'll do my best to describe what I'm seeing and hopefully you can guide me toward providing better info.

## Description

I'm unable to [`Blame`](https://godoc.org/gopkg.in/src-d/go-git.v4#Blame) multiple files concurrently at the same commit. My motivation here was to run the blames in parallel in light of [known performance issues](https://github.com/src-d/go-git/issues/211#issuecomment-272824675) .

Specifically I get 2 kinds of crashes:

- `fatal error: concurrent map writes` from `gopkg.in/src-d/go-git.v4/plumbing/format/packfile.(*Decoder).decodeByHeader`
- `fatal error: concurrent map read and map write` from `gopkg.in/src-d/go-git.v4/storage/filesystem.(*ObjectStorage).findObjectInPackfile`

Here's a [stack trace](https://gist.github.com/TheDahv/2ac24ad5ae9476b3f501d5b034d85116) for the bug from the repro program below.

The former would fire immediately in serial contexts. The latter I got in a concurrent context after adding locks to the Decoder methods to guard the `hashToOffset` and `offsetToHash` maps.

Here's [a little diff patch gist](https://gist.github.com/TheDahv/ed3445b3d83423b8e3da98bf9026b34f) to show what I tried changing to protect against the crashes. Unfortunately this slowed performance to a crawl. 

Frankly I don't feel like sprinkling locks in without a firmer understanding of the design goals is an effective approach for me anyway.

## Repro

Here's a [little snippet](https://gist.github.com/TheDahv/49e4c619a372c2e1de9944e3bc69a23c) from my larger code that illustrates how I isolated the bugs.

[...]

PR #1168, opened 2019-06-19
MemoryIndex allows concurrent read

niukuo:
Signed-off-by: niukuo <niukuo@gmail.com>

add a lock in MemoryIndex for concurrent read


mcuadros:
by design the library is not thread-safe. Thanks anyway

niukuo:
> by design the library is not thread-safe. Thanks anyway

https://github.com/src-d/go-git/issues/457#issuecomment-312815087

This pr didn't break any public api, so can you reconsider it?

PR #1183, opened 2019-07-21
Fixed tautological error conditions

muesli:
Removed err handling where err's value can't possibly have changed.

### DIFF
--- a/plumbing/format/commitgraph/encoder.go
+++ b/plumbing/format/commitgraph/encoder.go
@@ -24,8 +24,6 @@ func NewEncoder(w io.Writer) *Encoder {
 
 // Encode writes an index into the commit-graph file
 func (e *Encoder) Encode(idx Index) error {
-	var err error
-
 	// Get all the hashes in the input index
 	hashes := idx.Hashes()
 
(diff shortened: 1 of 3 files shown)


plumbing/format/commitgraph/encoder.go
```diff
@@ -56,9 +56,6 @@ func (e *Encoder) Encode(idx Index) error {
 			return err
 		}
 	}
-	if err != nil {
-		return err
-	}
```
filipnavara: Can you explain me why this is not necessary? It could be rewritten to `else` but the error handling should stay there.

muesli: Because there's no way for `err` to not be `nil`. It's only assigned as `var err error` above, everything else below uses err in its own `if` scope and deals with it directly.

muesli: @filipnavara I assume that means you actually want to handle any error returned from `encodeCommitData`? Happy to fix that, but currently it's not going to happen, because the outer `err` gets shadowed.

filipnavara: Yep, the shadowing was unintentional.

muesli: Fixed and pushed!

filipnavara: Thanks

PR #1188, opened 2019-07-21
Don't assign err from ForEach loop

muesli:
Since we don't check the value anyway, as it can't possibly be anything but nil.


mcuadros:
Thanks for keep go-git clean and tidy. Can you group this batch of PRs in one? Since all are related?

muesli:
@mcuadros Sure, which ones would you like to see merged? I actually tried to split them up in logical, related pieces already. For example, I don't think simple fixes like this one belong together with changes that actually change behavior, or test-related changes. Up to you though, just let me know.

mcuadros:
I suggest to merge all in one, since all are about the same nature.

mcuadros:
I did myself, thanks!

PR #1197, opened 2019-07-28
Make http.AuthMethod setAuth public. Fixes #1196

### LINKED ISSUE #1196: Allow providing custom http auth method
The setAuth method of the http.AuthMethod is private.
https://github.com/src-d/go-git/blob/1edb992dbc419a0767b1cf3a524b0d35529799f5/plumbing/transport/http/common.go#L178
This prevents anyone to provide a custom auth method.

Making this method public will allow implementation of custom auth.

hypnoce:


hypnoce:
@mcuadros Thanks for the merge.
When do you think it's possible to release a new version of go-git ?

mcuadros:
We need to merge a few others things, once this happens I will release a new version

PR #1204, opened 2019-08-04
Handle EOF error in commitFileIter.ForEach

knqyf263:
`commitFileIter.ForEach` always returns io.EOF error, so the first commit in this PR fails. 
The second commit fixes this problem by handling io.EOF.

### DIFF
--- a/plumbing/object/commit_walker_file.go
+++ b/plumbing/object/commit_walker_file.go
@@ -128,7 +128,9 @@ func isParentHash(hash plumbing.Hash, commit *Commit) bool {
 func (c *commitFileIter) ForEach(cb func(*Commit) error) error {
 	for {
 		commit, nextErr := c.Next()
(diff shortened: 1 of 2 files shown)


plumbing/object/commit_walker_file.go
```diff
@@ -128,7 +128,9 @@ func isParentHash(hash plumbing.Hash, commit *Commit) bool {
 func (c *commitFileIter) ForEach(cb func(*Commit) error) error {
 	for {
 		commit, nextErr := c.Next()
-		if nextErr != nil {
+		if nextErr == io.EOF {
+			break
+		} else if nextErr != nil {
```
mcuadros: remove this unnecessary else

knqyf263: Thank you for the review. I've fixed it.

PR #1206, opened 2019-08-04
Repository.Log: add limiting options to LogOptions

knqyf263:
Add commit limiting options.
It is equivalent to running `git log --since <date>` or `git log --until <date>`.

In the future, I want to add the other limiting options.
e.g. git log --author <pattern> or git log --merges.


knqyf263:
I resolved the conflict

PR #1243, opened 2019-11-15
Add global and system support

### LINKED ISSUE #760: Add .gitconfig support
Not sure if a first class object to represent the `.gitconfig` file should be provided or if the global config information should be squirreled away inside `config.Config`.  There is the `~/.gitignore_global` file to consider as well.

It seems to me that a first class object and storer interface is the way to go.  These are real physical entities and they should have their counterparts in the `config` package.

djgilcrease:
Loads the config the same way git does, System, then Global, then Local so overrides are maintained properly.

Closes #760


cap10morgan:
Thanks for submitting this PR, @djgilcrease. I would love to have this functionality for a project I'm working on using go-git. It seems that they've moved all development to this repo: https://github.com/go-git/go-git. Can you resubmit this PR over there?

cap10morgan:
Ran into an issue with this, which I may try to fix and submit an updated PR to the new go-git/go-git repo. Namely, when you load up your config with this code and then later run a `SetConfig`, it copies your merged system, global, local config into your local `.git/config` file. That's probably not what we want for general behavior there.

djgilcrease:
sounds good to me, I am on vacation for the next few weeks so will not be able to do anything with it for a long while.

PR #1246, opened 2019-11-20
plumbing/object: avoid O(N^2) string building when decoding commit me…

dsymonds:
…ssage

Most commits have relatively small messages, so this was never
noticeable. However, there are some repositories that have
semi-automated messages that can get very large (e.g.
github.com/riscv/riscv-clang and its riscv-trunk branch), on the order
of 109k lines. Changing from string += to using a bytes.Buffer reduces
the time for Commit.Decode for that specific case from 35s to 74ms.


dsymonds:
Hey, any chance this'll get merged soon, @mcuadros @smola ?

Issue #1220, opened 2019-09-23
Respect ~/.ssh/config for host to key mapping


swarupdonepudi:
It would be nice if go-git can use the config specified in ~/.ssh/config file to figure out the Auth part of the Clone operation. `git` already supports this.

dontlaugh:
Is there a reason this isn't implemented? E.g. would you accept a contribution that 

* locates and parses a user's $HOME/.ssh/config
* Selects an appropriate ssh key for auth

My team encountered this on CI, where we needed to select a key to push with. CircleCI will automatically write an .ssh/config for the user running in the container, and drop the additional key into the .ssh folder, but go-git doesn't find it automatically.

PR #1265, opened 2019-12-19
[WIP] Use the user and port specified in the ~/.ssh/config

### LINKED ISSUE #1261: SSH authentication failed twoards Gerrit
**Good morning!**
I'm having problems authenticating against my companys gerrit server.

## What I have tried
1. Issue `eval "$(ssh-agent)" && ssh-add` in my terminal running the experiment. This seems to work as expected. Also, don't know if it is needed. 
2. run script provided below:
```go
package main

import (
	git "gopkg.in/src-d/go-git.v4"
	"log"
)

var url = "ssh://gerrit.<name>:<port>/path/to/repository"

func main() {
	_, err := git.PlainClone("app", false, &git.CloneOptions{
		URL:      url,
		Progress: nil,
	})
	if err != nil {
		log.Println(err)
	}
}
```
Output: `ssh: handshake failed: ssh: unable to authenticate, attempted methods [publickey none], no supported methods remain`

However, When i run `git clone <url>` it work as expected.

I have a `.ssh/config` file and it looks like this:
```
Host gerrit.<base-url>
     User <my_username>
     Port <port>
```

I don't know where to go from here regarding the debugging. I hope you don't mind me asking here, I tried to gain access to your slack but the link had expired.

Best regards,
Måns Ansgariusson

Mattemagikern:
The current implementation disregards of the ~/.ssh/config in most cases in favor for default values etc.
When I was trying this out against the company servers I failed to clone projects from the company server since the cloning address assumed the user and port specified in the ssh config file.

I think these changes will solve my little problem however I'm having great difficulty to use the commit as a package in my other files to try it out.
Would be very happy If anyone would like to try it out/ explain how I should go about using the local project to build my binary so I'm certain that the commit contains the correct behavior.

This is a start regarding: #1220 but I need some more guidance. 

closes #1261 

Best regards,
Signed-off-by: Måns Ansgariusson <mattemagikern@gmail.com>

### DIFF
--- a/plumbing/transport/common.go
+++ b/plumbing/transport/common.go
@@ -232,7 +232,7 @@ func parseSCPLike(endpoint string) (*Endpoint, bool) {
 	user, host, portStr, path := giturl.FindScpLikeComponents(endpoint)
 	port, err := strconv.Atoi(portStr)
 	if err != nil {
-		port = 22
+		port = 0
 	}
 
 	return &Endpoint{
--- a/plumbing/transport/common_test.go
+++ b/plumbing/transport/common_test.go
@@ -90,7 +90,7 @@ func (s *SuiteCommon) TestNewEndpointSCPLike(c *C) {
 	c.Assert(e.User, Equals, "git")
 	c.Assert(e.Password, Equals, "")
 	c.Assert(e.Host, Equals, "github.com")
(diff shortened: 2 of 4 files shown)


Mattemagikern:
This is not a correct solution to the problem..
However I don't know where it fails either.

PR #1269, opened 2019-12-25
Add Plan 9 support

### LINKED ISSUE #756: Plan 9 Support
You can get go-git to build on Plan 9 by adding a file:

worktree_plan9.go
```c
// +build plan9

package git

import (
	"syscall"
	"time"

	"gopkg.in/src-d/go-git.v4/plumbing/format/index"
)

func init() {
	fillSystemInfo = func(e *index.Entry, sys interface{}) {
		// Plan 9		
		sysi, ok := sys.(*syscall.Dir)

                // Problem because CreatedAt isn't a field
		e.CreatedAt = time.Unix(int64(sysi.Atime), 0)

		e.Dev = uint32(sysi.Dev)

                // Problem because Plan 9 has no Inode
		e.Inode = sysi.Qid.Vers
		
                // Problem because Plan 9 has string UID/GID
		e.GID = uint32(0)
		e.UID = uint32(0)
	}
}
```

This doesn't fix Plan 9 support fully, however, and is only a placeholder. I'm not sure what the best integration approach is, but `sys` isn't a `Stat_t` on Plan 9, it is the `syscall.Dir` type seen here: https://golang.org/src/syscall/dir_plan9.go

Looking at plumbing/format/index/index.go leads me to believe that some part of the git implementation assumes posix, so I was wondering if there was a clean work around that could be made compatible with the Plan 9 model.

fhs:
Not all the tests pass yet but this makes go-git usable on Plan 9.
Please merge this after https://github.com/src-d/go-billy/pull/78.

Fixes #756

Signed-off-by: Fazlul Shahriar <fshahriar@gmail.com>


mcuadros:
Merge at https://github.com/go-git/go-git/pull/1

Issue #1295, opened 2020-02-29
Is this project still active?


zeripath:
The last commit to this project was 1 November 2019. Is this project still under active development?

filipnavara:
I am a bit concerned about it too. Source-d, the company behind the project, went dark around 4 months ago without any official news explaining what happened.

filipnavara:
I tried to find more information but there's not much to report. All the previous maintainers of this project are no longer listed on src-d organization. Source{d} went dark last November, disappeared from all social media, the site is down and the community Slack (which is no longer accessible for new members) has no further information either. At this point I think the only reasonable course of action is to fork this project.

cc @jfontan @smola @mcuadros ... I don't expect any official information or blessing but if there's anyone who could shed a little bit of light on this and say some yay/nay on the fork it would be appreciated.

mcuadros:
Hi everyone, yes we this project is still active, @smola, @jfontan  and myself, we will maintain the project. The problem was that in November source{d} went bankrupt and this repository felt in a legal void. Actually any of us has merge rights any more. But don't worry we are figuring the how to tackle this problem, but as hard deadline this will be cleared in mid-March.

I founded a new company and this company will be the official maintainer of `go-git`, but we need to resolve the legal issues about the repository itself. If the issues are not cleared before the deadline we will move the efforts to a new org and URL and make a hard fork, but we want to avoid this as much as possible. 

Thanks for the patience

zeripath:
Hi @mcuadros !

Thank you for the update. 

You must have had a terribly stressful and frustrating few months. I hope things are settling down a little. 

If you can, please keep us informed. I'll suggest Gitea makes a mirror of this repo in case things go awry in the meantime.

mcuadros:
We have the hard fork at: https://github.com/go-git/go-git but we don't want to update anything since if we need to do the hard fork, we will merge the PRs and after that update all the paths from the library.

So I suggest you hold any movement for two weeks, this should be more than enough. This next week we are going to take the decision and during the next week we are going to take the actions need it to make this project maintained again.

mcuadros:
Kamino cloned this issue to [mcuadros/go-billy](https://github.com/mcuadros/go-billy/issues/1)

lunny:
@mcuadros Thank you for your efforts on this project.

[...]

PR #1299, opened 2020-03-20
Fix typo 

jeveleth:


filipnavara:
Please submit new pull requests to https://github.com/go-git/go-git. See https://github.com/src-d/go-git/pull/1298 and https://github.com/src-d/go-git/issues/1295#issuecomment-592965250 for reasons.

